PR #1, opened 2016-08-20
Move git-sync to a new repo of its very own

thockin:
This just imports the existing state of contrib/git-sync into this new repo and scrubs paths for "contrib".  I have some further work I want to do on this sidecar, so I thought it was time to break it out.


mikedanese:
There is a process for new repos now https://docs.google.com/document/d/1ugAd9Zj-jW3YHdrNVdktmvDMEWtChPqyGHfkwWdQ3zo/edit#heading=h.bv4kogapeb2x

thockin:
The process doesn't apply well to code that is moving from and existing
repo.  If I recall, we just bypassed the process, but I'll go re-read it.

thockin:
Yeah, this falls under the "existing" section:

> New Kubernetes Community Projects can be created by pulling out code from the Kubernetes main repo ...
> To do this the OWNERS of that package can directly graduate to a Kubernetes Community Project...

That said, I have some significant followups to this, so a rubber stamp is all that is required, really.

mikedanese:
Conveniently you left out the (except contrib/)

> New Kubernetes Community Projects can be created by pulling out code from the Kubernetes main repo (except contrib/).

I'm not sure how this is an improvement over what we currently have but I'm not familiar enough with the new process (and I don't even know if it's been enacted yet (seems like it hasn't since https://github.com/kubernetes-incubator is empty)). I give this the rubber stamp.

thockin:
the "except contrib/" is about the existing `kubernetes/kubernetes/contrib`
rather than `kubernetes/contrib`.  Parse that! :)

mikedanese:
The quote also says main repo, which this code was not in. When it was in the main repo, it was in contrib/,

thockin:
OK, then I have no idea what the "except contrib" means in this context.  I
can't copy it to my personal repo because it's not really mine.  Whatever,
done for now.

mikedanese:
SGTM

PR #20, opened 2016-10-28
Use build template

### LINKED ISSUE #10: Use alpine as a base
I've tried alpine and size is reduced by ~100 mb but the mv version that comes with it doesn't have `--no-target-directory` option
Inputs?

thockin:
Faster builds and versions from tags.  Also use alpine as a base image.  We  need alpine for other architectures...

Fixes #10 

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/git-sync/20)


thockin:
This builds on #19, so we should merge that first

thockin:
I forgot about the issue in #10 - will have to think about it.


DO NOT MERGE

thockin:
I think I can just install coreutils.  Building now.

stp-ip:
For me tests fail as the git-sync location isn't changed within the test_e2e.sh.
Should move from ```/bin/git-sync-amd64``` to ```/bin/amd/git-sync```

thockin:
Fixed. I'll also update the test to run the docker image.

Makefile
```diff
@@ -12,82 +12,150 @@
...
+# TODO: get a baseimage that works for other architectures
+# arm arm64 ppc64le
 
-container: .container-$(ARCH)
-.container-$(ARCH): bin/$(BIN)-$(ARCH)
-	docker build -t $(IMAGE):$(TAG) --build-arg ARCH=$(ARCH) .
+# Set default base image dynamically for each arch
 ifeq ($(ARCH),amd64)
-	docker tag $(IMAGE):$(TAG) $(LEGACY_AMD64_IMAGE):$(TAG)
+    BASEIMAGE?=alpine
```
stp-ip: Should we really not version alpine here? Probably better to be explicit with
```
    BASEIMAGE?=alpine:3.4
```

stp-ip:
Yeah coreutils would fix #10, but add ~6mb. Probably worth it considering shaving off almost 100 due to moving to alpine.
Awesome work.

Makefile
thockin: done

thockin:
Test runs the docker image now.  PTAL

Makefile
```diff
@@ -12,82 +12,150 @@
...
+
+bin/$(ARCH)/$(BIN): build-dirs
+	@echo "building: $@"
+	@docker run                                                            \
+	    -ti                                                                \
+	    -u $$(id -u):$$(id -g)                                             \
+	    -v $$(pwd)/.go:/go                                                 \
+	    -v $$(pwd):/go/src/$(PKG)                                          \
+	    -v $$(pwd)/bin/$(ARCH):/go/bin                                     \
+	    -v $$(pwd)/bin/$(ARCH):/go/bin/linux_$(ARCH)                       \
```
stp-ip: This line seems unnecessary.

thockin:
It is needed when building other architectures.  I have a template Makefile
that this derives from, which is used in a bunch of other kube repos and
subdirs :)

thockin:
my desktop is still 1.9 - I'll have to upgrade :)

thockin:
hmm, no upgrade available from corp...

test_e2e.sh
```diff
@@ -49,8 +49,15 @@ function assert_file_eq() {
 
 # Build it
 echo "Building..."
-make >/dev/null
-GIT_SYNC=./bin/git-sync-amd64
+make container REGISTRY=e2e TAG=$(make version) >/dev/null
```
stp-ip: For me the $(make version) is not silenced and therefore is assigned "Entering".
Would use $(make -s version) instead.

test_e2e.sh
```diff
@@ -49,8 +49,15 @@ function assert_file_eq() {
...
+function GIT_SYNC() {
+    docker run \
+        -i \
+        -u $(id -u):$(id -g) \
+        -v "$DIR":"$DIR" \
+        e2e/git-sync-amd64:$(make version) \
```
stp-ip: Same here. Would use $(make -s version) instead.

thockin:
Fixed in new push, passes e2e.

TaiSHiNet:
Awesome!! Please let me know when you push this to grc, I'll grab the tag (tried v2.0.3 already and nope)
Thanks for your work!

stp-ip:
We'll have to wait for thockin on the push. But I assume this will happen soon. Will let you know.

thockin:
Pushed to GCR

PR #25, opened 2016-12-19
Remove chmod command since we can specify permissions for a Secret Volume

ApsOps:
- Fixed because of https://github.com/kubernetes/kubernetes/pull/28936

@thockin Should we just suggest the user to set correct permissions, or should we fail unless correct permissions are set?


stp-ip:
LGTM from my side.

thockin:
LGTM

ApsOps:
@thockin What's the release process here? I mean, when's the next release due? 😄

thockin:
I'm chewing on how to do it more automatically.  In the mean time:
https://github.com/kubernetes/git-sync/releases/tag/v2.0.4

PR #29, opened 2017-01-26
clean up Dockerfile, remove ssh-wrapper.sh and set env var

### LINKED ISSUE #28: Use env var `GIT_SSH_COMMAND` instead
The ssh-wrapper seems unnecessary to me, as the command can be set as: 
`GITHUB_SSH_COMMAND="ssh -q -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i /etc/git-secret/ssh"`

ideahitme:
Fixes #28


ideahitme:
I signed it!

stp-ip:
LTGM

stp-ip:
One thing might be the possibility to configure the ssh secret location. As that's not the current feature set it can be done in a new PR, if that sounds reasonable to implement.

ideahitme:
@stp-ip I don't really see any value for that at this point

stp-ip:
Agreed, that's why I suggested separate PR, when it makes sense.

Waiting for @thockin for a final merge.

thockin:
LGTM

thockin:
Hmm, it looks like I never set up Travis for this.  :(

Issue #8, opened 2016-10-04
Subsequent syncs fail due to destination path not being empty


stp-ip:
During some initial tests following error came up:
```text error syncing repo: error running command: exit status 128: "fatal: destination path '/etc/source' already exists and is not an empty directory.\n"```
This seems to only happen on subsequent sync runs. Before the directory is empty and syncs successfully.

Some ENVs for a better context:
        - name: GIT_SYNC_BRANCH
          value: master
        - name: GIT_SYNC_SSH
          value: "true"
        - name: GIT_SYNC_ONE_TIME
          value: "false"
        - name: GIT_SYNC_WAIT
          value: "2"
        - name: GIT_SYNC_ROOT
          value: /etc/source/
        - name: GIT_SYNC_DEST
          value: /etc/source/files

TaiSHiNet:
Running into the same issue

stp-ip:
Just as a sidenote the current workaround we use is to set GIT_SYNC_ONE_TIME to true and add a command "./git-sync && sleep 999d".
This prevents further syncs, but that is sorta acceptable in our current use case, as we deploy often enough.

TaiSHiNet:
@stp-ip instead of the sleep, you could use pod inits:
http://kubernetes.io/docs/user-guide/production-pods/#handling-initialization

stp-ip:
Yeah that works, if you wanna do only the initial sync. We actually want ongoing sync, but disabled it with GIT_SYNC_ONE_TIME and sleep due to the above bug.

thockin:
DEST should be a bare name, "files" not "/path/to/files".  2.0.1 coming up :)

thockin:
v2.0.1 is pushed with better help and an error in this case.  PTAL

stp-ip:
That's there GIT_SYNC_ROOT comes in. Root basically should be used for the path and DEST should be used as the last folder name. Thanks for clarifying.

stp-ip:
v2.0.1 is not showing up yet. See: https://gcr.io/v2/google_containers/git-sync/tags/list

thockin:
Curiously it pushed git-sync-amd64 but not git-sync - will repush

KingJ:
Would it be possible to update the [demo config](https://github.com/kubernetes/git-sync/blob/master/demo/config/pod.yaml) to reflect this? It uses `GIT_SYNC_DEST` as a path rather than bare name.

thockin:
[...]

PR #35, opened 2017-02-15
Use bare name for DEST in example

stp-ip:
Using a path for DEST will fail in subsequent syncs #8.
Using a bare name in examples.


stp-ip:
I signed it!

...again

stp-ip:
I signed it!

...

stp-ip:
Too small a change for a review. Sorry for pinging thockin.

thockin:
LGTM

Issue #34, opened 2017-02-02
Load key "/etc/git-secret/ssh": Permission denied


dohnto:
Hello, 

I am using this version of git sync `1a9138765af75007d88f77c985f4f2af200b1227`
I build a docker image using `make container REGISTRY=X TAG=Y`

And I am using following setup:
```
apiVersion: v1
kind: Pod
metadata:
  name: server
spec:
  containers:
  - image: nginx
    name: nginx
    volumeMounts:
    - mountPath: /mypath
      name: git-volume
  - image: MY_IMAGE:git-sync-amd64:v2.0.4-6-g1a91387
    name: git-sync
    env:
    - name: GIT_SYNC_REPO
      value: "git@MY_REPO"
    - name: GIT_SYNC_SSH
      value: "true"
    #command:
    #- tail
    #- -f
    #- /dev/null
    args:
    - -dest
    - foo
    - --v=9
    volumeMounts:
    - mountPath: /etc/git-secret
      name: git-secret
    - name: git-volume
      mountPath: /git
  volumes:
  - name: git-secret
    secret:
      secretName: creds
      defaultMode: 256
  - name: git-volume
    emptyDir: {}
```

But I cannot clone the repository.

When I use the commend version with `tail -f /dev/null` and execute into the container and I run manually:

```bash
# export GIT_SSH_COMMAND="ssh -vv -q -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i/etc/git-secret/ssh"
# git clone git@MY_REPO
...
Load key "/etc/git-secret/ssh": Permission denied
...
# ls -l /etc/git-secret/ssh
lrwxrwxrwx 1 root root 10 Feb  2 16:21 /etc/git-secret/ssh -> ..data/ssh
# -l /etc/git-secret/..data/ssh 
-r-------- 1 root root 1675 Feb  2 16:21 /etc/git-secret/..data/ssh
# whoami 
nobody
```

sjernigan:
We had the same issue.  Eventually worked around it by running the container as root :-(

        securityContext:
          runAsUser: 0

stp-ip:
Fixed via #37.

Issue #33, opened 2017-02-02
Issue cloning from Github with SSH


ursuad:
Because of #32 I tried switching to SSH, but the container keeps crashing again. 

The logs are:
```
$ kubectl logs dp-analytics-blog-2fjnv -c git-sync
I0202 13:28:42.608581       1 main.go:162] starting up: ["/git-sync"]
E0202 13:28:43.307843       1 main.go:169] error syncing repo: error running command: exit status 128: "Cloning into '/git'...\nfatal: Could not read from remote repository.\n\nPlease make sure you have the correct access rights\nand the repository exists.\n"
```

Just to test, I replaced the container command with:
```
#        command: ["cat"]
#        args: ["/etc/git-secret/ssh"]
```
I got the SSH key from the output, removed all my local keys from the SSH agent, imported that key (from the container) with `ssh-add` and it worked to clone the repo.  So it must be something from the container?

My config:
```
apiVersion: v1
kind: ReplicationController
metadata:
  name: my-rc-name
spec:
  replicas: 1
  template:
    metadata:
      name: my-pod-name
      labels:
        application: my-app
        component: my-comp
        environment: my-env
    spec:
      containers:
      - name: git-sync
        image: gcr.io/google_containers/git-sync:v2.0.4
#        command: ["cat"]
#        args: ["/etc/git-secret/ssh"]
        imagePullPolicy: Always
        volumeMounts:
        - name: git-secret
          mountPath: /etc/git-secret
        env:
        - name: GIT_SYNC_REPO
          value: "git@github.com:GITHUB_USER/GITHUB_REPO.git"
        - name: GIT_SYNC_SSH
          value: "true"
        - name: GIT_SYNC_BRANCH
          value: master
        - name: GIT_SYNC_DEST
          value: git
        - name: GIT_SYNC_DEPTH
          value: "1"
      volumes:
      - name: html
        emptyDir: {}
      - name: git-secret
        secret:
          secretName: github-secret-ssh
          defaultMode: 0400
```

stp-ip:
To "fix" it try running the container with "runAsUser: 0". Still looking into why this works or better why running it without it fails.

Also the recommended container to run is: "gcr.io/google_containers/git-sync-amd64:v2.0.4"

stp-ip:
Just to gather some more info:

What cluster version did you run?
Where did the cluster run? GKE?
Was it an updated version of the cluster?
What was the actual permission on the ssh key?

sjernigan:
[...]

PR #37, opened 2017-02-17
Update ssh.md

sjernigan:
Adding requirement for runAsRoot and suggested configuration for defaultMode.  These are required after previous change to remove the chmod on the key.

See disccussion #33 and #34 #

### DIFF
--- a/docs/ssh.md
+++ b/docs/ssh.md
@@ -40,7 +40,8 @@ volumes: [
     {
         "name": "git-secret",
         "secret": {
-          "secretName": "git-creds"
+          "secretName": "git-creds",
+          "defaultMode": 256
         }
     },
     ...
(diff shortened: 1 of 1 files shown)


stp-ip:
Afaik the runAsRoot shouldn't be necessary, when the credentials are modified via defaultMode. During my investigation it seemed as defaultMode has no effect in any of my clusters.
Will investigate further and see, if this PR makes sense to merge as is.

If you want to pinpoint the details, could you check, if defaultMode does in fact change the permissions on your clusters?

sjernigan:
I signed the CLA

sjernigan:
With the changes suggested (runAsRoot and defaultMode), it works.  If I remove the defaultMode only, I get 
"ERROR: can't configure SSH: Permissions -rw-r--r-- for SSH key are too open. It is recommeded to mount secret volume with `defaultMode: 256` (decimal number for octal 0400)."

sjernigan:
Perhaps it should also be noted in the docs that it will fail to pull a public repo via SSH if the SSH key config isn't correct.  However, you can use the https form in GIT_SYNC_REPO

sjernigan:
Finally, the defaultMode change by itself also doesn't work.  The SSH key is owned by root despite the container user being set to nobody in the dockerfile.  Seems like there should be a user option in volumeMount but I didn't find one.

stp-ip:
Thanks for checking it out. Yeah I assumed that the permission change is somehow affected by the container running as nobody. Will merge this for now and open an issue to try fixing it. I don't like running containers as root.

Issue #15, opened 2016-10-27
git-sync not working with Hugo. 


ursuad:
I tried the demo setup and played around a bit with the files. 

There seems to be an issue with the `git-sync` writing the checked out files resulting in a `SystemEvent: REMOVE` in Hugo resulting in an empty website. Manually editing the files from the `git-sync` container works.

My current setup is with Hugo `--verboseLog=true` enabled. 

Steps taken:
1) Have a git repo with a few `.md` files
2) Launch the pod
**Expected**
The website is displayed with all the pages
**Actual result**
Just as expected
**Hugo Logs**

```
$ kubectl logs my-hugo-pod-kgijd hugo -f
+ '[' '!' -d /src/git/site/themes ']'
+ ln -s /themes /src/git/site/themes
++ eval echo server '--source=${HUGO_SRC}' '--theme=${HUGO_THEME}' '--buildDrafts=${HUGO_BUILD_DRAFT}' '--baseUrl=${HUGO_BASE_URL}' --watch '--destination=${HUGO_DEST}' --appendPort=false --log --verbose '--verboseLog=${HUGO_VERBOSE_LOG}'
+++ echo server --source=/src/git/site --theme=hugo-theme-bootstrap4-blog --buildDrafts=true --baseUrl=minion-address:30258 --watch --destination=/dest --appendPort=false --log --verbose --verboseLog=true
+ hugo server --source=/src/git/site --theme=hugo-theme-bootstrap4-blog --buildDrafts=true --baseUrl=minion-address:30258 --watch --destination=/dest --appendPort=false --log --verbose --verboseLog=true
INFO: 2016/10/26 23:58:42 hugo.go:405: Using config file: /src/git/site/config.toml
INFO: 2016/10/26 23:58:42 hugo.go:517: using a UnionFS for static directory comprised of:
INFO: 2016/10/26 23:58:42 hugo.go:518: Base: /src/git/site/themes/hugo-theme-bootstrap4-blog/static
INFO: 2016/10/26 23:58:42 hugo.go:519: Overlay: /src/git/site/static/
INFO: 2016/10/26 23:58:42 hugo.go:551: syncing static files to /dest/
Started building sites ...
INFO: 2016/10/26 23:58:42 site.go:1489: found taxonomies: map[string]string{"tag":"tags", "category":"categories"}
INFO: 2016/10/26 23:58:42 htmlredirect.go:116: Alias "/tags/themes/page/1" translated to "tags/themes/page/1/index.html"
[...]

PR #45, opened 2017-03-22
[WIP] Document atomic switching and related issues

stp-ip:
Symlinks as used for the atomic switch might lead to issues
in various applications.

Closes: #15

### DIFF
--- a/docs/atomic.md
+++ b/docs/atomic.md
@@ -1,0 +1,6 @@
+# Atomic updates
+By design git-sync uses symlinks to atomically switch git content between versions.
+Various applications might not work well with symlinks in general.
+Symlink changes might not be recognized as an update to the content. (inotify might be configured to follow symlinks instead of watching for a symlink update)
+
+To disable the atomic switch via symlinks set ```GIT_SYNC_ATOMIC``` to false.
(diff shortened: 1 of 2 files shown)


stp-ip:
Seems like the Atomic switching option isn't available currently. Needs to be introduced back.

docs/atomic.md
```diff
@@ -0,0 +1,6 @@
+# Atomic updates
+By design git-sync uses symlinks to atomically switch git content between versions.
```
thockin: ... This means that your application will always have a consistent view of a single git revision.  When a new revision is available, the entire contents are pulled to a staging area and then the symlink is changed, at which point your application will see only the new revision.

docs/atomic.md
```diff
@@ -0,0 +1,6 @@
...
+Various applications might not work well with symlinks in general.
+Symlink changes might not be recognized as an update to the content. (inotify might be configured to follow symlinks instead of watching for a symlink update)
+
+To disable the atomic switch via symlinks set ```GIT_SYNC_ATOMIC``` to false.
```
thockin: ... This means that your application may see git's intermediate state as new revisions are being pulled.

stp-ip:
Close for now.

PR #51, opened 2017-03-22
Update copyright

stp-ip:


thockin:
In general we don't want to update copyright - it should be the FIRST date it was copyrighted.

stp-ip:
Sorry. Was one of my automation tools on the client-side and this made it too easy to submit. Reverted and will disable my tooling for git-sync.

PR #57, opened 2017-03-31
Avoid zombie processes

aledbf:


thockin:
This is a lot of dep for such little work.  I want to say we should just do it ourselves, but I am having trouble making it work.  It seems different docker versions do very different things.

This should work, shouldn't it?  I don't have time to cross-check docker versions, but 17.03.1-ce seems to wrap it in an `sh -c`, even if I ENTRYPOINT it...

```
    signal_chan := make(chan os.Signal, 1)
    signal.Notify(signal_chan, syscall.SIGTERM, syscall.SIGCHLD)

    go func() {
        for {
            s := <-signal_chan
            switch s { 
            case syscall.SIGTERM:
                fmt.Println("SIGTERM")

            case syscall.SIGCHLD:
                fmt.Println("SIGCHLD")
                pid, _ := syscall.Wait4(-1, nil, syscall.WNOHANG, nil)
                fmt.Println("TIM:", pid)

            default:
                fmt.Println("Unknown signal.")
            }
        }
    }() 
```

thockin:
```
func main() {
    signal_chan := make(chan os.Signal, 1)
    signal.Notify(signal_chan,
        syscall.SIGTERM,
        syscall.SIGCHLD)

    go func() {
        for {
            s := <-signal_chan
            switch s {
            case syscall.SIGTERM:
                fmt.Println("SIGTERM")

            case syscall.SIGCHLD:
                fmt.Println("SIGCHLD")
                for {
                    pid, _ := syscall.Wait4(-1, nil, syscall.WNOHANG, nil)
                    fmt.Println("TIM:", pid)
                    if pid <= 0 {
                        break
                    }
                }

            default:
                fmt.Println("Unknown signal.")
            }
        }
    }()

    run("/bin/sh", "-c", "cat /dev/zero &")
    time.Sleep(time.Second)
    run("/bin/sh", "-c", "sleep 15&")
    time.Sleep(time.Second)
    run("/bin/sh", "-c", "sleep 15&")
    time.Sleep(time.Second * 60)
}

func run(name string, args ...string) {
    exec.Command(name, args...).Run()
}
```

Seems to work as long as the process is actually the entrypoint, and if it
is not, it just gets -1 from waitpid().

aledbf:
@thockin and this code?
```go
package main

import (
	"fmt"
	"os"
	"os/exec"
	"os/signal"
	"sync/atomic"
	"syscall"
	"time"
)

func main() {
	var counter uint64
	callback := func(pid int, wstatus syscall.WaitStatus) {
		fmt.Printf("pid %d, wstatus: %+v\n", pid, wstatus)
		atomic.AddUint64(&counter, 1)
	}

	sig := make(chan os.Signal, 1024)
	signal.Notify(sig, syscall.SIGCHLD)
	go reapChildren(sig, callback)

	for i := 1; i < 31; i++ {
		go func(s int) {
			exec.Command("/bin/bash", "-c", fmt.Sprintf("sleep %v", s)).Run()
		}(i)
	}

	time.Sleep(time.Second * 40)
	fmt.Printf("%v\n", atomic.LoadUint64(&counter))
}

func reapChildren(signal chan os.Signal, callback func(int, syscall.WaitStatus)) {
	for {
		<-signal
		for {
			var status syscall.WaitStatus
			pid, _ := syscall.Wait4(-1, &status, 0, nil)
			if pid <= 0 {
				break
			}
			callback(pid, status)
			break
		}
	}
}
```

thockin:
They are pretty much the same, no?   You really need to loop on the wait call because POSIX doesn't guarantee queue-like behavior for signals, so if another SIGCHLD comes in at the wrong time, you could miss it.  Also you really should only do this for SIGCHLD.  Once you add thise, we have basically the same code

PR #58, opened 2017-04-21
Use secure defaults (check hostkeys)

pieterlange:
Somewhat related: https://bridge.grumpy-troll.org/2017/04/golang-ssh-security/

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -66,6 +66,8 @@ var flPassword = flag.String("password", envString("GIT_SYNC_PASSWORD", ""),
 
 var flSSH = flag.Bool("ssh", envBool("GIT_SYNC_SSH", false),
 	"use SSH for git operations")
+var flSSHKnownHosts = flag.Bool("ssh-known-hosts", envBool("GIT_KNOWN_HOSTS", false),
+	"enable SSH known_hosts verification")
 
 var log = newLoggerOrDie()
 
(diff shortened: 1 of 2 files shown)


pieterlange:
Signed it.

thockin:
This is a breaking change, if I understand it correctly.  It has to be opt-in because we can not just arbitrarily break users, unless we want to do a full major version bump and require this.

pieterlange:
How is it acceptable to not check host keys? Why was this done not done right in the first place?

I understand that it is a "breaking change" but isn't the security benefit worth forcing people to wake up and fix their configuration? If this requires a major version bump to fix then maybe that's what needs to be done.

thockin:
@stp-ip thoughts?

stp-ip:
I would be ok with a breaking change in general, but as the permission stuff might also require that, we might wanna postpone it and currently just ad a documentation warning that it uses the unstrict checking.
Sidenote the permission stuff is sort of blocked until this issue is closed: https://github.com/kubernetes/kubernetes/issues/34982

thockin:
I'm OK to do a major version bump, but we should get this in as a flag and
then flip the default in the major bump.

pieterlange:
Is this better? @thockin @stp-ip 

If this can be written shorter feel free to replace it.

cmd/git-sync/main.go
```diff
@@ -66,6 +66,8 @@ var flPassword = flag.String("password", envString("GIT_SYNC_PASSWORD", ""),
 
 var flSSH = flag.Bool("ssh", envBool("GIT_SYNC_SSH", false),
 	"use SSH for git operations")
+var flSSHKnownHosts = flag.Bool("ssh-knownhosts", envBool("GIT_KNOWN_HOSTS", false),
```
thockin: nit: known-hosts ?

pieterlange: Sure!

thockin:
Thanks!

PR #66, opened 2017-09-15
Update RELEASING doc a bit

thockin:


thockin:
I'm going to self-merge this on the presumption of being trivial.

Issue #59, opened 2017-04-26
Reminder to flip the default host checking to 'true' on 3.0 version bump


pieterlange:
Ref: #58

pieterlange:
@stp-ip can we cut a new release so people get secure defaults? i think that warrants a version bump anyway.

stp-ip:
@thockin is the person in charge of releases.

thockin:
I have cut a 2.0 release branch, so master can take 3.0 changes.

pieterlange:
Thanks @thockin!

PR #69, opened 2017-09-15
Flip the known_hosts enforcement to true so the git hosts actually get verified.

pieterlange:
Ref: #59 
Ref: #58


stp-ip:
LGTM

thockin:
Do we need to document this?  What is the failure mode if it is not specified - how will users of v2 experience this when they switch to v3?

Issue #71, opened 2017-09-19
e2e is broken on MacOS - pkill failure


jorianvo:
**OS:** Mac OS X Sierra 10.12.6
**Make version:** GNU Make 3.81
**Steps:** make test
**Error:**
```shell
$ make test
Running tests:
ok  	k8s.io/git-sync/cmd/git-sync	0.005s
?   	k8s.io/git-sync/pkg/version	[no test files]

Checking gofmt: PASS

Checking go vet: PASS

building: bin/amd64/git-sync
container: e2e/git-sync-amd64:v2.0.5-13-g267a78b-dirty
test root is /tmp/git-sync-test.28622
Initialized empty Git repository in /private/tmp/git-sync-test.28622/repo/.git/
testcase head-once: PASS
testcase default-sync: make: *** [test] Error 1
```

The issue seems to be the `pkill git-sync` it is unable to match a process and returns 1. The container remains also running:
```
$ docker ps
CONTAINER ID        IMAGE                                         COMMAND                  CREATED             STATUS              PORTS                    NAMES
0e6fe071a1ef        e2e/git-sync-amd64:v2.0.5-13-g267a78b-dirty   "/git-sync --logto..."   3 minutes ago       Up 3 minutes                                 modest_jennings
```
And is matched by `ps`:
```
$ ps auxw | grep git-sync
user 26736   0.0  0.0  2451236    828 s003  S+    8:52PM   0:00.00 grep --color=auto --exclude-dir=.bzr --exclude-dir=CVS --exclude-dir=.git --exclude-dir=.hg --exclude-dir=.svn git-sync
user 26525   0.0  0.1 556638724   9780 s003  S     8:47PM   0:00.40 docker run -i -u 501:20 -v /tmp/git-sync-test.28622:/tmp/git-sync-test.28622 e2e/git-sync-amd64:v2.0.5-13-g267a78b-dirty --logtostderr --v=5 --wait=0.1 --repo=/tmp/git-sync-test.28622/repo --root=/tmp/git-sync-test.28622/root --dest=link
```

I actually want to work on issue #54, so I thought I first get all tests working, but if irrelevant I'm happy to drop this issue

jorianvo:
This issue is fixed by #73

PR #73, opened 2017-09-21
Improve e2e tests

jorianvo:
This commit will fix issue #71 and is tested on both MacOS and Ubuntu.

### DIFF
--- a/test_e2e.sh
+++ b/test_e2e.sh
@@ -13,9 +13,8 @@ function testcase() {
 
 function fail() {
     echo "FAIL: " "$@"
-    sleep 2
-    pkill git-sync || true
-    ps auxw | grep git-sync
+    sleep 3
+    remove_sync_container || true
     exit 1
 }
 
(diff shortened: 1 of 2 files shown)


test_e2e.sh
```diff
@@ -124,25 +141,25 @@ GIT_SYNC \
...
-sleep 2
+sleep 3
 assert_link_exists "$ROOT"/link
 assert_file_exists "$ROOT"/link/file
 assert_file_eq "$ROOT"/link/file "$TESTCASE 1"
 # Move forward
 echo "$TESTCASE 2" > "$REPO"/file
 git -C "$REPO" commit -qam "$TESTCASE 2"
-sleep 2
+sleep 3
```
thockin: I tried to find a time that didn't flake but also wasn't an eternity.  Is this flaking for you?

thockin:
Thanks!  Test PRs are the best PRs!

test_e2e.sh
jorianvo: Not anymore, using the three second time-out it is stable

PR #84, opened 2018-02-13
add usage docs

arturozv:
Hi, thanks for the project! 
I've been having to check the source code to check what options does the cli have. It would be nice to have them on a more friendly format

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -1,0 +1,24 @@
+# Usage:
+| Command | Environment | Description |
+| --- | --- | --- |
+| -alsologtostderr | | log to standard error as well as files |
+| -branch string | GIT_SYNC_BRANCH | the git branch to check out (default "master") |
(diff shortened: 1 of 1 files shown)


arturozv:
CLA done

stp-ip:
The CLA is still having a problem as the username on the commit is different than the one from the commit.

thockin:
How do we ensure this is kept up to date?

Would you consider adding a step to the tests to regenerate it and diff, and fail if the diff is not empty?

We should set up Travis for this, too, but that's not your problem.

stp-ip:
Automating that would be great, but we could start by asking for an update on PRs adding new features.

thockin:
I promise this will go stale.  Did you run a script to generate this?

arturozv:
@thockin I did not use a script, if I remember well I took the output from the help command, fix some default values that were set up for my specific configuration and compare to the code to add some missing ones not specific to git-sync like the loglevel

PR #87, opened 2018-03-29
Run Go build-image without TTY

mandrean:
AFAIK it's not necessary to run the build-image with TTY and the `-t` flag can just be omitted as it breaks some CI builds trying to run `make container` when TTY isn't allowed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,7 +85,7 @@ build: bin/$(ARCH)/$(BIN)
 bin/$(ARCH)/$(BIN): build-dirs
 	@echo "building: $@"
 	@docker run                                                            \
-	    -ti                                                                \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -85,7 +85,7 @@ build: bin/$(ARCH)/$(BIN)
 bin/$(ARCH)/$(BIN): build-dirs
 	@echo "building: $@"
 	@docker run                                                            \
-	    -ti                                                                \
+	    -i                                                                 \
```
thockin: what about `make test` ?  Should we update that too?

thockin:
will merge this - followup with test?

mandrean:
@thockin True! Good catch. Same there.

PR #88, opened 2018-03-30
Add debug, enable pprof by flag

thockin:


cmd/git-sync/main.go
```diff
@@ -180,6 +193,7 @@ func main() {
 		}
 		if initialSync {
 			if *flOneTime {
+				log.V(2).Infof("synced one time, exiting", *flRev)
```
alarsyo: This line doesn't seem to print a space between `"synced one time, exiting"` and `*flRev`

alarsyo: This line doesn't seem to print a space between `"synced one time, exiting"` and `*flRev`

stp-ip:
Besides the small whitespace issue, from a code perspective I could see it merged, but from a general perspective...
We are embedding net/http just for pprof. Size difference is 2.5MB vs 8.8MB.
Not just the size increase, but adding a web server into a small sync tool feels quite a stretch. It's a debugging tool and might not be used often in production. I'm torn as I feel the added baggage is too much and not warranted by the debugging usage. An alternative would be to have a debugging container build with pprof enabled.

So with that considered, I would hold this for now.

PR #92, opened 2018-05-17
Add gitcookie configuration option

brycecr:
Add an option to specify a git http cookie file in the k8s yaml
that will be configured as the global git cookie file in the git-sync
client.

This is a useful access mechanism for some some hosted repositories
where SSH and passwords are not easily available, such as Google Cloud
Source Repositories.

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -69,6 +69,9 @@ var flSSH = flag.Bool("ssh", envBool("GIT_SYNC_SSH", false),
 var flSSHKnownHosts = flag.Bool("ssh-known-hosts", envBool("GIT_KNOWN_HOSTS", true),
 	"enable SSH known_hosts verification")
 
+var flCookieFile = flag.Bool("cookie-file", envBool("GIT_COOKIE_FILE", false),
+	"use git cookiefile")
+
 var log = newLoggerOrDie()
 
 func newLoggerOrDie() logr.Logger {
@@ -156,6 +159,13 @@ func main() {
 	if *flSSH {
 		if err := setupGitSSH(*flSSHKnownHosts); err != nil {
 			fmt.Fprintf(os.Stderr, "ERROR: can't configure SSH: %v\n", err)
+			os.Exit(1)
+		}
+	}
+
+	if *flCookieFile {
+		if err := setupGitCookieFile(); err != nil {
+			fmt.Fprintf(os.Stderr, "ERROR: can't set git cookie file: %v\n", err)
 			os.Exit(1)
 		}
 	}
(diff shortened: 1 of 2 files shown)


brycecr:
Following up, I have signed the CLA

brycecr:
I have also added the commit's authoring email address to my account, so I think check about that should be fixed at this point too. I'm not quite sure how to re-run that check?

cmd/git-sync/main.go
```diff
@@ -69,6 +69,9 @@ var flSSH = flag.Bool("ssh", envBool("GIT_SYNC_SSH", false),
 var flSSHKnownHosts = flag.Bool("ssh-known-hosts", envBool("GIT_KNOWN_HOSTS", true),
 	"enable SSH known_hosts verification")
 
+var flCookieFile = flag.Bool("cookie-file", envBool("GIT_COOKIE_FILE", false),
```
thockin: Is this compatible with or mutually-exclusive with the other modes?  E.g. should we throw an error if the user specifies both cookie-file and username ?

cmd/git-sync/main.go
```diff
@@ -503,3 +513,22 @@ func setupGitSSH(setupKnownHosts bool) error {
 
 	return nil
 }
+
+func setupGitCookieFile() error {
+	log.V(1).Infof("configuring git cookie file")
+
+	var pathToCookieFile = "/etc/git-secret/cookie_file"
```
thockin: why is this path hard-coded?  Why not make this the value of the flag?

brycecr: I am following the pattern used for the ssh secrets. These are paths within the git-sync container, so the host path would be handled on the docker/k8s configuration end.

If we would like git-sync to provide a more flexible interface here (for non container usage, I would think, as I don't see the advantage of flexible paths within the container), that seems fine, but we should unify the treatment between the cookie-file and the ssh secrets, I would think?

cmd/git-sync/main.go
thockin: so maybe we should validate that we don't do things that are confusing?

cmd/git-sync/main.go
thockin: blech.  Volumes are weak in this area.  the dir /etc/git-secret can only be used by a single secret, so either this goes into docs alongside the SSH secret or you pick a different dir (if we want them to work together).  My guess is that they should not work together, so that restriction is OK?

Need to write a doc, please.

cmd/git-sync/main.go
brycecr: I added an error if GIT_SYNC_SSH and any of password, username, or cookie-file are set.

PR #96, opened 2018-06-05
Update kubernetes demo/config for example

mattjmcnaughton:
A couple of changes to make the kubernetes config work
and follow best practices.

- Use `deployment` instead of `pod`.
- Use correct image names.

After these changes, running `kubectl create -f demo/config/` creates a
working deployment/service.


demo/README.md
```diff
@@ -20,11 +22,14 @@ docker build -t <some-registry>/hugo hugo/
 docker push <some-registry>/hugo <some-registry>/git-sync
 ```
 
+If you end up using a different registry, be sure to update `config/pod.yaml`.
```
stp-ip: s/pod/deployment/

demo/config/deployment.yaml
```diff
@@ -0,0 +1,57 @@
...
+    matchLabels:
+      name: blog
+  template:
+    metadata:
+      labels:
+        name: blog
+    spec:
+      containers:
+      - name: git-sync
+        image: staging-k8s.gcr.io/git-sync-amd64:v2.0.5
```
stp-ip: The latest version would be v2.0.6. Any reason why you used .5?

mattjmcnaughton: Nope, no reason! Other than I think I thought 2.0.5 was the most recent version haha...

demo/README.md
mattjmcnaughton: Great catch :)

mattjmcnaughton:
@stp-ip thanks for the feedback - I made your changes so this is ready for another look :)

mattjmcnaughton:
Yup!

PR #98, opened 2018-06-19
Add chown flag

TimBozeman:
Hiya,

Thank you for git-sync. It's pretty awesome! I just signed the CLA. Can we add a chown flag please?

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -58,6 +58,8 @@ var flMaxSyncFailures = flag.Int("max-sync-failures", envInt("GIT_SYNC_MAX_SYNC_
 	"the number of consecutive failures allowed before aborting (the first pull must succeed)")
 var flChmod = flag.Int("change-permissions", envInt("GIT_SYNC_PERMISSIONS", 0),
 	"the file permissions to apply to the checked-out files")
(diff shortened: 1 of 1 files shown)


TimBozeman:
CLA check...

thockin:
Can you describe the use-case for this in the PR message?  E.g. why not run it as the UID you want?

TimBozeman:
I'm using that new GIT_COOKIE_FILE flag which needs the RunAsUser 0 to be able to write the config file, but I also need the files writable by the web server.

thockin:
Why does GIT_COOKIE_FILE need to RunAsUser 0 ?  Cookiefile is not set to write-back, I think...

thockin:
(sorry for letting this languish, just want to understand)

TimBozeman:
I believe if you don’t RunAsUser 0 it tries to write to a git config file and fails cause it doesn’t have permissions.

thockin:
Aha!  Maybe we can just fix that.  Let me look at it today.  Not that I
don't want a patch from you, but I am worried that this is racy.

thockin:
Can you test something for me?

Add an env var to git-sync - set `HOME`=`/tmp`.  If that works for you I
will send a PR to do that by default.

thockin:
This will be fixed by default in the next git-sync.  Thanks for bringing it
to my attention.

PR #100, opened 2018-07-03
Update images and allow non-root use of gitcookie

thockin:
@TimBozeman re #98


stp-ip:
LGTM

thockin:
Setting HOME confirmed to fix gotcookie issues as non-root.

PR #101, opened 2018-07-11
Rename variable wait to interval

AndersonChoi:
The interval seems to be a better word than wait.


AndersonChoi:
@thockin thanks for the feedback - I made your changes so this is ready for another look :)

cmd/git-sync/main.go
```diff
@@ -50,8 +50,11 @@ var flRoot = flag.String("root", envString("GIT_SYNC_ROOT", envString("HOME", ""
 	"the root directory for git operations")
 var flDest = flag.String("dest", envString("GIT_SYNC_DEST", ""),
 	"the name at which to publish the checked-out files under --root (defaults to leaf dir of --repo)")
+// Deprecated: flWait is deprecated. Use flInterval instead.
```
thockin: Can you also add "(deprecated, see --interval)" to the string description?

cmd/git-sync/main.go
```diff
@@ -144,6 +147,16 @@ func main() {
 		flag.Usage()
 		os.Exit(1)
 	}
+	if *flOneTime == false && *flWait != 0 {
+		if *flInterval == 0 {
+			log.Errorf("wait is deprecated. use interval instead")
```
thockin: log.V(0).Infof please

thockin: "The --wait flag is deprecated, please use --interval instead."

cmd/git-sync/main.go
```diff
@@ -144,6 +147,16 @@ func main() {
...
+			*flInterval = *flWait
+		} else {
+			fmt.Fprintf(os.Stderr, "ERROR: use only one --interval or --wait. These cannot be used together.\n")
```
thockin: "ERROR: flags --wait and --interval are mutually exclusive" is sufficient

cmd/git-sync/main.go
```diff
@@ -144,6 +147,16 @@ func main() {
 		flag.Usage()
 		os.Exit(1)
 	}
+	if *flOneTime == false && *flWait != 0 {
```
thockin: Can you add a separate check - if --wait and --one-time are specified, it should be an error.  We can't make it an error (breaking change) but we can make it log.

cmd/git-sync/main.go
AndersonChoi: I got it.

cmd/git-sync/main.go
AndersonChoi: Thank you for comment. I added it.

AndersonChoi:
@thockin Thank you for your feedback. The details have been applied.

AndersonChoi:
Hello @thockin :) 
If you have any further changes to make, please let me know and I'll apply them. Thank you. Have a nice day.

test_e2e.sh
```diff
@@ -285,7 +285,7 @@ assert_file_exists "$ROOT"/link/file
 assert_file_eq "$ROOT"/link/file "$TESTCASE 1"
 # Wrap up
 remove_sync_container
-wait
+interval
```
thockin: These changes are wrong - `wait` here is a shell command.

AndersonChoi: I apologize for my mistake. 
I reverted command.

AndersonChoi:
Hi @thockin . I apologize for my mistake.
I reverted command.

thockin:
sorry, it seems you have a conflict with another PR that went it.

AndersonChoi:
I'm really sorry to resolve conflict source late. 
Would you check my pull request source please?
I greatly appreciate your review.

PR #102, opened 2018-07-30
GIT_SYNC_BRANCH accepts either a branch or a tag

odracci:
Currently git-sync uses GIT_SYNC_REV to check-out a tag or an hash, if
one-time is set is not necessary to lookup the hash of the tag because
is enough to use GIT_SYNC_BRANCH to clone the repository.
With this change is possible to clone a repository by a tag or a branch using
GIT_SYNC_BRANCH in case one-time is set to true.


thockin:
The alternative is just to set rev to HEAD, right?

thockin:
Could you add a test case to test_e2e.sh to show what works now that didn't before?

odracci:
@thockin added e2e test.
On `master` branch the test fails with the following error message:
```
E0815 22:30:18.676724       1 main.go:186] error syncing repo: error running command: exit status 128: "fatal: invalid reference: origin/tag-or-branch-once--TAG\n"
```

About your question:
> The alternative is just to set rev to HEAD, right?

What do you mean?

test_e2e.sh
```diff
@@ -129,6 +129,29 @@ assert_file_eq "$ROOT"/link/file "$TESTCASE"
...
+testcase "tag-once"
+TAG="$TESTCASE"--TAG
+# First sync
+echo "$TESTCASE" > "$REPO"/file
+git -C "$REPO" commit -qam "$TESTCASE"
+git -C "$REPO" tag -af "$TAG" -m "$TESTCASE" >/dev/null
+GIT_SYNC \
+    --logtostderr \
+    --v=5 \
+    --wait=0.1 \
```
thockin: don't need this for --one-time

odracci:
The main reason of this PR is to use `git-sync` with `airflow`, as you can see [here](https://github.com/apache/incubator-airflow/blob/48229328f28bfd1717006c5346ba64985a97935d/airflow/contrib/kubernetes/worker_configuration.py#L45-L60) Airflow uses `GIT_SYNC_BRANCH` to clone the repository.
With my changes is possible to clone the repo by tag or by branch.
An other reason is that in the context of automation is easier to use always one `option` otherwise is required to implement additional logic to recognise if the value passed is a tag or a branch and use the correct `option`.

thockin:
I get the point of simpler, but it only fixes one-time mode?  Can we fix it in general?

odracci:
In my opinion this fix would work only in one-time mode.
Sync a `tag` every 30 seconds doesn't make sense, it won't change the content.
In short if I want to sync the content of the repo periodically, the only option I can use is `branch`, if I use `rev `, one-time should be enforced to `true`. @thockin What do you think?

odracci:
ping @thockin

odracci:
@thockin any chance to have this PR merged?

thockin:
Hi, sorry for losing track of this.

While most people consider it "bad form" to change a tag, it is possible to do and I think we should support it.  we have an e2e test that covers it, for example.

Maybe the right thing to do is just to have the `--rev` flag accept branch, tag, or hash values (already does tag or hash), deprecate `--branch`, and document the autodetection mechanism along the lines of "whatever git does by default", and add some tests for it?

thockin:
See #213 for work on this topic

PR #108, opened 2018-08-22
Add `ssh-extra-args` argument for git & ssh

### LINKED ISSUE #107: GIT_SSH_COMMAND is overwritten
I'm specifically trying to add `-vvv` to the ssh command by setting `GIT_SSH_COMMAND=ssh -vvv` in my pod. PR #29 added behavior that sets `GIT_SSH_COMMAND` instead of using a wrapper, which make sense but make this impossible.

I think it would be nice if `GIT_SSH_COMMAND` could still be used to set flags since that's how google and stackoverflow suggest people enable more verbose logs. However, I'm not sure if it would make sense to try and merge the user defined set of arguments with the settings git-sync needs to configure, it seems risky and prone to weird errors.

It might be easier to just add a new option to git-sync for enabling verbose ssh commands.

Does anyone have opinions?

rlguarino:
This CL adds a new parameter to git-sync that lets users specify extra
arguments to pass to ssh when git is configured to clone over ssh. The
args are appended to the list of arguments that git-sync already
adds to the `GIT_SSH_COMMAND` environment variable.

Closes #107

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -72,6 +72,9 @@ var flSSHKnownHosts = flag.Bool("ssh-known-hosts", envBool("GIT_KNOWN_HOSTS", tr
 var flCookieFile = flag.Bool("cookie-file", envBool("GIT_COOKIE_FILE", false),
 	"use git cookiefile")
 
+var flSSHExtraArgs = flag.String("ssh-extra-args", envString("SSH_EXTRA_ARGS", ""),
+	"extra arguments to be passed to ssh command used by git")
+
 var log = newLoggerOrDie()
 
 func newLoggerOrDie() logr.Logger {
@@ -162,7 +165,7 @@ func main() {
 	}
 
 	if *flSSH {
-		if err := setupGitSSH(*flSSHKnownHosts); err != nil {
+		if err := setupGitSSH(*flSSHKnownHosts, *flSSHExtraArgs); err != nil {
 			fmt.Fprintf(os.Stderr, "ERROR: can't configure SSH: %v\n", err)
 			os.Exit(1)
 		}
@@ -498,7 +501,7 @@ func setupGitAuth(username, password, gitURL string) error {
 	return nil
 }
 
-func setupGitSSH(setupKnownHosts bool) error {
+func setupGitSSH(setupKnownHosts bool, extraArgs string) error {
 	log.V(1).Infof("setting up git SSH credentials")
 
 	var pathToSSHSecret = "/etc/git-secret/ssh"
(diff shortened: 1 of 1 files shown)


stp-ip:
/LGTM from a code perspective it looks good and the approach seems reasonable.

rlguarino:
This is a bug issue. Seems that some parts of git-sync rely on the output git commands for their input. If you enable -vvv then you get weird errors like:

```
E0824 22:40:56.629732       1 main.go:189] error syncing repo: error running command: exit status 128: "fatal: could not create directory of '.git/worktrees/-1 cc -1)\\r\\n\\r\\ndebug3: send packet: type 1\\r\\ndebug1: fd 0 clearing O_NONBLOCK\\r\\ndebug1: fd 1 clearing O_NONBLOCK\\r\\ndebug1: fd 2 clearing O_NONBLOCK\\r\\nTransferred: sent 2472, received 8936 bytes, in 0.6 seconds\\r\\nBytes per second: sent 3942.4, received 14251.2\\r\\ndebug1: Exit status 0\\r\\n5d3153b8c62736e48ef554260cd8499d34d01270\"': Filename too long\
```

Where git tries to create a file named:

`.git/worktrees/-1 cc -1)\\r\\n\\r\\ndebug3: send packet: type 1\\r\\ndebug1: fd 0 clearing O_NONBLOCK\\r\\ndebug1: fd 1 clearing O_NONBLOCK\\r\\ndebug1: fd 2 clearing O_NONBLOCK\\r\\nTransferred: sent 2472, received 8936 bytes, in 0.6 seconds\\r\\nBytes per second: sent 3942.4, received 14251.2\\r\\ndebug1: Exit status 0\\r\\n5d3153b8c62736e48ef554260cd8499d34d01270\"`

which I'm pretty certain is supposed to just be named 
`n5d3153b8c62736e48ef554260cd8499d34d01270`

thockin:
does the `-vvv` output go to stdout or stderr ?  It seems to be stderr in simple tests.  We currently use CombinedOutput() but we could be smarter if we knew the two streams were usefully different

terinjokes:
@thockin It looks like all the debug information from `ssh -vvv` ends up on stderr.

```console
$  GIT_SSH_COMMAND="ssh -vvv" git ls-remote -q origin refs/heads/master 2>/dev/null
8cd8f2145c6ffe3db4e877890d6e59cf2a38ee8c        refs/heads/master
```

We're still want to link the subprocess's stderr to our stderr, so end users like me can read what the problem is.

newbee1337:
I didn't use this parameter but still got the error - filename too long
`docker run -v /tmp/git-secret:/etc/git-secret -ti k8s.gcr.io/git-sync:v3.1.1 -ssh -repo="ssh://code@xxxxxxxxx/source/xxxxxx.git" -branch="master"`
```
error syncing repo: error running command: exit status 128: "Preparing worktree (detached HEAD 07cfecc)\nfatal: could not create directory of '.git/worktrees/rev-# Fetch received by \\\"code-bionic-web01\\\", forwarding to cluster host.\\n# Acquiring read lock for repository \\\"rPP\\\" on device \\\"code-bionic-repo04\\\"...\\n# Acquired read lock immediately.\\n# Device \\\"code-bionic-repo04\\\" is already a cluster leader and does not need to be synchronized.\\n# Cleared to fetch on cluster host \\\"code-bionic-repo04\\\".\\n07cfecc779ef157a066f0efa6f75fb62d5a5d468\"': Filename too long\n"
```

It happened on any repo I tried. I tried three repos from our phab.

wundrian:
For the record, this would be a really useful debugging tool for us if we can make it work reliably. Git by itself unfortunately doesn't capture all that much information when SSH to the repo fails and you're left retrying as your only option.

PR #85, opened 2018-02-15
Webhook call on sucessfull sync added

andjelx:

Issue #81, opened 2018-01-31
Add webhook call on successful sync


andjelx:
I'd like to contribute into the project adding webhook call on successful sync.
The case is: we trigger Prometheus to reload it's alerts \ rules stored in git repo.

The parameters are as below:
```
var doWebhook = flag.Bool("webhook", envBool("WEBHOOK_ENABLED", false),
	"trigger webhook on sucessfull git sync")
var webhook = flag.String("webhook-url", envString("WEBHOOK_URL", ""),
	"the url to send a request to when git synced")
var webhookMethod = flag.String("webhook-method", envString("WEBHOOK_METHOD", "POST"),
	"the HTTP method url to use to send the webhook")
var webhookStatusCode = flag.Int("webhook-status-code", envInt("WEBHOOK_CODE", 200),
	"the HTTP status code indicating successful triggering of reload")
```

emirozer:
Hey @andjelx are you working on this ?

I just started doing it today because, well having a very similar usecase..
https://github.com/emirozer/git-sync/commit/5aaffadae52ee201ff40dc78704373453ec12880
Its initial work only at the moment, so I'm gonna test & see if i can add something to the testing process about the hook as well. Do let me know if you already made progress?

andjelx:
Actually I made it, it’s in my fork, but haven’t received any feedback here to raise a PR :(

emirozer:
People are busy I guess :) why not raise the PR and discuss there, since it is going the be a very small change-set without side-effects to actually syncing ?

Ofc I'm not a maintainer so, no disrespect 🍻

andjelx:
Initiated: https://github.com/kubernetes/git-sync/pull/85

stp-ip:
Commented on the PR. Future feature issues might take a bit longer to respond due to priorities.

thockin:
Sorry, it's been a hellish few weeks for me.  Will look at the PR.  In general I don't have a problem with this, but I'd prefer fewer flags.  What about something like `--webhook='{"url": "http://localhost:1234/reload", method="POST", success=200}'` - that block can be parsed as JSON, with unspecified keys having default values.

Unclear if success should be a specific code or a range.

andjelx:
@thockin 
Your flags approach a good idea, but I just followed the same approach as in original code + technically passing ENV vars might be easier to use with K8 Deployments.

For now: success code is one code (not a range)

andjelx:
[...]

PR #110, opened 2018-08-29
Send a POST request when the symlink has updated

diericx:
Adds an ENV/Flag variable ```flSymlinkUpdatePostUrl``` that takes in a url. If this is set, whenever the symlink is updated git-sync will make a POST request to this URL.

**Motivation:** When using git-sync to update the configs for a container in a pod, we found ourselves having to reload the services in the container by making a POST request or SIGHUP. Integrating this directly into git-sync makes this much more straight forward as it seems to be a common pattern.

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -32,6 +32,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"net/http"
 
 	"github.com/thockin/glogr"
 	"github.com/thockin/logr"
@@ -58,6 +59,8 @@ var flMaxSyncFailures = flag.Int("max-sync-failures", envInt("GIT_SYNC_MAX_SYNC_
 	"the number of consecutive failures allowed before aborting (the first pull must succeed)")
 var flChmod = flag.Int("change-permissions", envInt("GIT_SYNC_PERMISSIONS", 0),
 	"the file permissions to apply to the checked-out files")
+var flSymlinkUpdatePostUrl = flag.String("symlink-update-post-url", envString("GIT_SYNC_SYMLINK_UPDATE_POST_URL", ""),
+	"a command to run when the symlink is updated")
 
 var flUsername = flag.String("username", envString("GIT_SYNC_USERNAME", ""),
 	"the username to use")
@@ -260,6 +263,19 @@ func updateSymlink(gitRoot, link, newDir string) error {
 		return fmt.Errorf("error replacing symlink: %v", err)
 	}
 	log.V(1).Infof("renamed symlink %s to %s", "tmp-link", link)
+
+	// If there is a symlink update callback, call it
+	if len(*flSymlinkUpdatePostUrl) > 0 {
+		log.V(1).Infof("sending post request to %s", *flSymlinkUpdatePostUrl)
+		// Send the post request
+		req, err := http.NewRequest("POST", *flSymlinkUpdatePostUrl, nil)
+		if err != nil {
+			fmt.Errorf("error sending post request (after symlink update): %v", err)
+		}
+		c := &http.Client{}
+		resp, err := c.Do(req)
+		resp.Body.Close()
+	}
 
 	// Clean up previous worktree
 	if len(currentDir) > 0 {


diericx:
I'm sending this as an employee for PubNub. PubNub already has a CLA signed, not sure why it's not working for me. I'm working with our lawyer to get everything signed and synced.

thockin:
This is a simpler form of https://github.com/kubernetes/git-sync/pull/85 - can we fold them together?  @andjelx

E.g. It seems reasonable to make the method be configurable (GET vs POST) but maybe we don't need to allow multiple webhooks.  Do we need to specify any payload?  Do we need to handle success or is it fire-and-forget?  Do we need to handle timeouts?

diericx:
@andjelx what do you think about keeping the multiple webhooks, but removing specified payload and converting it to fire-and-forget. I don't see why git-sync should worry about whether the request is successful or not.

andjelx:
@diericx Actually as for me it's better to have some timeout and error reporting as in this case you'd see was webhook sucesful or not.

In my another local change I have even added /metrics to the git-sync to track it's behaviour via Prometheus.

As for payload, I'd agree that it's not necessary.

diericx:
@andjelx Interesting. What do you think about making the ```success``` parameter of a webhook optional. If it exists, git-sync will wait to make sure the call was successful. If not, git-sync just sends the request and forgets about it.

andjelx:
@diericx Good point

diericx:
@andjelx would you mind if I took a shot at that update and used your code in this PR? I think it would be easier than trying to merge into your branch and then upstream.

andjelx:
@diericx Go on. Mention me somewhere in there Annals :-)
But I would squash all commits in your branch into one before merging.

jshaw86:
@andjelx @thockin anything else you guys need here? I can put in sometime to finish this out.

README.md
```diff
@@ -32,4 +33,26 @@ docker run -d \
     nginx
 ```
 
+## Example of webhooks usage
+**Webhook config example**
+A webhook config must be valid JSON. If ```success``` is not specified in the config, git-sync will not wait for a response.
+```json
```
thockin: nit: blank line befoe quote-blocks (some markdown parsers require it)

README.md
```diff
@@ -32,4 +33,26 @@ docker run -d \
     nginx
 ```
 
+## Example of webhooks usage
+**Webhook config example**
+A webhook config must be valid JSON. If ```success``` is not specified in the config, git-sync will not wait for a response.
```
thockin: If `success` is not specified, do we make any guarantees about behavior?  E.g. closing sockets?

If `success` is specified, how long will we wait for success?  We need a timeout or something.

Do we pass anything as body?

Do we need to allow multiple codes for success?  E.g. any 200?

Can the user assume this will follow 30{1,2,7,8} redirects properly?

README.md
```diff
@@ -32,4 +33,26 @@ docker run -d \
...
+
+```
+docker run -d \
+    -v /tmp/git-data:/git \
+    registry/git-sync:tag \
+        --repo=https://github.com/kubernetes/git-sync
+        --branch=master
+        --wait=30
+        --webhook='[{"url": "http://localhost:9090/-/reload", "method": "POST", "success": 200}]'
+        --webhook='[{"url": "http://1.2.3.4:9090/-/reload", "method": "POST"}]'
```
thockin: do we really have need for more than one webhook?

cmd/git-sync/main.go
```diff
@@ -58,6 +59,11 @@ var flMaxSyncFailures = flag.Int("max-sync-failures", envInt("GIT_SYNC_MAX_SYNC_
 	"the number of consecutive failures allowed before aborting (the first pull must succeed)")
 var flChmod = flag.Int("change-permissions", envInt("GIT_SYNC_PERMISSIONS", 0),
 	"the file permissions to apply to the checked-out files")
+	
+var flWebhooks = flag.String("webhook", envString("GIT_SYNC_WEBHOOK", ""),
+	"the JSON formatted array of webhooks to be sent when git is synced")
+var flWebhookTimeout = flag.Int("webhook-timeout", envInt("GIT_SYNC_WEBHOOK_TIMEOUT", 60),
```
thockin: Why make this a separate flag vs part of the webhook config?  or why not make the webhook config be 4 flags (--webhook-url, --webhook-method, --webhook-timeout, --webhook-success) ?

cmd/git-sync/webhooks.go
```diff
@@ -0,0 +1,71 @@
...
+	"fmt"
+)
+// Create an http client that has our timeout by default
+var netClient = &http.Client{
+	Timeout: time.Duration(time.Second * time.Duration(*flWebhookTimeout) ),
+}
+
+// Trigger channel for webhook requests. If anything is received into this channel
+//   it triggers the webhook goroutine to send new requests.
+var WebhookCallTriggerChannel = make(chan struct{})
```
thockin: Rather than make this a global, why not create it in `main` and pass it into `ServeWebhooks()` as an argument

cmd/git-sync/main.go
stp-ip: I assume it was done because of the comment here: https://github.com/kubernetes/git-sync/issues/81#issuecomment-366018849

thockin:
Hmm.   I never promised to be self-consistent. :)

If we are only supporting a single webhook (which I hope...) Then flags
seem easier and less likely to explode (typos, etc)...

Argue me?

stp-ip:
Having a single webhook seems simpler and the most used case.
One could always let the webhook endpoint spread out to other webhooks.

PR #112, opened 2018-09-18
Update example volume mount

jmcarp:
Since the default clone path is $HOME/git and the docker image sets $HOME to /tmp, we have to mount the volume at `/tmp/git` for the container to write files to the volume. As written, the example doesn't write to the volume.


jmcarp:
ping @thockin if you have time

thockin:
Sorry, I lost track!

PR #125, opened 2019-01-14
Trim ".git" suffix in git repo URL

xychu:
Both GitHub and GitLab web provides URL with ".git" suffix,
and `git clone`'s default behavior is to get rid of it,
so it's better to trim it as the default destination.

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -136,7 +136,7 @@ func main() {
 		os.Exit(1)
 	}
 	if *flDest == "" {
-		parts := strings.Split(strings.Trim(*flRepo, "/"), "/")
+		parts := strings.Split(strings.TrimSuffix(strings.Trim(*flRepo, "/"), ".git"), "/")
 		*flDest = parts[len(parts)-1]
 	}
 	if strings.Contains(*flDest, "/") {


thockin:
Sorry, why is it better?  Is something not working as it is written?  I have not heard anyone complain.

I understand what you're trying to express here, this would be a breaking change, I think, so we can't really do it.  Unless I misunderstand?

xychu:
Everything is working fine :-) What I mean "better" just means no need to add "GIT_SYNC_DEST" when the "GIT_SYNC_REPO" is end up with ".git" and user want the same behavior of `git clone`. Take the demo(https://github.com/kubernetes/git-sync/blob/master/demo/config/deployment.yaml#L22-L25) as an example:

        - name: GIT_SYNC_REPO
          value: https://github.com/kubernetes/git-sync.git
        - name: GIT_SYNC_DEST
          value: git-sync

will just be:

        - name: GIT_SYNC_REPO
          value: https://github.com/kubernetes/git-sync.git

thockin:
We could also change the demo to drop the `.git`, which is not required.

I get what you're going for here, but it would be a breaking change, which would mean adding a new flag to enable it, which more or less negates the value of this.

Unless there's some other reason, I don't think this is going to work. :(

Thanks for the PR, though.

PR #131, opened 2019-01-17
Add Webhook on git-sync update

jacksontj:
This is a continuation of #110. I have pulled over all the commits from that PR and addressed the majority of the comments therein.

The outstanding questions I know of:

- do we need multiple webhooks (array vs single): multiple seems nice and its minimal change, but no strong opinion though either way
- https://github.com/kubernetes/git-sync/pull/110/files#r242277494 -- left as a global for now to avoid passing it through the various functions -- although it seems a bit odd it ends up being smaller/cleaner code.


cmd/git-sync/main.go
```diff
@@ -62,6 +63,9 @@ var flMaxSyncFailures = flag.Int("max-sync-failures", envInt("GIT_SYNC_MAX_SYNC_
 var flChmod = flag.Int("change-permissions", envInt("GIT_SYNC_PERMISSIONS", 0),
 	"the file permissions to apply to the checked-out files")
 
+var flWebhooks = flag.String("webhook", envString("GIT_SYNC_WEBHOOK", ""),
```
thockin: As per comments in #110 I think it's net simpler to support just one webhook and to do so via 3 flags:

string URL (default "", meaning no webhook)
string method (default "POST")
int success (default 200)

cmd/git-sync/webhooks.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	return nil
+}
+
+// Wait for trigger events from the channel, and send webhooks when triggered
+func ServeWebhooks() {
+	for {
+		// Wait for trigger
+		<-WebhookCallTriggerChannel
```
thockin: allocate in main and pass in here?

jacksontj: This was addressed in the initial PR description, I'll paste down here for easy reading:

> https://github.com/kubernetes/git-sync/pull/110/files#r242277494 -- left as a global for now to avoid passing it through the various functions -- although it seems a bit odd it ends up being smaller/cleaner code.

thockin:
I don't get it - it's only passed into `ServeWebhooks()`?  You could change
the return from syncRepo to indicate whether it actually synced or not, or
you could just pass the chan to syncRepo and let it deliver, rather than
deep down in updateSymlink.

jacksontj:
@thockin the issue isn't that pass down-- there is easy. The issue is the channel is triggered in `updateSymlink` which would mean it would need to be passed from `main` -> `syncRepo` -> `addWorktreeAndSwap` -> `updateSymlink`.

So my initial goal here was to introduce the least change. But it sounds like you are okay with making changes to all those other methods so I'll go ahead and do that :)

cmd/git-sync/main.go
jacksontj: Moved to all separate CLI flags

cmd/git-sync/webhooks.go
jacksontj: Methods changed around, this is now passed through

cmd/git-sync/main.go
```diff
@@ -62,6 +62,17 @@ var flMaxSyncFailures = flag.Int("max-sync-failures", envInt("GIT_SYNC_MAX_SYNC_
 var flChmod = flag.Int("change-permissions", envInt("GIT_SYNC_PERMISSIONS", 0),
 	"the file permissions to apply to the checked-out files")
 
+var flWebhookURL = flag.String("webhook-url", envString("GIT_SYNC_WEBHOOK_URL", ""),
+	"the URL for the webook to send to. Default is \"\" which disables the webook.")
+var flWebhookMethod = flag.String("webhook-method", envString("GIT_SYNC_WEBHOOK_METHOD", "POST"),
+	"the method for the webook to send with")
+var flWebhookStatusSuccess = flag.Int("webhook-success-status", envInt("GIT_SYNC_WEBHOOK_SUCCESS_STATUS", 200),
+	"the status code which indicates a successful webhook call")
+var flWebhookTimeout = flag.Duration("webhook-timeout-duration", envDuration("GIT_SYNC_WEBHOOK_TIMEOUT_DURATION", time.Second),
```
thockin: I think webhook-timeout and ...WEBHOOK_TIMEOUT are good enough - do we really need "duration" ?

cmd/git-sync/main.go
```diff
@@ -62,6 +62,17 @@ var flMaxSyncFailures = flag.Int("max-sync-failures", envInt("GIT_SYNC_MAX_SYNC_
...
+	"the timeout used when communicating with the webhook target")
+var flWebhookBackoffDuration = flag.Duration("webhook-backoff-duration", envDuration("GIT_SYNC_WEBHOOK_BACKOFF_DURATION", time.Second*3),
```
thockin: same as above - do we need "duration" ?

cmd/git-sync/main.go
```diff
@@ -184,11 +207,26 @@ func main() {
...
 
+	// Startup webhooks goroutine
+	webhookTriggerChan := make(chan struct{}, 1)
+	if *flWebhookURL != "" {
+		webhook := Webhook{
+			URL:             *flWebhookURL,
+			Method:          *flWebhookMethod,
+			Success:         flWebhookStatusSuccess,
+			Timeout:         *flWebhookTimeout,
+			BackoffDuration: *flWebhookBackoffDuration,
```
thockin: Backoff

cmd/git-sync/main.go
```diff
@@ -184,11 +207,26 @@ func main() {
...
+		go webhook.run(webhookTriggerChan)
+	}
+
 	initialSync := true
 	failCount := 0
 	for {
 		ctx, cancel := context.WithTimeout(context.Background(), time.Second*time.Duration(*flSyncTimeout))
-		if err := syncRepo(ctx, *flRepo, *flBranch, *flRev, *flDepth, *flRoot, *flDest); err != nil {
+		if changed, err := syncRepo(ctx, *flRepo, *flBranch, *flRev, *flDepth, *flRoot, *flDest); err != nil {
+			syncDuration.WithLabelValues("error").Observe(time.Now().Sub(start).Seconds())
```
thockin: This depends on metrics, which are not in yet?

cmd/git-sync/main.go
```diff
@@ -62,6 +62,17 @@ var flMaxSyncFailures = flag.Int("max-sync-failures", envInt("GIT_SYNC_MAX_SYNC_
 var flChmod = flag.Int("change-permissions", envInt("GIT_SYNC_PERMISSIONS", 0),
 	"the file permissions to apply to the checked-out files")
 
+var flWebhookURL = flag.String("webhook-url", envString("GIT_SYNC_WEBHOOK_URL", ""),
+	"the URL for the webook to send to. Default is \"\" which disables the webook.")
+var flWebhookMethod = flag.String("webhook-method", envString("GIT_SYNC_WEBHOOK_METHOD", "POST"),
+	"the method for the webook to send with")
+var flWebhookStatusSuccess = flag.Int("webhook-success-status", envInt("GIT_SYNC_WEBHOOK_SUCCESS_STATUS", 200),
```
thockin: This defaults to 200, but you have teh option in webhook to fire-and-forget.  Which should be default?

cmd/git-sync/webhook.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+// WebHook structure
+type Webhook struct {
+	// URL for the http/s request
+	URL string
+	// Method for the http/s request
+	Method string
+	// Code to look for when determining if the request was successful.
+	//   If this is not specified, request is sent and forgotten about.
```
thockin: How would this ever be `nil`  if the flag has a default?  Should default be fire-and-forget?

cmd/git-sync/webhook.go
```diff
@@ -0,0 +1,66 @@
...
+	req, err := http.NewRequest(w.Method, w.URL, nil)
+	if err != nil {
+		return err
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), w.Timeout)
+	defer cancel()
+	req = req.WithContext(ctx)
+
+	//
```
thockin: I agree with this statement, but maybe it doesn't need to be said?

cmd/git-sync/main.go
jacksontj: botched merge, fixed

cmd/git-sync/webhook.go
jacksontj: Seems that this is pointing at an incorrect line now? There is an empty comment here-- so I will remove that. But I assume this was meant for another line.

jacksontj:
I've updated the README and flags to hopefully better explain what is happening with the webhook retries etc.

PR #132, opened 2019-01-18
Add flags/env-vars for additional SSH options

jacksontj:
This simply moves the hard-coded values into flag options, with defaults being the options they were previously.

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -69,6 +69,8 @@ var flPassword = flag.String("password", envString("GIT_SYNC_PASSWORD", ""),
 
 var flSSH = flag.Bool("ssh", envBool("GIT_SYNC_SSH", false),
 	"use SSH for git operations")
+var flSSHSecret = flag.String("ssh-key", envString("GIT_SSH_KEY", "/etc/git-secret/ssh"),
+	"the ssh key to use")
 var flSSHKnownHosts = flag.Bool("ssh-known-hosts", envBool("GIT_KNOWN_HOSTS", true),
 	"enable SSH known_hosts verification")
 
@@ -513,7 +515,7 @@ func setupGitAuth(username, password, gitURL string) error {
 func setupGitSSH(setupKnownHosts bool) error {
 	log.V(1).Infof("setting up git SSH credentials")
 
-	var pathToSSHSecret = "/etc/git-secret/ssh"
+	var pathToSSHSecret = *flSSHSecret
 	var pathToSSHKnownHosts = "/etc/git-secret/known_hosts"
 
 	fileInfo, err := os.Stat(pathToSSHSecret)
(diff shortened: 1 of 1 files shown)


cmd/git-sync/main.go
```diff
@@ -69,8 +69,12 @@ var flPassword = flag.String("password", envString("GIT_SYNC_PASSWORD", ""),
 
 var flSSH = flag.Bool("ssh", envBool("GIT_SYNC_SSH", false),
 	"use SSH for git operations")
+var flSSHSecret = flag.String("ssh-key", envString("GIT_SSH_KEY", "/etc/git-secret/ssh"),
+	"the ssh key to use")
```
thockin: Can you make the comment here be something like 'the path to the SSH key to use" - specifically it is a path, not the key itself.

thockin: or name it to ssh-key-file (to parallel ssh-known-hosts-file)

thockin: or ssh-key-path, ssh-known-hosts-path and cookie-file-path

cmd/git-sync/main.go
```diff
@@ -69,8 +69,12 @@ var flPassword = flag.String("password", envString("GIT_SYNC_PASSWORD", ""),
 
 var flSSH = flag.Bool("ssh", envBool("GIT_SYNC_SSH", false),
 	"use SSH for git operations")
+var flSSHSecret = flag.String("ssh-key", envString("GIT_SSH_KEY", "/etc/git-secret/ssh"),
```
thockin: flSSHFile to parallel flSSHKnownHostsFile  (see other options below)

jacksontj:
@thockin changed the flag names as requested, also rebased on master

cmd/git-sync/main.go
```diff
@@ -69,8 +69,12 @@ var flPassword = flag.String("password", envString("GIT_SYNC_PASSWORD", ""),
 
 var flSSH = flag.Bool("ssh", envBool("GIT_SYNC_SSH", false),
 	"use SSH for git operations")
+var flSSHKeyFile = flag.String("ssh-key-file", envString("GIT_SSH_KEY", "/etc/git-secret/ssh"),
```
thockin: nit: `GIT_SSH_KEY_FILE`

cmd/git-sync/main.go
```diff
@@ -69,8 +69,12 @@ var flPassword = flag.String("password", envString("GIT_SYNC_PASSWORD", ""),
...
+	"the ssh key to use")
 var flSSHKnownHosts = flag.Bool("ssh-known-hosts", envBool("GIT_KNOWN_HOSTS", true),
 	"enable SSH known_hosts verification")
+var flSSHKnownHostsFile = flag.String("ssh-known-hosts-file", envString("GIT_KNOWN_HOSTS_FILE", "/etc/git-secret/known_hosts"),
```
thockin: `GIT_SSH_KNOWN_HOSTS_FILE`

jacksontj:
@thockin that should do it :)

PR #133, opened 2019-01-18
Add option to allow for unlimited failures

jacksontj:
As it stands today git-sync will exit (which will in-turn cause the pod to die) if a git pull fails a number of times. If this failure was due to a remote (such as the git remote being unavailable, such as a github outage) it may be undesirable for your pods to churn. This add the option for the user to define `-1` as the failure limit which allows git-sync to retry indefinitely.

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -58,7 +58,7 @@ var flSyncTimeout = flag.Int("timeout", envInt("GIT_SYNC_TIMEOUT", 120),
 var flOneTime = flag.Bool("one-time", envBool("GIT_SYNC_ONE_TIME", false),
 	"exit after the initial checkout")
 var flMaxSyncFailures = flag.Int("max-sync-failures", envInt("GIT_SYNC_MAX_SYNC_FAILURES", 0),
-	"the number of consecutive failures allowed before aborting (the first pull must succeed, <0 allows for infinite failures)")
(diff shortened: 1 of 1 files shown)


cmd/git-sync/main.go
```diff
@@ -58,7 +58,7 @@ var flSyncTimeout = flag.Int("timeout", envInt("GIT_SYNC_TIMEOUT", 120),
 var flOneTime = flag.Bool("one-time", envBool("GIT_SYNC_ONE_TIME", false),
 	"exit after the initial checkout")
 var flMaxSyncFailures = flag.Int("max-sync-failures", envInt("GIT_SYNC_MAX_SYNC_FAILURES", 0),
-	"the number of consecutive failures allowed before aborting (the first pull must succeed)")
+	"the number of consecutive failures allowed before aborting (the first pull must succeed, <0 allows for infinite failures)")
```
stp-ip: Would be specific here. No need to allow -1+. `"-1 disables aborting for any number of failures"`

jacksontj: Is your suggestion just to doc it as such? Or to actually change the code to only do this if `-1`?

stp-ip: So doc change for sure as there is no need to make it more complicate as needed.
On enforcing this. I think being explicit is a better long term thing even, if it makes the code a bit longer, but it explicitly "documents" this within the code. So my vote for also enforcing it.

PR #134, opened 2019-01-18
Upgrade from Godeps to dep

jacksontj:


stp-ip:
If we move, we should probably move to gomods right away.

PR #135, opened 2019-01-18
Add metrics to the git-sync process

### LINKED ISSUE #130: Metrics on git-sync state
I have a variety of use-cases that require a sidecar such as this. For this type of a service (async syncer from git) I need to have some metrics so that I can monitor the health of the syncer.

Specifically I'm looking for:
- last_sync_time: this allows me to see that is in-fact pulling things
- total_error_count: this shows me that there are errors in pulling

I'm more than happy to submit the PR for this, I just wanted to get some input first -- specifically on how these metrics would be exposed. Personally I'm using prometheus exporters for most of my world, but this daemon doesn't currently have an HTTP endpoint -- would it be okay to add one? If not, what metrics format would be acceptable?

jacksontj:
To provide metrics git-sync needs an http endpoint. This is currently done by adding a CLI flag which defines the bind address, if this is set then the http server is started. 

This adds metrics for total count and latency of git-sync attempts split by status

Fixes #130


cmd/git-sync/main.go
```diff
@@ -78,8 +82,30 @@ var flCookieFile = flag.Bool("cookie-file", envBool("GIT_COOKIE_FILE", false),
 var flGitCmd = flag.String("git", envString("GIT_SYNC_GIT", "git"),
 	"the git command to run (subject to PATH search)")
 
+var flHTTPBind = flag.String("http-bind", envString("GIT_SYNC_HTTP_BIND", ""),
+	"the bind address for git-sync's HTTP endpoint")
```
stp-ip: nit: `"the bind address including port for git-sync's HTTP endpoint"`
It's assumed, but making it explicit might reduce additional questions.

jacksontj:
> flag to enable the http server parts aka full feature flag for anything starting up around the http server

Will do

> default HTTPBind?

Preference on port?

>configuring HTTPBind automatically enables metrics? Feels there should be a specific flag for metrics at least to disable them.

Also can do  :)

stp-ip:
> > flag to enable the http server parts aka full feature flag for anything starting up around the http server
> 
> Will do
> 
> > default HTTPBind?
> 
> Preference on port?
:8080 should be fine here in my view.
> 
> > configuring HTTPBind automatically enables metrics? Feels there should be a specific flag for metrics at least to disable them.
> 
> Also can do :)

cmd/git-sync/main.go
```diff
@@ -82,6 +82,8 @@ var flCookieFile = flag.Bool("cookie-file", envBool("GIT_COOKIE_FILE", false),
 var flGitCmd = flag.String("git", envString("GIT_SYNC_GIT", "git"),
 	"the git command to run (subject to PATH search)")
 
+var flHTTPEnable = flag.String("http-enable", envBool("GIT_SYNC_HTTP_ENABLE", true),
```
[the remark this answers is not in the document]
jacksontj: ah, I had this one on by default since the suggestion was to have the metrics endpoint on by default-- easy fix :)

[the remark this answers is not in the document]
stp-ip: Yeah as long as the http stuff is gated I feel having metrics on by default is fine. Just don't want to have a http server started without people knowing this.

stp-ip:
Sidenote: I would probably jump over dep and move to plain gomods without vendoring right away.

jacksontj:
I've added a commit to move to gomods instead of dep, I'm fine with doing either. The only "issue" with gomods is that it requires go 1.11 -- which I imagine isn't an issue?

thockin:
Can you also merge with https://github.com/kubernetes/git-sync/commit/a2c98a69da2a842dcdfb1710be66ffb7ffe595d6

cmd/git-sync/main.go
```diff
@@ -78,6 +80,9 @@ var flCookieFile = flag.Bool("cookie-file", envBool("GIT_COOKIE_FILE", false),
 var flGitCmd = flag.String("git", envString("GIT_SYNC_GIT", "git"),
 	"the git command to run (subject to PATH search)")
 
+var flHTTPBind = flag.String("http-bind", envString("GIT_SYNC_HTTP_BIND", ""),
+	"the bind address for git-sync's HTTP endpoint")
```
thockin: example value here, maybe some docs.

Maybe add a TODO (or better - file a bug) to make this default-on in the next major rev?

cmd/git-sync/main.go
```diff
@@ -82,8 +82,10 @@ var flCookieFile = flag.Bool("cookie-file", envBool("GIT_COOKIE_FILE", false),
 var flGitCmd = flag.String("git", envString("GIT_SYNC_GIT", "git"),
 	"the git command to run (subject to PATH search)")
 
-var flHTTPBind = flag.String("http-bind", envString("GIT_SYNC_HTTP_BIND", ""),
-	"the bind address for git-sync's HTTP endpoint")
+var flHTTPBind = flag.String("http-bind", envString("GIT_SYNC_HTTP_BIND", ":8080"),
```
thockin: I don't think we can make it default-on.  People doing a minor update would get a major behavioral change.

cmd/git-sync/main.go
```diff
@@ -82,6 +82,8 @@ var flCookieFile = flag.Bool("cookie-file", envBool("GIT_COOKIE_FILE", false),
 var flGitCmd = flag.String("git", envString("GIT_SYNC_GIT", "git"),
 	"the git command to run (subject to PATH search)")
 
+var flHTTPEnable = flag.Bool("http-enable", envBool("GIT_SYNC_HTTP_ENABLE", false),
```
thockin: Why make this an enable flag and an address flag?  Less is more when it comes to flags.  Having the address flag also mean "enable this" seems reasonable to me - convince me otherwise?

cmd/git-sync/main.go
```diff
@@ -181,14 +211,30 @@ func main() {
...
+	if *flHTTPEnable {
+		ln, err := net.Listen("tcp", *flHTTPBind)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "ERROR: unable to bind HTTP endpoint: %v\n", err)
+		}
+		go func() {
+			if *flHTTPMetrics {
+				http.Handle("/metrics", promhttp.Handler())
+			}
+			http.Serve(ln, http.DefaultServeMux)
```
thockin: Should we have a do-nothing success handler for / ?

cmd/git-sync/main.go
stp-ip: I suggested this to have a global feature gate for http as this is completely new behaviour.
That enables use to enable various http features by default without enabling the whole http server by default.
That was the main reasoning.

cmd/git-sync/main.go
stp-ip: Depends on the decision on a global http flag or not. If we have the global gate, then default enable seems reasonable. If not, then I agree with default off.

cmd/git-sync/main.go
stp-ip: From a prometheus perspective most implementations I've seen default to 404 as there is nothing serving so why give back a 200 return.

stp-ip:
> Why is this removing glogr and logr? We use those...
It's removing them from vendor with the switch to gomods, but keeps them within the gomods files. Not lost as far as I can see, just not vendored. 
We can use add manual support for vendor with gomods:
`go mod vendor` <- builds the vendor folder based on gomods files
`go build -mod=vendor` or setting `GOFLAGS=-mod=vendor` <- use vendor folder during build

cmd/git-sync/main.go
jacksontj: Agreed, this was default-on only after I added the global flag.

cmd/git-sync/main.go
jacksontj: IMO the address flag is sufficient, but it depends on what the long-term plans are. If the plan is to have this eventually be default on then having the separate flag which then gets removed makes sense.

cmd/git-sync/main.go
jacksontj: I'm fine to add a dumb liveliness handler to `/` but that really isn't part of this metrics PR IMO -- but I'd be more than happy to add it in.

jacksontj:
The reason for the go mods thing is because godeps (the current vendor tool) is old and deprecated. I had originally just migrated to `dep` but @stp-ip pointed out (rightly so) that `go mods` is the "correct" way to do it nowadays. Personally I'm fine either way, since this isn't a lib people include we have complete freedom here. I have both commits in this PR so we just need to pick which one and I'll axe the other.

cmd/git-sync/main.go
thockin: @stp-ip I don't quite follow.  How could you enable pprof or healthz or metrics without enabling HTTP?  My thinking was that the http-bind-address is the flag that disables all HTTP (if empty) or enables it (if not empty).  Each use-case of HTTP may want its own flag to enable/disable handler registration.  That saves us one flag, which seems small but isn't :)

cmd/git-sync/main.go
jacksontj: dumb liveliness handler added

thockin: As per other comment, the ideal review sequence would be to add HTTP with a do-nothing / handler in one PR, followed by adding metrics in a different PR.  I'll take them together, if need be.

thockin:
I am fine to switch to modules, if they work for this repo, but I think we need to vendor.  Not vendoring is playing chicken with a brick wall.  Eventually we will hit it.

Maybe break the vendoring/mods stuff to a distinct PR?

jacksontj:
`go mods` removes the need for vendoring. So if you want to *keep* vendoring then it sounds like `dep` is the solution.

This PR adds new dependencies (prom, etc.) and as such changes need to be made to the vendor directory. I had some issues getting `godep` to work for me-- so I'd rather update. If `dep` is fine I'll just remove the `go mods` patch and stick with `dep` (which does manage a vendor directory).

thockin:
Sorry, nothing removes the need for vendoring.  Anyone who says they
don't need vendoring is either lying or just hasn't met reality yet.

Locking to a rev solves one problem (knowing what your dep is).
Vendoring solves a different problem (making sure your dep doesn't
disappear underneath you).  We have experienced the latter and I never
want to waste a day on that again.

jacksontj:
@thockin I've added a commit which re-adds the vendor directory.

This PR adds a few dependencies (prom, etc.) and I was having problems getting `godep` to add them. If we want that as a separate PR it would more-or-less need to be before this PR.


> Sorry, nothing removes the need for vendoring. Anyone who says they don't need vendoring is either lying or just hasn't met reality yet.

I agree -- I have *also* been bit by that, but that *was* the original pitch :) I didn't realize you could use `go mods` and keep the vendor dir-- thanks @stp-ip for pointing that out. So this PR now has both mods and vendor.

cmd/git-sync/main.go
stp-ip: http-enable is a global flag. It needs to be enabled for all http features to function or else they are feature gated.

So pprof, healthz and metrics can be enabled by default, but http-enable is not.

The idea behind it would be to have metrics and pprof on different ports later on, if that becomes necessary and also making it easier to be backwards compatible while enabling new http enabled things by default on a case by case basis.

Happy to simplify here, just felt like a saner version to introduce this considering people also use this as a cli so disableing http by default cuts down on their need to disable http features, while cloud native people usually just need to enable http and get pprof, metrics etc.

Might just be me liking being explicit.

jacksontj: The one odd thing about having the 2 flags is that you can have it `enabled` but still have an empty listen port which is just a startup error.

cmd/git-sync/main.go
```diff
@@ -86,8 +86,10 @@ var flCookieFile = flag.Bool("cookie-file", envBool("GIT_COOKIE_FILE", false),
 var flGitCmd = flag.String("git", envString("GIT_SYNC_GIT", "git"),
 	"the git command to run (subject to PATH search)")
 
-var flHTTPBind = flag.String("http-bind", envString("GIT_SYNC_HTTP_BIND", ""),
-	"the bind address for git-sync's HTTP endpoint")
+var flHTTPBind = flag.String("http-bind", envString("GIT_SYNC_HTTP_BIND", ":8080"),
+	"the bind address (including port) for git-sync's HTTP endpoint")
+var flHTTPMetrics = flag.Bool("http-metrics", envBool("GIT_SYNC_HTTP_METRICS", true),
```
thockin: This commit is still part of the series?

cmd/git-sync/main.go
thockin: This is still part of the commit series?

thockin: I still think one flag is simpler.

thockin:
this looks great, thanks.

What can we do to get some testing on this?  Maybe e2e test that just hits the URLs we expect?

thockin:
I'm going to approve to clear the docket, but please circle back for a small test

Issue #54, opened 2017-03-23
Add sparse checkout feature


sdurrheimer:
Hi there,

Would be nice if we could sparse checkout a specified list of folders/files.

stp-ip:
Could you give a specific use case, why that would be needed?

sdurrheimer:
This is very usefull when you have multi-purpose repository but you don't want to clone the full repository each each time, it saves time and disk space. 

eg. a repository containing all sort of infrastructure configurations like kubernetes yml files and your json dashboards for grafana but you only want to clone those json as a volume for your grafana deployment.

thockin:
Is there a way to express that to git?

On Mon, Mar 27, 2017 at 12:29 AM, Steve Durrheimer <notifications@github.com

sdurrheimer:
https://git-scm.com/docs/git-read-tree#_sparse_checkout

```
git config core.sparseCheckout true
echo "{REPO_PATH}/{DESIRED_SUBDIR}/*" > .git/info/sparse-checkout
git checkout {BRANCH}
```

trnl:
The problem with sparseCheckout is that still the whole repo will be cloned and then only specified folder will be checked into working tree. So, disk space and time still used.

sdurrheimer:
@trnl `--depth 1` should solve this problem.

thockin:
I don't think --depth solves the same problem OP was asking about.

stp-ip:
So since 2.19 git has a partial clone option, which most likely isn't supported on the server-side widely.
Adding in for context: https://github.com/git/git/blob/v2.19.0/Documentation/technical/partial-clone.txt

In general sparse checkout would still grab the data, but it wouldn't show within the tree.
Reducing the depth is already an optional option.
So sparse checkout seems to really only make sense with partial cloning support in my view.
As the main point is not downloading unnecessary files and not download unnecessary files, but hide them (sparse checkout):

If I missed something, please correct me and add your thoughts.

ribbybibby:
I think that even without the benefits of a partial checkout, sparse checkouts can still be useful in some situations. 

For instance, I have an application that iterates through the folders at the top level of a repository and then does some work with the yaml files it finds in each directory.

See this structure:
```
.
├── team-a_access
│   └── access.yaml
├── team-a_app
│   └── foo.yaml
├── team-b_service
│   └── foo.yaml
│   └── bar.yaml
├── team-c_access
│   └── hola.yaml
│   └── helloworld.yaml
[...]

PR #138, opened 2019-01-23
Use depth option when doing fetches

jacksontj:
Previously the `depth` flag was only used for the initial clone-- so
although you might start with a depth=10 as more commits show up you are
always >10. With this diff we enforce that depth on each fetch, this way
old-commits can get GCd off to reduce the size of the local checkout
required.

Related to #54


thockin:
Needs a rebase.

Can you look at a small e2e test that uses the depth flag, if there isn't one yet?

jacksontj:
@thockin there wasn't an existing e2e test, so I covered both the base clone as well as updates in the same test.

PR #141, opened 2019-01-28
Build rel tweaks

thockin:
This syncs to the latest template I maintain.  It reduces replicated work when possible.  We don't handle multiple platforms here yet, but we should.


stp-ip:
arm was blocked by alpine not having an arm version I think. Not sure about the other platforms so.
/approve
/lgtm

PR #142, opened 2019-01-31
~ -> $HOME

EugeneChung:
On Mac, ~ is not translated to the home directory. I guess $HOME is more portable.


EugeneChung:
I signed it

stp-ip:
Thanks. This looks good and is just a small docs change. Didn't know about `~` on the Mac, but as `$HOME` should be set on any POSIX compliant system, that's a good solution.

/approve
/lgtm

PR #144, opened 2019-02-02
Fix SSH

### LINKED ISSUE #129: Enforcing 0400 permission on git ssh key has other security implications
Because kubernetes secret volumes are mounted as owned by root, this means that if /etc/git-secret/ssh is mounted as a secret volume, to meet the 0400 permission requirement, the key is readable only by root. 

This means that the container needs to run as root, which is a bad practice (and might breach some PodSecurityPolicies) 

There is no amazing workaround for this - the [canonical discussion](https://github.com/kubernetes/kubernetes/issues/2630) suggests using fsGroups, which would work well, but would need 0440 permissions to be readable by a non-root user.

Could we have a flag that turns the permission checking off, or allows a more permissive permission ot be set?

thockin:
* Create a git-sync user to run as, with an entry in /etc/passwd and
writable homedir

* Remove our own validation of key perms - let SSH do that.

* Update docs

Fixes #129

### DIFF
--- a/Dockerfile.in
+++ b/Dockerfile.in
@@ -24,8 +24,8 @@ RUN apk update --no-cache && apk add \
     git \
     openssh-client
 
-# Set $HOME so git can create things like .gitconfig
-ENV HOME /tmp
+RUN echo "git-sync:x:65533:65533::/tmp:/sbin/nologin" >> /etc/passwd
 
-USER nobody:nobody
+WORKDIR /tmp
+USER git-sync:nobody
 ENTRYPOINT ["/{ARG_BIN}"]
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -641,13 +641,9 @@ func setupGitSSH(setupKnownHosts bool) error {
 	var pathToSSHSecret = *flSSHKeyFile
 	var pathToSSHKnownHosts = *flSSHKnownHostsFile
 
-	fileInfo, err := os.Stat(pathToSSHSecret)
+	_, err := os.Stat(pathToSSHSecret)
 	if err != nil {
 		return fmt.Errorf("error: could not find SSH key Secret: %v", err)
-	}
-
-	if fileInfo.Mode() != 0400 {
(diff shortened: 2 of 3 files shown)


willthames:
@thockin is this PR available as a container somewhere? If it is I'll happily try it out.

willthames:
Tested - works with a pod security context with fsGroup set.

stp-ip:
Tested as well and code looks good as well.
/approve

george-angel:
Sorry for necrobumping, happy to create a new Issue (or Pull Request?)

Using `fsGroup` is a bad option here. It has an unwanted effect of making all files "group readable" in the pod, which overwrites `defaultMode` of the volume. More details here: https://github.com/kubernetes/kubernetes/issues/57923

Specifically what this means is that running git-sync and another, application container in a single pod, with suggested ssh config for git-sync, makes it impossible to use openssh with key files in the application container. Private key file gets group read permission from `fsGroup` and openssh complains that file permissions are too open.

I have solved this the following way:

- remove `Pod.spec.securityContext.fsGroup`
- add `Pod.spec.containers["git-sync"].securityContext.RunAsGroup` `0`

Which is not great, but the best compromise I think, alternatives are:

- mounting secret in an initContainer, copying secrets over to another volume and chmoding them
- running all as root
- running as privileged (same?)

I confirmed this on #sig-storage https://kubernetes.slack.com/archives/C09QZFCE5/p1562078579224300

willthames:
@george-angel I've been using this fix for months without any complaints that the file permissions are too open.

Do you have a test case that fails? Might be worth opening a new issue with that.

I'm not tied to the fsGroup solution but it's better than running as root which was the previous solution 

RunAsGroup is nice but it's only been available through a beta feature gate since 1.14, which means AKS, EKS etc. might not even have it for 3-6 months

george-angel:
@willthames the exact issue I'm seeing is described in detail in https://github.com/kubernetes/kubernetes/issues/57923

From openssh man pages:

```
~/.ssh/id_rsa
         Contains the private key for authentication.  These files contain sensitive 
         data and should be readable by the user but not
         accessible by others (read/write/execute).  ssh will simply ignore a private 
         key file if it is              
         accessible by others.  It is possible to specify a
         passphrase when generating the key which will be used to encrypt the sensitive 
         part of this file using 3DES.
```

Issue #120, opened 2018-12-17
Should rename (compatibly) the --dest flag to --link or something


thockin:
`--dest` and `$GIT_SYNC_DEST` are somewhat obtuse.

techdragon:
@thockin #148 was approved but never merged, it might be good to just re-open it and complete the process to be able to close this issue.

thockin:
Superseded by #284

PR #148, opened 2019-02-19
rename dest

huanwei:
Signed-off-by: huanwei <huan@harmonycloud.cn>

fix issue #120


stp-ip:
Thanks for taking up the issue.

The PR currently only renames "dest" to "link", but isn't backwards compatible. We need to keep it backwards compatible so that old setups don't break, but are just encouraged to move to `--link` via a log message at start or something.

So take in `--dest` and `--link`, where link takes precedent, but dest might be used as a value,  if link isn't set. Same for the ENVs etc.
When an ENV or flag is used for dest we should log a deprecation warning on startup or similar.

stp-ip:
/approve

Final review appreciated @thockin

huanwei:
PTAL, appreciated @thockin

PR #155, opened 2019-03-18
Add libnss_wrapper and a wrapper to use it

Vampouille:
Compile libnss_wrapper from source and create a script that populate at
runtime passwd/group databases with current running user.


Vampouille:
I'm not very confortable with building libnss_wrapper in this image as said by @tianon here : https://github.com/docker-library/postgres/pull/543 it is the responsibility of this repository to maintain build instructions of libnss_wrapper ?

thockin:
If we switch to debian as a base, this gets easier right?  #154

thockin:
Closing as abandoned for now.

PR #157, opened 2019-03-22
Convert to use go modules

thockin:


stp-ip:
Checking over the weekend.

thockin:
Superceded by #158

PR #158, opened 2019-03-24
Convert to use go modules

stp-ip:
Based on #157 

This fixes a few things.
* use vendor for tests, build etc. `-mod=vendor`
* remove need for git in alpine container as vendor is used `-mod=vendor`
* fix permission and GOPATH issues from package installations by removing manual installation via `go test -i` as that is discouraged since v1.10 due to the go cache afaik


stp-ip:
Sidenote: Should we really use "go install..." to build? Thought that was discouraged especially with us now having the option to cross compile to various archs.

thockin:
`go build` can only build a single binary (though it will happily build multiple, it will discard the results :-/ - thanks, Go).  This doesn't matter for us, but the build here is tooled around all the binaries ending up in bin/... so go install or go build are pretty much the same result.

/lgtm
/approve

PR #162, opened 2019-04-02
git-sync: adds flag to skip abort on first error

### LINKED ISSUE #161: Add an option to keep retrying even if first sync attempt fails
I'd like to have a settable option by which `git-sync` will not abort on a failed first `git sync` attempt, but would keep retrying.

This seems a bit counter-intuitive.  I've found, however, that restarting a git-sync deployment that is seeing a transient error is less confusing than a git-sync that has reached a full 5 minute backoff delay and then gets fixed -- because if you are not careful you may miss the fact that you are seeing a backed-off git-sync.

In practice, we end up power-cycling git-sync anyways.  At which point, we may as well keep git-sync up and running.

Long story short, PR upcoming.

filmil:
Adds the flag to allow git-sync to soldier on on first failure.  This
is useful if the git-sync lifecycle is managed by a controller that
has its own ways of reporting sync failures.

Fixes #161.

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -64,6 +64,8 @@ var flSyncTimeout = flag.Int("timeout", envInt("GIT_SYNC_TIMEOUT", 120),
 	"the max number of seconds for a complete sync")
 var flOneTime = flag.Bool("one-time", envBool("GIT_SYNC_ONE_TIME", false),
 	"exit after the initial checkout")
+var flAbortOnFirstSyncError = flag.Bool("abort-on-first-sync-error", envBool("GIT_ABORT_ON_FIRST_SYNC_ERROR", true),
+	"if set, aborts git-sync if the first attempt to sync results in an error, otherwise keeps retrying")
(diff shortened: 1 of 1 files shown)


filmil:
> What if, instead of a new flag, we just simplified and make this the main-path behavior? I am not sure that being different at first sync is worthwhile. Do you want to take a run at this?

Works for me.  I'll make a new PR tomorrow.

I mostly didn't want to touch the code paths since I didn't understand how critical the current behavior is.  Obligatory https://xkcd.com/1172

filmil:
I thought about this a bit.  The benefit of aborting on first sync or after a number of failed retries is that we get a chance to recover from a botched configuration, such as
say a typo in the git repo path.  This is seems like it would be valuable in settings where no controller is managing git-sync.  If the user fixes the configuration, the user may then rely on Kubernetes restart mechanism to fix the pod.  Updated this PR with a comment to that effect for posterity.

Now, I don't know how much this matters in practice. I'll make another PR which just removes the special code path for the first sync. Feel free to choose either of the approaches.

filmil:
FYI added https://github.com/kubernetes/git-sync/pull/163

filmil:
Since #163 was approved, this is no longer needed.  Closing.

PR #163, opened 2019-04-02
git-sync: removes special exit on first error

### LINKED ISSUE #161: Add an option to keep retrying even if first sync attempt fails

filmil:
Old code used to exit at any error seen on first sync attempt.  This
didn't prove useful in practice, so removing that special case.

This may make git-sync slower to recover after user fixes a
non-retryable error, as now flMaxSyncFailures are needed before the pod
fails.  It may make sense in practice.

Fixes #161, in a different way than is proposed in PR #162.

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -289,7 +289,8 @@ func main() {
 		if changed, err := syncRepo(ctx, *flRepo, *flBranch, *flRev, *flDepth, *flRoot, *flDest); err != nil {
 			syncDuration.WithLabelValues("error").Observe(time.Now().Sub(start).Seconds())
 			syncCount.WithLabelValues("error").Inc()
-			if initialSync || (*flMaxSyncFailures != -1 && failCount >= *flMaxSyncFailures) {
+			if *flMaxSyncFailures != -1 && failCount >= *flMaxSyncFailures {
+				// Exit after too many retries, maybe the error is not recoverable.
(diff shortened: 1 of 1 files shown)


cmd/git-sync/main.go
```diff
@@ -289,7 +289,8 @@ func main() {
 		if changed, err := syncRepo(ctx, *flRepo, *flBranch, *flRev, *flDepth, *flRoot, *flDest); err != nil {
 			syncDuration.WithLabelValues("error").Observe(time.Now().Sub(start).Seconds())
 			syncCount.WithLabelValues("error").Inc()
-			if initialSync || (*flMaxSyncFailures != -1 && failCount >= *flMaxSyncFailures) {
```
thockin: Can we get rid of the initialSync variable entirely now?

thockin: Oops, no we can't

PR #165, opened 2019-04-06
exec: adds an option to log the command line

### LINKED ISSUE #164: git-sync should have an option to log a failed `exec.Command` call
It would help debugging tremendously if git-sync would log (either always, or as an option) the command line that it used to subprocess git.  This way, if something fails, one has more than a cryptic server-side error message to work with.

@frankfarzan originally suggested this.

filmil:
Having a way to log the full command line that is actually being
executed is of huge help in debugging.  When we know what command
line failed, we have a chance to try and reproduce the error ourselves.

Calls to os/exec.Command{,Context} are moved to their own package to
provide a drop-in logging wrapper, but also to allow independent setting
of logging levels in pkg/exec through the use of --vmodule=...

Fixes #164.

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -602,10 +602,12 @@ func runCommand(ctx context.Context, cwd, command string, args ...string) (strin
 	}
 	output, err := cmd.CombinedOutput()
 	if ctx.Err() == context.DeadlineExceeded {
+		log.V(3).Info("command timed out", "err", err, "cwd", cwd, "cmd", cmdForLog(command, args...))
 		return "", fmt.Errorf("command timed out: %v: %q", err, string(output))
 
 	}
 	if err != nil {
+		log.V(3).Info("error running command", "err", err, "cwd", cwd, "cmd", cmdForLog(command, args...))
 		return "", fmt.Errorf("error running command: %v: %q", err, string(output))
 	}
 


thockin:
I don't understand this PR.  We already run everything through `runCommand()` which logs the whole commandline at V(5).  This would log the commandline again in case of error.  Is that the intention?

If so, why not change `runCommand()` rather than add a whole new file?

E.g. if err is nil, log at V(3) the full commandline, err, and output.

That way I can get verbose exec errors at V(3) and verbose exec non-errors at V(5).

filmil:
> I don't understand this PR. We already run everything through `runCommand()` which logs the whole commandline at V(5). 

Not everything is funneled through runCommand. See for example https://github.com/kubernetes/git-sync/blob/master/cmd/git-sync/main.go#L599
The `V(5)` is sadly not useful as it stands today, as it can't be enabled without in the process also enabling a bunch of other logging.

> If so, why not change `runCommand()` rather than add a whole new file?
One reason is so that `--vmodule` can then be used to enable focused verbose logging, without affecting any other logging statements in the program.

> E.g. if err is nil, log at V(3) the full commandline, err, and output.
> That way I can get verbose exec errors at V(3) and verbose exec non-errors at V(5).

That would work for me, just as well.

thockin:
>> I don't understand this PR. We already run everything through runCommand() which logs the whole commandline at V(5).
>
> Not everything is funneled through runCommand. See for example https://github.com/kubernetes/git-sync/blob/master/cmd/git-sync/main.go#L599

That is *inside* `runCommand()` - there is any exec that is not using `runCommand()` it should be.

Yeah, vmodule is easier, but I think a simpler gradient of  V levels is sufficient here.  If you update PR to that, ping me.

filmil:
OK, this makes the change *much* simpler.  @thockin PTAL?

There are still code paths that don't use `runCommand` but that can be fixed separately. I had previously posted a link to the [inside of runCommand][1] which didn't illustrate the issue, but there are others like say [this one in setupGitAuth][2] which still seem separate. I'll file a bug to fix that.

[1]: https://github.com/kubernetes/git-sync/blob/master/cmd/git-sync/main.go#L599
[2]: https://github.com/kubernetes/git-sync/blob/master/cmd/git-sync/main.go#L617

thockin:
I was thinking of a little more nuance.  e.g.

```
runCommand() {
 klog.V(5).Infof("running %v", commandline)
 if err := exec(commandline); err != nil {
   log.V(3).Infof("Failed to run %v", commandline)
}
```

This way V(3) logs any failure and V(5) logs any run, success or failure.   Failures get logged twice, but I am OK with that

filmil:
@thockin PTAL, changed to `V(...)`-log the errors more or less as you advised.

fairyhunter13:
@thockin Any ways to make this application verbose? It seems for me to hard to debug this apps.

PR #178, opened 2019-07-26
Fix multi-platform build in Makefile

thockin:
I needed this to cut v3.1.2


thockin:
self-approving to cut the release

PR #185, opened 2019-10-02
:sparkles: add submodules support

michallula:
This PR adds support for git submodules. Current code doesn't fetch submodules if such are present in git repository. Submodules are empty directories instead. We use `git-sync` in our project as a sidecar container for Apache Airflow and `git-sync` is responsible for shipping Airflow Dags code to Airflow webserver/workers. Git submodules support is really important for us, cause our current code repository heavily use submodules to demarcate different teams responsibilities. I've tested this patch on our Airflow instance and it works as expected. I hope that this PR will be useful for the community.

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -452,6 +452,14 @@ func addWorktreeAndSwap(ctx context.Context, gitRoot, dest, branch, rev string, 
 	}
 	log.V(0).Info("reset worktree to hash", "path", worktreePath, "hash", hash)
 
+	// Update submodules
+
+	_, err = runCommand(ctx, worktreePath, *flGitCmd, "submodule", "update", "--init", "--recursive")
+	if err != nil {
+		return err
+	}
+	log.V(0).Info("updating submodules")
+
 	if *flChmod != 0 {
 		// set file permissions
 		_, err = runCommand(ctx, "", "chmod", "-R", strconv.Itoa(*flChmod), worktreePath)
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -454,6 +454,9 @@ func addWorktreeAndSwap(ctx context.Context, gitRoot, dest, branch, rev string, 
 
 	// Update submodules
 	// NOTICE: it works for repos with or without submodules
+	// TODO: add --depth flag support
+	// possible bug in git: git fetches full submodule history even with --depth flag is provided
+	// (tested with git version 2.23.0 on Mac)
 	_, err = runCommand(ctx, worktreePath, *flGitCmd, "submodule", "update", "--init", "--recursive")
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


michallula:
I signed it

cmd/git-sync/main.go
```diff
@@ -452,6 +452,14 @@ func addWorktreeAndSwap(ctx context.Context, gitRoot, dest, branch, rev string,
 	}
 	log.V(0).Info("reset worktree to hash", "path", worktreePath, "hash", hash)
 
+	// Update submodules
+
+	_, err = runCommand(ctx, worktreePath, *flGitCmd, "submodule", "update", "--init", "--recursive")
```
thockin: I thin you should respect the `--depth` flag here, too, shouldn't you?

thockin: Can you please add a comment here that repos that have no sub-modules will succeed here with no errors?

cmd/git-sync/main.go
```diff
@@ -452,6 +452,14 @@ func addWorktreeAndSwap(ctx context.Context, gitRoot, dest, branch, rev string,
...
+	if err != nil {
+		return err
+	}
+	log.V(0).Info("updating submodules")
```
thockin: "updated"

cmd/git-sync/main.go
michallula: ok, will add test cases soon, I've tested it in our project and it works fine for both cases: repo with submodules and without submodules. But of course it should be covered by tests.

michallula: > I thin you should respect the `--depth` flag here, too, shouldn't you?

right, how should it be supported? additional command line flag or reuse of existing `--depth` flag ?

michallula: I've added --depth flag support

michallula:
@thockin I've updated PR with tests. I also tried to add `--depth` flag support but it occurred that it doesn't work correctly in git (unless for git version 2.23.0 on Mac). Command `git submodule update --init --recursive --depth 1` always fetches full submodules history. I've tried also `--recomend-shallow` options, but it wasn't worked as well. 

What I would suggest is to provide `--depth` support in separate PR, cause even without it this PR seems to be useful (at least for my company :).

My WIP PR with `--depth` support can be found here: https://github.com/michallula/git-sync/pull/1

thockin:
Regarding --depth, do you think this is a bug in git or just I misunderstand the feature?

cmd/git-sync/main.go
thockin: If we're NOT going to add --depth here, can you leave a TODO explaining why, so we can investigate later?

michallula:
> Regarding --depth, do you think this is a bug in git or just I misunderstand the feature?

it looks like bug in git, but maybe I've done something wrong. I will push `--depth` flag support with some tests in separate branch. Maybe you could cross-check if I made something stupid.

cmd/git-sync/main.go
michallula: done

PR #187, opened 2019-10-21
support --depth flag for submodules

michallula:
Follow up of https://github.com/kubernetes/git-sync/pull/185 with --depth flag support.

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -453,11 +453,12 @@ func addWorktreeAndSwap(ctx context.Context, gitRoot, dest, branch, rev string, 
 	log.V(0).Info("reset worktree to hash", "path", worktreePath, "hash", hash)
 
 	// Update submodules
-	// NOTICE: it works for repos with or without submodules
-	// TODO: add --depth flag support
-	// possible bug in git: git fetches full submodule history even with --depth flag is provided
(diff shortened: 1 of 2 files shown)


michallula:
@thockin PR with depth support as I promised before. As I said, tests fails (at least on Mac, git version 2.23.0) cause `--depth` flag doesn't prevent from fetching whole history of submodule repository. Any ideas what is wrong here?

thockin:
git.2.23.0 on Linux also seems to fail

```
testcase submodule-sync-depth: FAIL:  forward submodule depth mismatch expected=1 actual=2
```

thockin:
Running manually I see:

```
$ git submodule update --init --recursive --depth 1
Cloning into '/tmp/git-sync-test.3215/root/rev-719f2db72cbd4446681db56818fc2e26a6d3003b/sub'...
warning: --depth is ignored in local clones; use file:// instead.
```

Changing the `submodule add` line to "file://$SUBMODULE" makes the test pass.

Can you integrate that (and in the other submodule test) and confirm.  Ping me when ready for another look.

Thanks!

michallula:
@thockin nice catch, haven't noticed it before. Thanks a lot for debugging. I've updated tests and they pass indeed. Please make review once again.

Issue #166, opened 2019-04-09
Funnel all `exec.Command` calls through `runCommand`


filmil:
`runCommand` has a nice facility that allows us to log the executed commands, which helps debugging.

There are code paths like say here: https://github.com/kubernetes/git-sync/blob/master/cmd/git-sync/main.go#L617 which don't use `runCommand` but call `exec.Command` directly, which miss out on this opportunity.

Seem like it would be good to have all code paths that subprocess a command go through `runCommand` at least to get the benefit of the debug logging.

thockin:
Agree.  All exec's should use runCommand

PR #190, opened 2019-10-24
Replace exec.Command with runCommand

hypnoglow:
Fixes https://github.com/kubernetes/git-sync/issues/166

---

Probably I need some assistance in the following questions:

- [x]  I'm not sure what to do with [this one](https://github.com/kubernetes/git-sync/blob/master/cmd/git-sync/main.go#L634), as it requires piping which is not part of runCommand. What should we do here?
  -  Extend `runCommand` to support it (maybe via `func (cmd *exec.Cmd)` parameter that allows to manipulate the command
  - Introduce new function like `runCommandHook` with the extended signature (as described above)
  - Just keep it as is
- [x] I put `context.TODO()` for the context parameter for now.  
  - Should we move [context creation](https://github.com/kubernetes/git-sync/blob/master/cmd/git-sync/main.go#L288) up so `setupGitAuth`/`setupGitCookieFile` can use it?
  - Or should we create a new separate context for "initialization" functions?

Thanks in advance.

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -123,6 +123,9 @@ var (
 	}, []string{"status"})
 )
 
+// initTimeout is a timeout for initialization, like git credentials setup.
+const initTimeout = time.Second * 30
+
 func init() {
 	prometheus.MustRegister(syncDuration)
 	prometheus.MustRegister(syncCount)
@@ -217,8 +220,12 @@ func main() {
 		os.Exit(1)
 	}
 
+	// This context is used only for git credentials initialization. There are no long-running operations like
+	// `git clone`, so initTimeout set to 30 seconds should be enough.
+	ctx, cancel := context.WithTimeout(context.Background(), initTimeout)
+
 	if *flUsername != "" && *flPassword != "" {
-		if err := setupGitAuth(*flUsername, *flPassword, *flRepo); err != nil {
+		if err := setupGitAuth(ctx, *flUsername, *flPassword, *flRepo); err != nil {
 			fmt.Fprintf(os.Stderr, "ERROR: can't create .netrc file: %v\n", err)
 			os.Exit(1)
 		}
@@ -232,11 +239,14 @@ func main() {
 	}
 
 	if *flCookieFile {
-		if err := setupGitCookieFile(); err != nil {
+		if err := setupGitCookieFile(ctx); err != nil {
 			fmt.Fprintf(os.Stderr, "ERROR: can't set git cookie file: %v\n", err)
 			os.Exit(1)
 		}
 	}
+
+	// The scope of the initialization context ends here, so we call cancel to release resources associated with it.
+	cancel()
 
 	if *flHTTPBind != "" {
 		ln, err := net.Listen("tcp", *flHTTPBind)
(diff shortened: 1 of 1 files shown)


cmd/git-sync/main.go
```diff
@@ -625,21 +625,21 @@ func runCommand(ctx context.Context, cwd, command string, args ...string) (strin
...
-	output, err := cmd.CombinedOutput()
+
+	_, err := runCommand(context.TODO(), "", *flGitCmd, "config", "--global", "credential.helper", "cache")
 	if err != nil {
-		return fmt.Errorf("error setting up git credentials %v: %s", err, string(output))
+		return fmt.Errorf("error setting up git credentials: %v", err)
 	}
 
-	cmd = exec.Command(*flGitCmd, "credential", "approve")
+	cmd := exec.Command(*flGitCmd, "credential", "approve")
```
thockin: Can you please add a comment where that this is not using `runCommand` because we need to fiddle with stdin ?

cmd/git-sync/main.go
```diff
@@ -687,10 +687,16 @@ func setupGitCookieFile() error {
 		return fmt.Errorf("error: could not access git cookie file: %v", err)
 	}
 
-	cmd := exec.Command(*flGitCmd, "config", "--global", "http.cookiefile", pathToCookieFile)
-	output, err := cmd.CombinedOutput()
-	if err != nil {
-		return fmt.Errorf("error configuring git cookie file %v: %s", err, string(output))
+	if _, err = runCommand(
```
thockin: Can you please linewrap this as:

```
if _, err = runCommand(context.TODO(), "",
        *flGitCmd, "config", "--global", "http.cookiefile", pathToCookieFile); err != nil {
```

or just not bother to linewrap it? :)

thockin:
We probably should just make a context for all initializations and pass it in to them.

Regarding the pipe, something like `runCommandWithStdin` seems fine.

cmd/git-sync/main.go
hypnoglow: Since I've added `runCommandWithStdin` this is not needed anymore

cmd/git-sync/main.go
hypnoglow: Done

cmd/git-sync/main.go
```diff
@@ -217,8 +219,10 @@ func main() {
 		os.Exit(1)
 	}
 
+	ctx, cancel := context.WithTimeout(context.Background(), initTimeout)
```
thockin: Can you please comment the scope of this (just initialization, not clone, etc) and why 30s is "enough" ?

cmd/git-sync/main.go
```diff
@@ -232,12 +236,14 @@ func main() {
...
 
 	if *flCookieFile {
-		if err := setupGitCookieFile(); err != nil {
+		if err := setupGitCookieFile(ctx); err != nil {
 			fmt.Fprintf(os.Stderr, "ERROR: can't set git cookie file: %v\n", err)
 			os.Exit(1)
 		}
 	}
 
+	cancel()
```
thockin: Comment why calling this is important

cmd/git-sync/main.go
hypnoglow: Done

cmd/git-sync/main.go
hypnoglow: Done

Issue #191, opened 2019-10-25
Send git hash and branch in the HTTP header of the webhook


peick:
Example HTTP request could look like this:

```
POST / HTTP/1.1
Content-Length: 0
Accept-Encoding: gzip
User-Agent: Go-http-client/1.1
Git-Sync-Hash: f9b1e721c3086937f3d80536d03c6d5ee82eea26
Git-Sync-Branch: master
```

PR #193, opened 2019-10-25
Sent git hash and branch in the HTTP header of the webhook.

peick:
implements #191


cmd/git-sync/main.go
```diff
@@ -695,3 +691,27 @@ func setupGitCookieFile() error {
 
 	return nil
 }
+
+func triggerWebhook(ctx context.Context, ch chan webhookRepoInfo, branch, rev, gitRoot string) error {
+	info := webhookRepoInfo{
+		Branch: branch,
```
thockin: do we need branch or just hash?  Hash should be unique enough for anything.  Branch and tag are symbolic references

cmd/git-sync/main.go
```diff
@@ -695,3 +691,27 @@ func setupGitCookieFile() error {
...
+	if err != nil {
+		return err
+	}
+
+	info.Hash = hash
+
+	// Trigger webhooks to be called. We do a non-blocking write to the channel as we
+	// don't want to backup the syncing locally because we can't complete a webhook call.
+	// Since the channel has a buffer of 1 we ensure that it is called for a change, but
+	// this allows us to de-dupe calls if they happen before the webhook call completes.
```
thockin: This assumption (de-dup) was appropriate when this was merely a trigger.  If the receiver actually cares about the payload, though, this becomes an unreliable signal.  E.g. If there are 2 updates quickly, A then B, and A has not been fully delivered via webhook, the hook will get the message for A but never for B.

Options:

1) Document this and live with it.
2) Make the buffer bigger to reduce likelihood of this happening (e.g. 8) and then document it
3) Make the payload a shared object in memory, with a mutex.  Take the mutex when doing the webhook HTTP call.  Write the most-recent branch/hash to the object (with mutex) and then do this single-buffered queue.  That way if the chan has not been read, we can update the value to latest.  If it has been read in the meantime, we re-queue it.

#3 is the only "correct" answer, but I have only sketched the solution here.  Would need a really good test.

Thoughts?

cmd/git-sync/webhook.go
```diff
@@ -22,8 +22,15 @@ type Webhook struct {
...
 
-func (w *Webhook) Do() error {
+type webhookRepoInfo struct {
+	Hash   string
+	Branch string
+}
+
+func (w *Webhook) Do(info webhookRepoInfo) error {
 	req, err := http.NewRequest(w.Method, w.URL, nil)
+	req.Header.Set("Git-Sync-Branch", info.Branch)
```
thockin: nit: "Gitsync" since that is the proper name of this tool

cmd/git-sync/main.go
peick: I agree. Hash should be enough.

cmd/git-sync/webhook.go
peick: Ok.

cmd/git-sync/main.go
peick: Great observation. #3 should be the way to go. Otherwise we would loose the last hash information.

Will do so.

peick: I think we don't need a mutex here. But the last consumed value has to be compared with the new value. If we do it like this it will work the way expected:

```
package main

import (
        "fmt"
        "time"
)

type Data struct {
        new  int
        last int
        ch   chan struct{}
}

func (d *Data) UpdateAndTrigger(new int) {
        d.new = new

        select {
        case d.ch <- struct{}{}:
        default:
        }
}

func (d *Data) Next() int {
        for {
                <-d.ch
                new := d.new
                if new != d.last {
                        d.last = new
                        return new
                }
        }
}

func main() {
        d := Data{ch: make(chan struct{}, 1)}

        go receiver(&d)

        for i := 0; i < 10; i++ {
                fmt.Println("sending", i)
                d.UpdateAndTrigger(i)
        }

        time.Sleep(time.Second * 10)
        fmt.Println("done")
}

func receiver(d *Data) {
        for {
                i := d.Next()

                fmt.Println("got", i)
                time.Sleep(time.Second)
        }
}
```

cmd/git-sync/main.go
```diff
@@ -695,3 +692,20 @@ func setupGitCookieFile() error {
...
+
+func triggerWebhook(ctx context.Context, webhook *Webhook, rev, gitRoot, dest string) error {
+	target := path.Join(gitRoot, dest)
+
+	hash, err := hashForRev(ctx, rev, target)
+	if err != nil {
+		return err
+	}
+
+	// Trigger webhooks to be called. We do a non-blocking write to the channel as we
```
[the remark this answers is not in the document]
peick: Right. Using meaningful function names should be self explaining.

thockin:
Need you to sign that CLA, or we can't accept this PR :(

/lgtm
/approved

thockin:
Ahh, I cloned it last night to test and I wanted to send some comments.  My
bad, I guess I will just followup with a PR of my own.  Thanks for the PR

PR #197, opened 2019-10-26
Fix minor issues from linters

hypnoglow:
I've run some linters against the code, which revealed a few issues. This PR fixes them.

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -287,7 +287,7 @@ func main() {
 		start := time.Now()
 		ctx, cancel := context.WithTimeout(context.Background(), time.Second*time.Duration(*flSyncTimeout))
 		if changed, err := syncRepo(ctx, *flRepo, *flBranch, *flRev, *flDepth, *flRoot, *flDest); err != nil {
-			syncDuration.WithLabelValues("error").Observe(time.Now().Sub(start).Seconds())
+			syncDuration.WithLabelValues("error").Observe(time.Since(start).Seconds())
 			syncCount.WithLabelValues("error").Inc()
 			if *flMaxSyncFailures != -1 && failCount >= *flMaxSyncFailures {
 				// Exit after too many retries, maybe the error is not recoverable.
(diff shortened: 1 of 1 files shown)


cmd/git-sync/main.go
```diff
@@ -353,7 +354,7 @@ func setFlagDefaults() {
 // it is deadlocked.
 func sleepForever() {
 	c := make(chan os.Signal, 1)
-	signal.Notify(c, os.Interrupt, os.Kill)
+	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
```
hypnoglow: SIGKILL cannot be trapped

cmd/git-sync/main.go
```diff
@@ -527,7 +528,7 @@ func revIsHash(ctx context.Context, rev, gitRoot string) (bool, error) {
 func syncRepo(ctx context.Context, repo, branch, rev string, depth int, gitRoot, dest string) (bool, error) {
 	target := path.Join(gitRoot, dest)
 	gitRepoPath := path.Join(target, ".git")
-	hash := rev
```
hypnoglow: rationale: `switch` has `default` case where hash variable is assigned, so this assignement has no effect.

cmd/git-sync/main.go
```diff
@@ -659,7 +660,7 @@ func setupGitSSH(setupKnownHosts bool) error {
 	}
 
 	if setupKnownHosts {
-		_, err := os.Stat(pathToSSHKnownHosts)
+		_, err = os.Stat(pathToSSHKnownHosts)
```
hypnoglow: rationale: `err` assignment on 668 affects only `if` scope so error check on 674 won't see the error if it occurs.

cmd/git-sync/main.go
thockin: That was the point - it's not actually trying to do anything, just trick Go's runtime into thinking it is not hung.  This leaves "real" signals to be handled elsewhere if needed, which seems plausible for SIGINT.

I propose we nix this - I am thinking about changing this functionality anyway, so it won't matter long term.

cmd/git-sync/main.go
thockin: good one.

cmd/git-sync/main.go
thockin: good one

cmd/git-sync/main.go
hypnoglow: Ok, thanks for the explanation. Reverted.

PR #199, opened 2019-11-16
Add support for process signalling

bewing:
Add two new flags to support sending signals to named processes.
Utilize https://github.com/shirou/gopsutil to iterate through all processes
looking for matches.

Requires signalled process/container to be in same process namespace,
and have the capability to send signals to other processes.

Example k8s deployment:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test-deployment
  labels:
    app: test
spec:
  replicas: 1
  selector:
    matchLabels:
      app: test
  template:
    metadata:
      labels:
        app: test
    spec:
      shareProcessNamespace: true
      containers:
      - name: git-sync
        image: k8s.gcr.io/git-sync:build
        env:
          - name: GIT_SYNC_REPO
            value: "https://github.com/test-repo/test-repo"
          - name: GIT_SYNC_PROCESS
            value: telegraf
          - name: GIT_SYNC_WAIT
            value: "10"
        securityContext:
          runAsUser: 0
      - name: telegraf
        image: telegraf:latest
```

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -114,6 +114,11 @@ var flHTTPMetrics = flag.Bool("http-metrics", envBool("GIT_SYNC_HTTP_METRICS", t
 var flHTTPprof = flag.Bool("http-pprof", envBool("GIT_SYNC_HTTP_PPROF", false),
 	"enable the pprof debug endpoints on git-sync's HTTP endpoint")
 
+var flProcName = flag.String("process", envString("GIT_SYNC_PROCESS", ""),
+	"comm name to send signal on changes. Default is \"\" which disables signalling. Should be the first 15 characters of the command(s) to signal")
+var flProcSignal = flag.String("signal", envString("GIT_SYNC_SIGNAL", "SIGHUP"),
+	"signal name/number to send to matched processes.  Default is 1 (SIGHUP)")
+
 var log = glogr.New()
 
 // Total pull/error, summary on pull duration
@@ -283,6 +288,11 @@ func main() {
 		}
 	}
 
+	procSignal, err := ConvertSignal(*flProcSignal)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "ERROR: failed to convert signal: %v\n", err)
+	}
+
 	// The scope of the initialization context ends here, so we call cancel to release resources associated with it.
 	cancel()
 
@@ -355,8 +365,14 @@ func main() {
 			cancel()
 			time.Sleep(waitTime(*flWait))
 			continue
-		} else if changed && webhook != nil {
-			webhook.Send(hash)
+		} else {
+			if changed && webhook != nil {
+				webhook.Send(hash)
+			}
+			if changed && *flProcName != "" {
+				log.V(0).Info("Sending signals", "procname", *flProcName, "signal", procSignal)
+				go SignalProcs(*flProcName, procSignal)
+			}
 		}
 		syncDuration.WithLabelValues("success").Observe(time.Since(start).Seconds())
 		syncCount.WithLabelValues("success").Inc()
--- a/cmd/git-sync/signal.go
+++ b/cmd/git-sync/signal.go
@@ -1,0 +1,110 @@
+package main
+
+import (
+	"io/ioutil"
+	"os"
+	"strconv"
+	"strings"
+	"syscall"
+)
+
+var signalMap = map[string]int{
+	"SIGHUP":    1,
+	"SIGINT":    2,
+	"SIGQUIT":   3,
+	"SIGILL":    4,
+	"SIGTRAP":   5,
+	"SIGABRT":   6,
+	"SIGBUS":    7,
+	"SIGFPE":    8,
+	"SIGKILL":   9,
+	"SIGUSR1":   10,
+	"SIGSEGV":   11,
+	"SIGUSR2":   12,
+	"SIGPIPE":   13,
+	"SIGALRM":   14,
+	"SIGTERM":   15,
+	"SIGSTKFLT": 16,
+	"SIGCHLD":   17,
+	"SIGCONT":   18,
+	"SIGSTOP":   19,
+	"SIGTSTP":   20,
+	"SIGTTIN":   21,
+	"SIGTTOU":   22,
+	"SIGURG":    23,
+	"SIGXCPU":   24,
+	"SIGXFSZ":   25,
+	"SIGVTALRM": 26,
+	"SIGPROF":   27,
+	"SIGWINCH":  28,
+	"SIGIO":     29,
+	"SIGPWR":    30,
+	"SIGSYS":    31,
+}
+
+func ConvertSignal(flProcSignal string) (syscall.Signal, error) {
+	sig, err := strconv.ParseInt(flProcSignal, 10, 32)
+	if err == nil {
+		return syscall.Signal(sig), nil
+	} else {
+		if sig, ok := signalMap[flProcSignal]; ok {
+			return syscall.Signal(sig), nil
+		}
+	}
+	return syscall.Signal(1), nil
+}
+
+func SignalProcs(flProcName string, sig syscall.Signal) error {
+	pids, err := getPids()
+	if err != nil {
+		return err
+	}
+	for _, pid := range pids {
+		if getName(pid) == flProcName {
+			err := syscall.Kill(int(pid), sig)
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func getHostProc() string {
+	procDir, found := os.LookupEnv("HOST_PROC")
+	if !found {
+		procDir = "/proc"
+	}
+	return procDir
+}
+
+func getPids() ([]int32, error) {
+	var ret []int32
+	files, err := ioutil.ReadDir(getHostProc())
+	if err != nil {
+		return nil, err
+	}
+	for _, file := range files {
+		if file.IsDir() {
+			pid, err := strconv.ParseInt(file.Name(), 10, 32)
+			if err == nil {
+				ret = append(ret, int32(pid))
+			}
+		}
+	}
+	return ret, nil
+}
+
+func getName(pid int32) string {
+	statusFile := getHostProc() + "/" + strconv.Itoa(int(pid)) + "/status"
+	fileBytes, err := ioutil.ReadFile(statusFile)
+	if err != nil {
+		return ""
+	}
+	lines := strings.Split(string(fileBytes), "\n")
+	parts := strings.Split(lines[0], ":")
+	if parts[0] == "Name" {
+		return strings.Trim(parts[1], "\t ")
+	}
+	return ""
+}


bewing:
CLA signed.

thockin:
Hi, thanks for the PR.

Can you explain why we want this?  What use-case does it solve?

thockin:
Is this to send a signal when the sync happens?  

I'd like to discuss the need for this.  The dependencies are HUGE and I really do not want to take them.  I feel like we could write a simple adapter program that watches a symlink and when that symlink changes, delivers a signal to a process or pgrp or something.  This would be more general-purpose (could be used for kube projected volumes like configmap and secret, too) and composable.

bewing:
Correct, the desire is to signal to other containers that a change/sync has occurred, if they do not support methods such as inotify.  My use case would be configuration data for telegraf, and causing Telegraf to reload its configuration on change.

Agree that the dependency set is large -- would it be more feasible to just extract out the Linux-specific process for enumerating process names (looking at $HOST_PROC) and vendoring that?

Regarding the adapter concept:  Would you see that as a separate go thread in git-sync, or a second sidecar container in a pod with shared namespace?

thockin:
If the deps could be totally eliminated, I could MAYBE see it here, but I think a separate sidecar program might actually be a better overall solution - more reusable and simpler.

bewing:
Force-pushed new changeset that strips a bunch of gopsutil into pkg/ and uses that instead, which lowers the size of the changeset.

thockin:
Hey there!  First, thanks.  This is much closer to workable.  A few problems, and some design thoughts.

I am assuming this is taken from another package?  We can't simply copy it like this.  We really have 2 options:
1) vendor the dependency as it exists upstream
2) put it in a third_party/ directory which is like vendor/ but can be sparse and can be modified.  E.g. `third_party/github.com/shirou/gopsutil` (and include a README.md in there saying "Forked from <github.com/shirou/gopsutil@<sha> because <explain why...>"

If we can keep the deps small, I am OK with adding this feature.  From a UX perspective, some thoughts.

1) flag `--signal-process-name`, where the value is "literal-process-name".  If we REALLY need it, maybe allow "prefix-*", meaning anything starting with "prefix-".  How should we handle full paths vs not?  E.g. "bash" vs "/bin/bash" ?

2) flag `--signal` should probably take either a signal name (SIGHUP, SIGINT) or a number, since the numbers are actually platform dependent.

bewing:
Thanks for taking the time to review!  This is my first Golang project, so I knew it would take a lot of patience on the part of the maintainers.

gopsutil is the upstream package -- my first changeset (https://github.com/kubernetes/git-sync/commit/1e6cdd6bc6f4ae8b6710516fda7cdf9c7b21fb8c) did just import and use it, but the large number of other dependencies it introduces due to attempting to be multi-platform pushed the changeset into the XXL range.  I'd be happy to update this changeset to move the copied code into /third_party as recommended with a LICENSE and README.md file if we think that's the best approach.  Would it be better if I either found a more basic, Linux-only psutil package to import?  Is it feasible to have myself re-write using procps to identify the necessary processes to signal, or am I "tainted" having read all of the gopsutil source?  I'm not 100% on the licensing impact of that.

Regarding command line options and documentations, perhaps something like `--signal-process-cmdline`, and a help message indicating an exact match against the `cmdline` entry of a process?  That would leave the burden on the end-user who is using this to signal other containers in their own environment.  

Agree that user experience would be better with either signal name or number, but had the concern that signal number -> name mapping changed from platform to platform.  If we're comfortable just saying "signal names for x86 platforms will be correctly translated", we can make a simple map, otherwise we'd either have to figure out dynamically the name->integer mapping, or include different maps at compile-time.

thockin:
> I'd be happy to update this changeset to move the copied code into /third_party as recommended

This would be acceptable

> Would it be better if I either found a more basic, Linux-only psutil package to import?

Also acceptable (and preferable as it is more "normal").

> Is it feasible to have myself re-write using procps to identify the necessary processes to signal, or 
> am I "tainted" having read all of the gopsutil source? I'm not 100% on the licensing impact of that.

No, you are not tainted :)

> Regarding command line options and documentations, perhaps something like 
> --signal-process-cmdline,

I think that is overly pedantic.  The most common use cases for git-sync are in containers, so there are relatively few processes.

The problem I was asking about is that tools like `pidof` sometimes produce different results for full paths and not - we should be precise in the specification of behavior

> Agree that user experience would be better with either signal name or number, but had the concern 
> that signal number -> name mapping changed from platform to platform.

You'll have to look it up.  https://godoc.org/golang.org/x/sys/unix#SignalNum

bewing:
Third changeset pushed.  Re-wrote the signaling code, and also updated the help string to attempt to clarify that we are exactly matching the comm name (first 15 characters of executable) to signal.

cmd/git-sync/signal.go
```diff
@@ -0,0 +1,110 @@
...
+func ConvertSignal(flProcSignal string) (syscall.Signal, error) {
+	sig, err := strconv.ParseInt(flProcSignal, 10, 32)
+	if err == nil {
+		return syscall.Signal(sig), nil
+	} else {
+		if sig, ok := signalMap[flProcSignal]; ok {
+			return syscall.Signal(sig), nil
+		}
+	}
+	return syscall.Signal(1), nil
```
thockin, line 54: This will convert "unparseable text string" into "1".  There's a default value for a  reason.  If we can't parse it as int and we can't match it by name, it's an error

thockin: And make sure the error messge is useful is I pass "foobar" as a signal.

cmd/git-sync/main.go
```diff
@@ -355,8 +365,14 @@ func main() {
 			cancel()
 			time.Sleep(waitTime(*flWait))
 			continue
-		} else if changed && webhook != nil {
-			webhook.Send(hash)
+		} else {
```
thockin, line 368: We don't really need the else - the if end in a continue.

cmd/git-sync/main.go
```diff
@@ -355,8 +365,14 @@ func main() {
...
+			if changed && webhook != nil {
+				webhook.Send(hash)
+			}
+			if changed && *flProcName != "" {
+				log.V(0).Info("Sending signals", "procname", *flProcName, "signal", procSignal)
```
thockin, line 373: log in side the called function, like webhook does.

cmd/git-sync/signal.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func SignalProcs(flProcName string, sig syscall.Signal) error {
+	pids, err := getPids()
+	if err != nil {
+		return err
+	}
+	for _, pid := range pids {
+		if getName(pid) == flProcName {
```
thockin, line 63: Do we really need a second-check?  I assume this is defending against races, but it's still a race, so this is wasted energy.  Or is there some other reason?

cmd/git-sync/signal.go
```diff
@@ -0,0 +1,110 @@
...
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func getHostProc() string {
+	procDir, found := os.LookupEnv("HOST_PROC")
```
thockin, line 74: This is not a standard that I know of?  If not, it should be be prefixed with `GIT_SYNC_`, eg GIT_SYNC_PROCFS_PATH

cmd/git-sync/signal.go
```diff
@@ -0,0 +1,110 @@
...
+		return syscall.Signal(sig), nil
+	} else {
+		if sig, ok := signalMap[flProcSignal]; ok {
+			return syscall.Signal(sig), nil
+		}
+	}
+	return syscall.Signal(1), nil
+}
+
+func SignalProcs(flProcName string, sig syscall.Signal) error {
```
thockin, line 57: `fl` prefix is for global-ish flags.  just call it procName?

cmd/git-sync/signal.go
```diff
@@ -0,0 +1,110 @@
...
+	pids, err := getPids()
```
thockin, line 58: Simpler, I think (?) to pass procName down o getPids

cmd/git-sync/signal.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func getPids() ([]int32, error) {
+	var ret []int32
+	files, err := ioutil.ReadDir(getHostProc())
+	if err != nil {
+		return nil, err
+	}
+	for _, file := range files {
+		if file.IsDir() {
```
thockin, line 88: Comments here would be nice

cmd/git-sync/signal.go
```diff
@@ -0,0 +1,110 @@
...
+			if err == nil {
+				ret = append(ret, int32(pid))
+			}
+		}
+	}
+	return ret, nil
+}
+
+func getName(pid int32) string {
+	statusFile := getHostProc() + "/" + strconv.Itoa(int(pid)) + "/status"
```
thockin, line 99: status and comm both truncate to 15 characters, which is weak.  If we're going to do this, lets do it right.

Parse /proc/pid/cmdline (split on NUL, I think) and match either a full path (if any slash in the flag) or (if no slash) on the basename of the contents of cmdline.

cmd/git-sync/main.go
```diff
@@ -114,6 +114,11 @@ var flHTTPMetrics = flag.Bool("http-metrics", envBool("GIT_SYNC_HTTP_METRICS", t
 var flHTTPprof = flag.Bool("http-pprof", envBool("GIT_SYNC_HTTP_PPROF", false),
 	"enable the pprof debug endpoints on git-sync's HTTP endpoint")
 
+var flProcName = flag.String("process", envString("GIT_SYNC_PROCESS", ""),
+	"comm name to send signal on changes. Default is \"\" which disables signalling. Should be the first 15 characters of the command(s) to signal")
```
thockin, line 118: See below - 15 chars is weak :)

cmd/git-sync/signal.go
```diff
@@ -0,0 +1,110 @@
...
+
+func getHostProc() string {
+	procDir, found := os.LookupEnv("HOST_PROC")
+	if !found {
+		procDir = "/proc"
+	}
+	return procDir
+}
+
+func getPids() ([]int32, error) {
```
thockin, line 81: All the helper functions in here can be rigorously unit tested, please.

Then we need to think about e2e tests (see test_e2e.sh)

cmd/git-sync/signal.go
```diff
@@ -0,0 +1,110 @@
...
+
+import (
+	"io/ioutil"
+	"os"
+	"strconv"
+	"strings"
+	"syscall"
+)
+
+var signalMap = map[string]int{
```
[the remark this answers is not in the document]
bewing: Would it be acceptable to add `x/sys/unix` to `go.mod` to accomplish this?  It appears that they define lists of signals per architecture there: https://github.com/golang/sys/blob/master/unix/zerrors_linux_amd64.go#L748

thockin:
Yes, that is a dep I am happy to accept. :)

cmd/git-sync/signal.go
bewing: Will work on this.  Will probably have to refactor a bit to provide testable interfaces.

cmd/git-sync/signal.go
bewing: I guess I saw it in Telegraf, and thought it made enough sense that it was a standard.  Agree that it's not after future research, will modify.

bewing:
Apologies -- ran into some computer issues, lost some work, and need to revisit.

thockin:
No sweat!  I'll just take it out of your paycheck.  Oh, this is all
volunteer -- whatever you get to is fine, then :)

thockin:
Hi @bewing - if you find time to spend on this, I'd be happy to see it re-opened.  No pressure.

PR #213, opened 2019-12-02
Use git mirror

### LINKED ISSUE #202: Use git --mirror
I have found that using git --mirror makes it easier to resolve symbolic revs.  We can then get rid of the --branch flag and JUST keep --rev.

I propose a v4 with breaking changes (flags)

thockin:
I propose this be the start of the v4 effort.  It is a major revamp of internal sync mechanism, with breaking changes.

`git --mirror` does not check out one branch, but the whole repo.  This makes it easier to understand any string as a branch/tag/hash, just like `git checkout`.  The obvious side-effect here is that we don't need `--branch` and `--rev` -- just `--rev` is sufficient.  This PR gets rid of the `--branch` flag (breaking change).

This also rethinks how error-recovery is done at startup, since the checkout mechanism is different now.

The main downside of this is that it makes --depth somewhat less obviously useful.  I think this is a worthwhile change, but I will see if I can do better yet (hence the hold).

Fixes #202


cmd/git-sync/main.go
```diff
@@ -569,66 +563,108 @@ func revIsHash(ctx context.Context, rev, gitRoot string) (bool, error) {
...
-		// Not the first time. Figure out if the ref has changed.
-		local, remote, err := getRevs(ctx, target, branch, rev)
-		if err != nil {
-			return false, "", err
-		}
-		if local == remote {
-			log.V(1).Info("no update required", "rev", rev, "local", local, "remote", remote)
-			return false, "", nil
+		// Make sure the directory we found is actually usable.
+		if err := sanityCheck(ctx, gitRoot, dest); err != nil {
```
stp-ip: Running the sanityCheck every time feels quite expensive. We are running this on each sync run even if nothing changes aka local == remote.
I see the reasoning, but this might be problematic especially considering the normal use case of it being run in a small sidecar.
What was the specific thinking here? Validity and better error recovery > performance considerations?

cmd/git-sync/main.go
```diff
@@ -569,66 +563,108 @@ func revIsHash(ctx context.Context, rev, gitRoot string) (bool, error) {
...
+	if _, err := runCommand(ctx, gitRoot, *flGitCmd, args...); err != nil {
+		return err
 	}
 
-	// Figure out what hash the remote resolves ref to.
-	remote, err := remoteHashForRef(ctx, ref, localDir)
-	if err != nil {
-		return "", "", err
+	// GC the repo.
+	if _, err := runCommand(ctx, gitRoot, *flGitCmd, "gc", "--prune=all"); err != nil {
```
stp-ip:
```suggestion
	if _, err := runCommand(ctx, gitRoot, *flGitCmd, "gc", "--prune=now"); err != nil {
```

stp-ip: What are you trying to clean up with gc? Stale worktrees? Wouldn't that need a "worktree remove" beforehand or else it never becomes stale?

stp-ip:
Waiting for a response to the review here. @thockin

stp-ip:
Let me know, when you want me to take another stab/look at this.

thockin:
I should close this and open a new one when ready


/close

PR #216, opened 2019-12-17
Support GIT_ASKPASS_URL callback

cydu-cloud:
GIT_ASKPASS_URL callback support external HTTP Service to provide password and username to access Git Repo.

See https://github.com/cydu-cloud/git-askpass-gce-node as end to end test which enables git-sync access GCP Cloud Source Repo with service account's oauth bear token. 

See https://git-scm.com/docs/gitcredentials for more about the GIT_ASKPASS protocols.


cydu-cloud:
I signed it

cydu-cloud:
I signed it

docs/askpass-url.md
```diff
@@ -0,0 +1,40 @@
...
+## Step 2: Configure git-sync container
+
+In your git-sync container configuration, specify the GIT_ASKPASS_URL
+
+The credentials will pass in plain text, make sure the connection between git-sync
+and GIT ASKPASS Service are secure.
+
+The recommended way is the ASKPASS Service running within the same pod as git-sync.
+
+See <https://github.com/cydu-cloud/git-askpass-gce-node> as a full example which use GCE Node Service Account credential to access Google Cloud Source Repo.
```
rcolline: We should find a better place to put this.. maybe GoogleCloudPlatform's repo?

thockin: Can we cook down an e2e test that can be set up and run locally?  E.g. spin up a git server and an HTTP server, tell git-sync to pull the password from the HTTP server.

test_e2e.sh should cover this

docs/askpass-url.md
```diff
@@ -0,0 +1,40 @@
+# Using an Http Auth URL with git-sync
+
+## Step 1: Create a GIT_ASKPASS HTTP Service
+
+The GIT ASKPASS Service expose via HTTP and provide the answer to GIT_ASKPASS.
+
+Example of the servcie's output, see more at <https://git-scm.com/docs/gitcredentials>
```
thockin: typo: service

docs/askpass-url.md
```diff
@@ -0,0 +1,40 @@
+# Using an Http Auth URL with git-sync
+
+## Step 1: Create a GIT_ASKPASS HTTP Service
+
+The GIT ASKPASS Service expose via HTTP and provide the answer to GIT_ASKPASS.
```
thockin: is exposed

provides

docs/askpass-url.md
```diff
@@ -0,0 +1,40 @@
...
+The credentials will pass in plain text, make sure the connection between git-sync
+and GIT ASKPASS Service are secure.
+
+The recommended way is the ASKPASS Service running within the same pod as git-sync.
+
+See <https://github.com/cydu-cloud/git-askpass-gce-node> as a full example which use GCE Node Service Account credential to access Google Cloud Source Repo.
+
+```json
+{
+    name: "git-sync",
```
thockin: json requires quoted keys

docs/askpass-url.md
```diff
@@ -0,0 +1,40 @@
...
+```
+
+## Step 2: Configure git-sync container
+
+In your git-sync container configuration, specify the GIT_ASKPASS_URL
+
+The credentials will pass in plain text, make sure the connection between git-sync
+and GIT ASKPASS Service are secure.
+
+The recommended way is the ASKPASS Service running within the same pod as git-sync.
```
thockin: worth mentioning http://metadata?

docs/cookie-file.md
```diff
@@ -2,27 +2,27 @@
...
 
-# Step 2: Configure Pod/Deployment Volume
+## Step 2: Configure Pod/Deployment Volume
 
 In your Pod or Deployment configuration, specify a Volume for mounting the
 cookie-file Secret. Make sure to set `secretName` to the same name you used to
 create the secret (`git-cookie-file` in the example above).
 
-```
+```json
```
thockin: This is valid YAML but not JSON.  To be JSON we'd need to quote the key

docs/cookie-file.md
```diff
@@ -34,15 +34,15 @@ volumes: [
...
 In your git-sync container configuration, mount your volume at
 "/etc/git-secret". Make sure to pass the `--cookie-file` flag or set the
 environment variable `GIT_COOKIE_FILE` to "true", and to use a git repo
 (`--repo` flag or `GIT_SYNC_REPO` env) is set to use a URL with the HTTP
 protocol.
 
-```
+```json
 {
     name: "git-sync",
```
thockin: quoted keys

cydu-cloud: Changed to real yaml.

docs/askpass-url.md
cydu-cloud: For the e2e test, it depends on askpass_git test I just created in another PR, don't want the PR grow into too big, can I add it as follow up PR? This won't hurt any existing functionality even the callbalk is broken.

For GoogleCloudPlatform's repo, it might take a while to setup, need to go through some process.

thockin: I don't want to e2e against GCP.  I want to set up a local git repo that demands user/pass and a local HTTP server which serves the content.  I am working on a more complex SSH example - will link it here, hopefully helps you with infra

thockin:
See https://github.com/kubernetes/git-sync/pull/218 for e2e infra ideas (last commit)

docs/askpass-url.md
cydu-cloud: For "local git repo that demands user/pass", trying to use askpass_git as I did for user/pass e2e test, the `git credential fill` is exactly what git do when demands user/pass, so the simulation should be very good enough for the test here.  
Removed the reference docs to gcp. It's purely local e2e test right now.

Thanks for the SSH example, I can not image any better way to do it for SSH, but for user/pass, I personally prefer askpass_git a bit more since it's light and easier to debug/maintain. 

Make sense? 

Sorry for the growing PR, I just manually merged the askpass_git from https://github.com/kubernetes/git-sync/pull/217.

test_e2e.sh
```diff
@@ -621,6 +624,93 @@ remove_sync_container
...
+assert_file_eq "$ROOT"/link/file "$TESTCASE 1"
+# Wrap up
+pass
+
+##############################################
+# Test askpass_url
+##############################################
+testcase "askpass_url"
+echo "$TESTCASE 1" > "$REPO"/file
+NCPORT=8888
```
thockin: Can we do this without hard-coding a port?  e.g. either allocate it randomly or run it in a docker container (as with the SSH example).

cydu-cloud: Allocated it randomly.

test_e2e.sh
```diff
@@ -666,11 +666,56 @@ assert_file_eq "$ROOT"/link/file "$TESTCASE 1"
 # Wrap up
 pass
 
+##############################################
+# Test askpass_url
+##############################################
+testcase "askpass_url"
+echo "$TESTCASE 1" > "$REPO"/file
+NCPORT=$(($RANDOM+2000))
```
thockin: Sorry, this is not what I meant by random.  This has a higher probability of success, but could randomly fail e2e tests.

I don't now if there's a better CLI way to bind to a kernel-assigned (never in conflict) port number, or if you need to run a dedicated container image..

cydu-cloud: Looks like it's feasible to find a free port via bash.

test_e2e.sh
```diff
@@ -52,6 +52,14 @@ function assert_file_eq() {
     fail "file $1 does not contain '$2': $(cat $1)"
 }
 
+NCPORT=8888
+function freencport() {
```
thockin: This is gross and racy and not guaranteed to work, but diminishes the risk to the point that I am not willing to argue about it any more.

cydu-cloud: Thank you!

PR #217, opened 2019-12-18
Change git credential from cache to store and also add test for username/password case

### LINKED ISSUE #196: Need a test for username/password
We claim to support it but have no e2e

cydu-cloud:
For cache to store change:
* By default, `cache` only last 900 seconds, gitsync will break after
  that for password setup. See https://git-scm.com/docs/git-credential-cache.
* The test won't work with cache since the test don't have access to
  the default unix socket location; XDG_CACHE_HOME override also cann't
  pre-create a socket in advance.
* `store` put the credential into a file, much easier to debug than cache.
* Considering anyone have access to the pod already able to get the
  credential via environment variables or yaml configs, so put it in
  file won't make it less secure.

For the new password test:
*  `askpass_git.sh` provided to simulate a git with password challenge.
*  Need and only need to challenge for "clone" action. Need to bypass all the other
  actions like config/credential setup.
*  See `credential fill` is the official git action to ask password,
  see https://git-scm.com/docs/git-credential.

Fixes #196.

### DIFF
--- a/askpass_git.sh
+++ b/askpass_git.sh
@@ -1,0 +1,25 @@
+#!/bin/sh
+# Ask pass when cloning new repo, fail if it mismatched the magic password.
+
+mkdir -p "${XDG_CONFIG_HOME}/git/"
+# Override the default 'git --global' config location, the default location
+# outside the e2e test environment. See https://git-scm.com/docs/git-config
+touch "${XDG_CONFIG_HOME}/git/config"
+# Override the default 'git credential store' config location, the default location
+# outside the e2e test environment. See https://git-scm.com/docs/git-credential-store
+touch "${XDG_CONFIG_HOME}/git/credentials"
+
+if [ "$1" != "clone" ]; then
+  git "$@"
+  exit $?
+fi
+
+REPO=$(echo "$@" | grep -o "file://[^ ]*")
+PASSWD=$(echo "url=${REPO}" | git credential fill | grep -o "password=.*")
+# Test case much match the magic password below.
+if [ "${PASSWD}" != "password=Lov3!k0os" ]; then
+  echo "invalid password ${PASSWD}, try Lov3!k0os next time."
+  exit 1
+fi
+
+git "$@"
--- a/askpass_git.sh
+++ b/askpass_git.sh
@@ -14,6 +14,8 @@ if [ "$1" != "clone" ]; then
   exit $?
 fi
 
+# `git credential fill` requires the repo url match to consume the credentials stored by git-sync.
+# Askpass git only support repo started with "file://" which is used in test_e2e.sh.
 REPO=$(echo "$@" | grep -o "file://[^ ]*")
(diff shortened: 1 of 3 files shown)


test_e2e.sh
```diff
@@ -621,6 +624,52 @@ remove_sync_container
...
+    --password="I have no idea what the password is." \
+    --logtostderr \
+    --v=5 \
+    --wait=0.1 \
+    --repo="file://$REPO" \
+    --branch=master \
+    --rev=HEAD \
+    --root="$ROOT" \
+    --dest="link" \
+    > "$DIR"/log."$TESTCASE" 2>&1 &
```
thockin: You can change this to not run in the background and not sleep if you switch to `--one-time`

test_e2e.sh
```diff
@@ -621,6 +624,52 @@ remove_sync_container
...
+    --password="Lov3!k0os" \
+    --logtostderr \
+    --v=5 \
+    --wait=0.1 \
+    --repo="file://$REPO" \
+    --branch=master \
+    --rev=HEAD \
+    --root="$ROOT" \
+    --dest="link" \
+    > "$DIR"/log."$TESTCASE" 2>&1 &
```
thockin: Same here

askpass_git.sh
```diff
@@ -0,0 +1,25 @@
...
+# Override the default 'git credential store' config location, the default location
+# outside the e2e test environment. See https://git-scm.com/docs/git-credential-store
+touch "${XDG_CONFIG_HOME}/git/credentials"
+
+if [ "$1" != "clone" ]; then
+  git "$@"
+  exit $?
+fi
+
+REPO=$(echo "$@" | grep -o "file://[^ ]*")
```
thockin: Comment here?  It took me a minute to realize this will try to consume the credentials that git-sync stored before.

askpass_git.sh
```diff
@@ -0,0 +1,27 @@
...
+  git "$@"
+  exit $?
+fi
+
+# `git credential fill` requires the repo url match to consume the credentials stored by git-sync.
+# Askpass git only support repo started with "file://" which is used in test_e2e.sh.
+REPO=$(echo "$@" | grep -o "file://[^ ]*")
+PASSWD=$(echo "url=${REPO}" | git credential fill | grep -o "password=.*")
+# Test case much match the magic password below.
+if [ "${PASSWD}" != "password=Lov3!k0os" ]; then
```
thockin: should we also test username?

cydu-cloud: We can, but don't think it's necessary, the magic password can match already approved the whole process working as intended.

thockin:
If there were a bug and we didn't feed in the username properly, we would
not see it...

cydu-cloud:
Yeah, that’s possible in theory,
I can file a follow up PR to add the username check if you have strong opinion about it.

thockin:
If it's not a big deal, it would be one less thing in the back of my mind.

PR #218, opened 2019-12-19
Support ssh as arbitrary users, add e2e for SSH

### LINKED ISSUE #176: git-sync doesn't work under arbitrary user
I'm trying to get the container to work under an arbitrary user ID, since this is a requirement in OpenShift depending on the cluster policies. I keep getting this error:

/git-sync -repo https://github.com/some-repo -username some-username -password *****
ERROR: can't create .netrc file: error setting up git credentials exit status 255: error: could not lock config file //.gitconfig: Permission denied

The usual course of action would be to make some files group-writable by the root group, but I can't understand what files I need to change - the /tmp directory is writable by any user and the .gitconfig file doesn't exist yet.

Would be grateful for any guidance!

### LINKED ISSUE #195: Need a test for SSH mode
We claim to support it but have no e2e

### LINKED ISSUE #215: SSH private key file permissions are too open
https://github.com/kubernetes/git-sync/blob/master/docs/ssh.md

**Problem**
Permissions in this file refer to 0440, but it should be changed to 0600 for example

**Current Behavior:**
@         WARNING: UNPROTECTED PRIVATE KEY FILE!          @
Permissions 0440 for '/root/.ssh/id_ed25519' are too open.
It is required that your private key files are NOT accessible by others.
This private key will be ignored.

**Suggested change:**
change permissions to 0600

thockin:
This adds a --add-user flag to write UID and GID to /etc/passwd, and then adds a test for `--ssh`.  Hopefully this can be a blueprint for other test cases.

Fixes #176 

Fixes #195 

Fixes #215

### DIFF
--- a/_test_tools/sshd/Dockerfile
+++ b/_test_tools/sshd/Dockerfile
@@ -1,0 +1,39 @@
+# Stolen from https://github.com/linuxkit/linuxkit/tree/master/pkg/sshd/
+
+FROM alpine AS base
+
+RUN mkdir -p /out/etc/apk && cp -r /etc/apk/* /out/etc/apk/
+RUN apk add --no-cache --initdb -p /out \
+    alpine-baselayout \
+    apk-tools \
+    busybox \
+    ca-certificates \
+    git \
+    musl \
+    openssh-server \
+    tini \
+    util-linux \
+    wireguard-tools \
+    && true
+
+###############
+
+FROM scratch
+
+ENTRYPOINT []
+WORKDIR /
+
+COPY --from=base /out/ /
+
+RUN mkdir -p /etc/ssh && rm /etc/motd
+COPY sshd_config /etc/ssh/
+COPY sshd.sh /
+
+# Callers should mount a .ssh directory here. Our sshd.sh will copy it and
+# manage permissions.
+VOLUME /dot_ssh
+
+# Callers can SSH as user "test"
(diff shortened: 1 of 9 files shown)


thockin:
PTAL

thockin:
I want to cut a release today/tomorrow and I'd like this in, if you can find 10 minutes to review.  Hopefully easy.

_test_tools/sshd/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+RUN mkdir -p /etc/ssh && rm /etc/motd
+COPY sshd_config /etc/ssh/
+COPY sshd.sh /
+
+# Callers should mount a .ssh directory here. Our sshd.sh will copy it and
+# manage permissions.
+VOLUME /dot_ssh
+
+# Callers can SSh as user "test"
```
cydu-cloud: nit: SSH

stp-ip:
Small nit as suggested by @erbesharat:
Using os.Current() instead of 3 os.Getuid(), os.Getgid() and os.Getenv() should reduce the footprint of the syscalls.

thockin:
These calls happen exactly once (and on linux, they should be a vsyscall anyway, if it is supported).

stp-ip:
Agreed not necessary, just an interesting suggestion nonetheless.

/lgtm
/approve

PR #219, opened 2019-12-20
Verify both username and password in the askpass_git test.

cydu-cloud:
As promised.


cydu-cloud:
/assign @thockin 

Still assign to Tim, please see https://github.com/kubernetes/git-sync/pull/217#issuecomment-567362384 for the background of this cl. Thanks.

PR #221, opened 2019-12-30
E2E fixes and cleanup

thockin:
In prep for an SSH test.


thockin:
Manually merging because tests are broken right now

PR #223, opened 2020-01-07
Update deployment to use new version which is multi-arch

dmarkey:
So it can run on a Raspberry Pi


thockin:
Thanks!  Need to get the CLA signed though :)

/lgtm
/approve

dmarkey:
Ok, I signed up. How do it get it to re-run the check?

PR #228, opened 2020-01-23
Fix typo in readme

rlukin:


rlukin:
I signed it

stp-ip:
This is not a type. The directory is atomically switched aka in one go aka either it succeeds or it fails.

Thanks for the contribution, but as this is working as intended, I'll close this for now.

PR #232, opened 2020-02-10
added table for env vars

jimangel:
I was looking for a table like this containing all valid env vars when getting started. I created this content from reviewing `main.go`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,4 +76,38 @@ docker run -d \
         --webhook-url="http://localhost:9090/-/reload"
 ```
 
+## Parameters
+
+| Environment Variable            | Flag                       | Description                                                                                                                            | Default                       |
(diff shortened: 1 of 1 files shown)


thockin:
This is awesome!  Thanks.

For some reason, the env vars are what many/most people use.  I personally find the flags to be easier.  What would you think about adding flag names here, too ?

thockin:
In fact, what if help could show this, e.g.:

```
  -depth int
    	[env: GIT_SYNC_DEPTH] use a shallow clone with a history truncated to the specified number of commits
  -repo string
    	[env: GIT_SYNC_REPO] the git repository to clone
  -rev string
    	[env: GIT_SYNC_REV] the git revision (tag or hash) to check out (default "HEAD")
```

jimangel:
Good with either solution. I updated the PR to include flags, but a help that dumps all this stuff would be nice too!

PR #240, opened 2020-03-24
Add support for additional git configs

### LINKED ISSUE #235: Support using custom ca(self-signed) for interacting with git
There are some usecases where git is using a self-signed certificate
and the git command supports specifying the cert file to use.

Support using self-singed cert for interacting with git

or support skip tlsVerify in git (this may be not secure)

onesolpark:
Fixes #235 
Add support for adding additional git configs through 
args `--git-configs=k1:v1,k2:v2` or through env variable `GIT_CONFIGS` 

All key/values are added as global values to gitconfig

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -114,6 +114,9 @@ var flHTTPMetrics = flag.Bool("http-metrics", envBool("GIT_SYNC_HTTP_METRICS", t
 var flHTTPprof = flag.Bool("http-pprof", envBool("GIT_SYNC_HTTP_PPROF", false),
 	"enable the pprof debug endpoints on git-sync's HTTP endpoint")
 
+var flGitConfigs = flag.String("git-configs", envString("GIT_CONFIGS", ""),
+	"add additional git configs in 'key1:value1,key2:value2' format (ex - http.sslCAInfo:/path/to/cert/file,http.version:HTTP/2)")
+
 var log = glogr.New()
 
 // Total pull/error, summary on pull duration
@@ -279,6 +282,13 @@ func main() {
 	if *flAskPassURL != "" {
 		if err := setupGitAskPassURL(ctx); err != nil {
 			fmt.Fprintf(os.Stderr, "ERROR: failed to call ASKPASS callback URL: %v\n", err)
+			os.Exit(1)
+		}
+	}
+
+	if *flGitConfigs != "" {
+		if err := setupGitConfigs(ctx); err != nil {
+			fmt.Fprintf(os.Stderr, "ERROR: can't set additional git configs: %v\n", err)
 			os.Exit(1)
 		}
 	}
@@ -879,6 +889,26 @@ func setupGitAskPassURL(ctx context.Context) error {
 
 	if err := setupGitAuth(ctx, username, password, *flRepo); err != nil {
 		return fmt.Errorf("error setup git auth: %v", err)
+	}
+
+	return nil
+}
+
+
+func setupGitConfigs(ctx context.Context) error {
+	log.V(1).Info("adding git configs from flag")
+
+	for _, config := range strings.Split(*flGitConfigs, ",") {
+		keyValues := strings.SplitN(config, ":", 2)
+		if len(keyValues) != 2 {
+			return fmt.Errorf("git configs must be in `key1:value1,key2:value2` format")
+		}
+
+		if _, err := runCommand(ctx, "",
+			*flGitCmd, "config", "--global", keyValues[0], keyValues[1]); err != nil {
+			return fmt.Errorf("error configuring git configs %s %s: %v", keyValues[0], keyValues[1], err)
+		}
+
 	}
 
 	return nil
--- a/README.md
+++ b/README.md
@@ -109,5 +109,6 @@ docker run -d \
 | GIT_SYNC_HTTP_BIND              | `--http-bind`              | the bind address (including port) for git-sync's HTTP endpoint                                                                         | ""                            |
 | GIT_SYNC_HTTP_METRICS           | `--http-metrics`           | enable metrics on git-sync's HTTP endpoint                                                                                             | true                          |
 | GIT_SYNC_HTTP_PPROF             | `--http-pprof`             | enable the pprof debug endpoints on git-sync's HTTP endpoint                                                                           | false                         |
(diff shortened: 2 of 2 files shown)


cmd/git-sync/main.go
```diff
@@ -114,6 +114,9 @@ var flHTTPMetrics = flag.Bool("http-metrics", envBool("GIT_SYNC_HTTP_METRICS", t
 var flHTTPprof = flag.Bool("http-pprof", envBool("GIT_SYNC_HTTP_PPROF", false),
 	"enable the pprof debug endpoints on git-sync's HTTP endpoint")
 
+var flGitConfigs = flag.String("git-configs", envString("GIT_CONFIGS", ""),
+	"add additional git configs in 'key1:value1,key2:value2' format (ex - http.sslCAInfo:/path/to/cert/file,http.version:HTTP/2)")
```
thockin: nit: change "ex - " to "e.g. "

cmd/git-sync/main.go
```diff
@@ -283,6 +286,13 @@ func main() {
 		}
 	}
 
+	if *flGitConfigs != "" {
+		if err := setupGitConfigs(ctx); err != nil {
```
thockin: please pass the flag in here so that only main() deals with flags (mostly)

cmd/git-sync/main.go
```diff
@@ -883,3 +893,23 @@ func setupGitAskPassURL(ctx context.Context) error {
 
 	return nil
 }
+
+
+func setupGitConfigs(ctx context.Context) error {
+	log.V(1).Info("adding git configs from flag")
+
+	for _, config := range strings.Split(*flGitConfigs, ",") {
```
thockin: Are there any git-config keys or values that embed commas?  I know it's weird, but if a path had a comma in it, this would break.

This also applies to colons.  Simple string split probably is not enough.

A more robust approach would be a function that walks through the input string, one rune at a time and produces a `[]struct{k, v string}`, which you can then iterate over safely.  As you walk the string, any time you find a literal backslash, replace it with whatever comes next.

"\a\b\c" => "abc"
"foo\n" => "foon"
"hello\\there" => "hello\there"
"/weird/path/with/\,comma" => "/weird/path/with/,comma"

Or are we SURE that we never have to deal with this?

cmd/git-sync/main.go
```diff
@@ -283,6 +286,13 @@ func main() {
 		}
 	}
 
+	if *flGitConfigs != "" {
```
thockin: Comment here that this should be the very last config section, so it overrides all others.

cmd/git-sync/main.go
thockin: Then we want a unit test :)

onesolpark: As you pointed out, I believe I can't be sure that there won't be any commas and semicolons in these variables.

I'll change the function to go over rune at a time and escape based on backslashes.

The thing that concerns me is that you can actually put backslashes in gitconfig with double backslash `\\`

And if users want to put double backslash it would mean that they have to put `\\\\` with the function logic. 

`"\a\b\c"` => `"abc"`
`"foo\n"` => `"foon"`
`"hello\there"` => `"hellothere"`
`"/weird/path/with/\,comma"` => `"/weird/path/with/,comma"`

`"weird\\\\backslashid"` => `"weird\\backslashid"`

Does this sound fine?

onesolpark: I'll definitely add a unit test to check this logic :)

thockin: This got long.  I spent a lot of time playing with git-config, finding edge-case behaviors. 
 PLEASE double check me.

Looking at the git-config docs:

```
Section names are case-insensitive. Only alphanumeric characters, - and . are allowed in section names.
```

and 

```
The variable names are case-insensitive, allow only alphanumeric characters and -, and must start with an alphabetic character.
```

So keys can not hold comma or colon.  if you walk the string and parse the flag-string into a []string, delimited by "true" commas, you can take everything up to the first `:` as key (stripping leading/trailing whitespace) and everything after that (stripping leading/trailing whitespace) as value.  No need to escape `:` at all. 

Regarding whitespace on values:

```
Leading whitespaces after name =, ... and trailing whitespaces of the line are discarded unless they are enclosed in double quotes. Internal whitespaces within the value are retained verbatim.
````

To break the flag-string, though, you need to be smart.

Let's consider a sort of pathological case.  A key with a value that is a list, and another key with a simple value.

key1 = "k1"
value = "a:b, b:c, c:d"
key2 = "k2"
value = "v2"

Naive answer: `k1:a:b, b:c, c:d, k2:v2`

Obviously that doesn't work - it's ambiguous.

Option 1, escape commas: `k1:a:b\, b:c\, c:d, k2:v2`

Option 2 embed quotes: `k1:"a:b, b:c, c:d", k2:v2`

So I THINK we can say something like:

* The value is a series of comma-delimited key-value pairs.
* Each key must be a valid git-config compatible "<section>.<variable>" (see git-config docs)
* Each value must be a either a quoted string or an unquoted string
* Within all values, the following escapes are supported:
  - \n => <newline>
  - \t => <tab>
  - \" => "
  - \, => ,
  - \\ => \
* Within unquoted values, commas MUST be escaped.  Within quoted values, commas MAY be escaped, but are not required to be.
* Any other escape sequence is an error

When we build the string to pass to `git config` we DO NOT reformat it (e.g. no %q).  Git is weird here, and (empirically) it seems correct to pass literal newlines, which git will internally escape.

If there's demand for more escapes we can consider expanding it later (e.g. git-config supports \b, but I don't see why we should).

Now, keep in mind that to get these INTO git-sync, we need to pass them through a shell or YAML or something, so they may get escaped again.  That seems unavoidable.  In truth, I hope the vast vast majority of people just use simple tokens, buy we should be precise and code defensively. :)

We'll need a good test to cover the various ways one could try to break this.

onesolpark: I've gone through the documentation as well.
I'll go and work on the codes for this and a solid test code as well.
btw thanks for the thorough review :)

thockin:
Any updates on this?

thockin:
ping?

ntdef:
I'd really like to see this feature added if possible. I've never written any Go, otherwise I'd give it a crack myself. 

Is there much that needs to be done to finalize this PR?

thockin:
Go is easy.  Getting all the edge-cases in parsing right is the only hard part here :)

thockin:
Hi, I have proposed #341 as a way to pass arbitrary args.  That subsumes this abandoned PR.  Thanks for the idea.

PR #243, opened 2020-04-24
Relax wait between sync

killuazhu:
The default wait time `0` is too aggressive. My team manages a Github Enterprise instance, we had a user who use the default value and cause 200 calls per second against his own repo.

As a default value, no wait is very aggressive.


cmd/git-sync/main.go
```diff
@@ -59,7 +59,7 @@ var flRoot = flag.String("root", envString("GIT_SYNC_ROOT", envString("HOME", ""
 	"the root directory for git-sync operations, under which --dest will be created")
 var flDest = flag.String("dest", envString("GIT_SYNC_DEST", ""),
 	"the name of (a symlink to) a directory in which to check-out files under --root (defaults to the leaf dir of --repo)")
-var flWait = flag.Float64("wait", envFloat("GIT_SYNC_WAIT", 0),
+var flWait = flag.Float64("wait", envFloat("GIT_SYNC_WAIT", 3),
```
thockin, line 62: This is a pretty significant behavioral change.  Changing it to 1 second I could maybe buy, but 3 feels like a lot.

I am planning to cut a new major version eventually (I have some big, breaking changes I want to make) and we could discuss this in that context.  Can you file an issue so we can keep track of it?

Also, there are at least 2 e2e test cases that do not set --wait, so they will probably break after this change ("webhook" and "http").

killuazhu: @thockin I created issue https://github.com/kubernetes/git-sync/issues/244 and reduce the wait time to 1 second. Will look into the test cases.

killuazhu: @thockin after updating the default wait time from 3 to 1 second, I'm able to pass all the tests when running `make test` locally.

thockin:
Thanks!  As trivial as this is, I'd like to sit on it until I fork for a new major release.  I will leave this open and merge once I do that.

Thanks.

thockin:
Thanks!  This will target 3.2.0

/lgtm
/approve

PR #246, opened 2020-04-27
add option to NOT recursively clone submodules

palixthepalalix:
Make recursive submodule cloning optional, but still enabled by default.


palixthepalalix:
I signed it

palixthepalalix:
Yeah, that makes more sense to also add the ability to turn it off entirely.

thockin:
Are you game to update this PR to handle that?

palixthepalalix:
yep i can get to it when i get a free moment

cmd/git-sync/main.go
```diff
@@ -55,6 +55,8 @@ var flRev = flag.String("rev", envString("GIT_SYNC_REV", "HEAD"),
 	"the git revision (tag or hash) to check out")
 var flDepth = flag.Int("depth", envInt("GIT_SYNC_DEPTH", 0),
 	"use a shallow clone with a history truncated to the specified number of commits")
+var flSubmoduleMode = flag.String("submodule-mode", envString("GIT_SYNC_SUBMODULE_MODE", "recursive"),
+	"Submodule clone directive. Options are 'recursive', 'shallow' and 'off'. Default is recursive")
```
thockin, line 59: nits on wording and style.  Something like:

"how to handle submodules - options are 'recursive', 'shallow' and 'off'."

The default should be handled by the flag lib, so no need to reprint here I think

cmd/git-sync/main.go
```diff
@@ -251,6 +259,11 @@ func main() {
 		}
 	}
 
+	if *flSubmoduleMode != SubmoduleModeRecursive && *flSubmoduleMode != SubmoduleModeOff && *flSubmoduleMode != SubmoduleModeShallow {
```
thockin, line 262: Can we make this more readable like:

```
map[string]bool submoduleModes = {
    SubmoduleModeRecursive: true,
    SubmoduleModeShallow: true,
    SubmoduleModeOff: true,
}
 if !submoduleModes[*flSubmoduleMode] {
    error and exit
}
```

thockin: or 
```
switch *flSubmoduleMode {
case SubmoduleModeRecursive, SubmoduleModeShallow, SubmoduleModeOff:
default:
    error, exit
}
```

cmd/git-sync/main.go
```diff
@@ -55,6 +55,8 @@ var flRev = flag.String("rev", envString("GIT_SYNC_REV", "HEAD"),
 	"the git revision (tag or hash) to check out")
 var flDepth = flag.Int("depth", envInt("GIT_SYNC_DEPTH", 0),
 	"use a shallow clone with a history truncated to the specified number of commits")
+var flSubmoduleMode = flag.String("submodule-mode", envString("GIT_SYNC_SUBMODULE_MODE", "recursive"),
```
thockin, line 58: I am not a big fan of "mode" as a term for things like flags.  I don't think it adds much.  Is `--submodules` much worse in your opinion?

palixthepalalix: it's not, yeah "mode" is weird here... i always think working with enum like things is weird with things like command like flags

cmd/git-sync/main.go
palixthepalalix: oh yep switches are fun. when i was doing this a couple months ago i was deep in python and forgot about em.

palixthepalalix:
sorry it took me so long to get back to this

PR #247, opened 2020-05-01
Build updates

thockin:
Move to go 1.14

Explicitly apt-get update when building


thockin:
Self-approving so I can cut a release

stp-ip:
Retrospective lgtm ;)

PR #248, opened 2020-05-03
use uid,gid instead of username and groupname

aneesh-joseph:
Many Pod security policies expect  the container to run with  numeric user , so that it can verify if the container is running with a non-root user(!=0)

Updating the dockerfile to use the numeric equivalent of the git-sync user

### DIFF
--- a/Dockerfile.in
+++ b/Dockerfile.in
@@ -32,5 +32,5 @@ RUN chmod 0666 /etc/passwd
 ADD bin/{ARG_OS}_{ARG_ARCH}/{ARG_BIN} /{ARG_BIN}
 
 WORKDIR /tmp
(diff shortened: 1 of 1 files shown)


aneesh-joseph:
I signed it

aneesh-joseph:
I signed it

stp-ip:
Not sure on the specific reason to go for a username instead of numeric. Waiting for @thockin to jump in.

thockin:
Thanks!

/lgtm
/approve

aneesh-joseph:
thank you @thockin  - are there any upcoming releases which this change will be part of? looking for an image which I can directly pull from the registry(with this change included)

PR #270, opened 2020-09-05
Remove useless test case

thockin:
Semantically this test is identical to other tests.


stp-ip:
To make sure I'm not missing something.
Assumption is that the branch test and the tag test basically test similar things already so a cross branch test, doesn't really add much.

If that's the case I can see the reasoning and lgtm.

thockin:
Tags are tags, regardless of branch (unlike, something like CVS), so yeah
this is just redundant.

stp-ip (approved):

PR #275, opened 2020-09-05
Add metric for askpass, noop sync != success

thockin:

PR #278, opened 2020-09-05
Clean up askpass function to take url as arg

thockin:
Depends on #275


stp-ip (approved):
lgtm, but waiting for #275 to be iterated.

thockin:
Rebased

PR #283, opened 2020-09-09
[Feature] Authenticate using a GitHub App

iwilltry42:
This PR adds basic functionality to authenticate against GitHub using a GitHub app.

## New Flags

- `gh-app-id`: The ID of your GitHub App, used to create a JWT token and to request an installation token
- `gh-app-pem`: Path to the private key file of that GitHub App, used for authentication against the GitHub API when requesting an installation token
- `gh-app-inst-id`: The ID of the installation of that GitHub app you want to use, used to get an access token for it

## New Behaviour

- only works with http/https (same blocking behaviour against ssh as the username/password flags)
- modifies the repo URL directly with a new `x-access-token` prefix, instead of using a credential helper or the like
- sync loop now also takes care of refreshing the token if it's being used and is expired

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /.container-*
 /.dockerfile-*
 /.qemu-initialized
+.idea/
--- a/README.md
+++ b/README.md
@@ -110,5 +110,9 @@ docker run -d \
 | GIT_SYNC_HTTP_BIND              | `--http-bind`              | the bind address (including port) for git-sync's HTTP endpoint                                                                         | ""                            |
 | GIT_SYNC_HTTP_METRICS           | `--http-metrics`           | enable metrics on git-sync's HTTP endpoint                                                                                             | true                          |
 | GIT_SYNC_HTTP_PPROF             | `--http-pprof`             | enable the pprof debug endpoints on git-sync's HTTP endpoint                                                                           | false                         |
+| GIT_SYNC_APP_ID                 | `--gh-app-id`              | the ID of the GitHub App to use for git auth                                                                                           | ""                            |
+| GIT_SYNC_APP_INST_ID            | `--gh-app-inst-id`         | the ID of the GitHub App installation to use for git auth                                                                              | ""                            | 
+| GIT_SYNC_APP_PEM                | `--gh-app-pem`             | the path of the pem file of the GitHub App to use for git auth                                                                         | ""                            | 
+
 
 [![Analytics](https://kubernetes-site.appspot.com/UA-36037335-10/GitHub/git-sync/README.md?pixel)]()
--- a/cmd/git-sync/app.go
+++ b/cmd/git-sync/app.go
@@ -1,0 +1,152 @@
+// Package app provides supporting functionality to authenticate as a GitHub App
+package main
+
+import (
+	"context"
+	"crypto/x509"
+	"encoding/json"
+	"encoding/pem"
+	"fmt"
+	jwt "github.com/dgrijalva/jwt-go"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
+	"time"
+)
+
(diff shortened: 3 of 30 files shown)


thockin:
Hi, I wanted to let you know that I saw this, but I will not have a chance to look at it for a while (it's a big one).

I do feel like I am missing a doc or something to explain why a user might want to do this and how they would use it.

It includes some fairly significant deps, so I wonder if there's a way to reduce that?

iwilltry42:
@thockin , you're definitely right there, I will add some docs 👍
The main reason for me was, that in our organization, using personal access tokens and bot users for interaction is discouraged, since managing access for those is more tricky, especially if the responsible user leaves the company. Additionally, GitHub Apps have higher API rate limits (15000 rpa for enterprise).

I will check if/how I can reduce the dependencies here 👍

iwilltry42:
The only non-stdlib dependency this is adding is `jwt-go`, which I cannot find an easy replacement to unfortunately, but then in terms of compatibility, I guess we're quite fine here, since we're only using it to generate a new token in two function calls.
The PR also "only" adds <70kB to the binary size.
Is it still to problematic due to the dependencies, even if most of them are stdlib? :thinking:

thockin:
It's not a "problem" per se, but there are things I have to consider:

* I read recently an article or something about various JWT packages in Go
and now I need to go find that and understand it
* I don't have bandwidth to deep-review 3rd-party deps, so the mere act of
including it puts EVERY git-sync user at some risk
* Now I have to track that lib for security updates (especially since it is
related to security)
* At then end of this, it's a pretty niche use-case, which makes me
question the return-on-investment

Please don't take that as me being inherently against the idea, but I do
want to explore if there's a better way to handle it.

E.g. some ideas (brainstorm, don't judge :)

* Could we build it into a loadable module that is only loaded if those
flags are set (most users would not use it and not be subject to the risks)
* Could we build it as a "helper" executable that git-sync could exec()
when needed (most users would not use it)
* Could we build it as a helper container, which git-sync could use either
via a volume or via IPC or something (e.g. open a UNIX socket on a shared
volume)

Any other ideas you have would be valuable - I want to make it possible for
you to use this but I'd love to segment it more.  I am not an authn expert,
so it's a bit outside my core domain :)

thockin:
Taking another quick look, tell me if something like this would work:

git-sync takes a --auth-socket flag, which points to a file, ostensibly a UNIX socket.  It writes auth requests there, and reads responses (we can define a trivial API).

Build a new `gh-app-auth` container (or just a binary?) which creates that socket.  It reads requests, does the gh-app auth and responds.

Users who need this (you :) can run this as a sidecar-sidecar.

The real win is that the NEXT request for a new auth mechanism can use the same approach.

What do you think?

stp-ip:
Second the idea to not directly integrate this, but let the authentication token be provided/generated separately and shared through a socket/file etc.

As the github app authentication generates a time limited password like access token wouldn't it be possible to keep updating the password ENV from a sidecar container as a rudimentary alternative and later make --password understand a file, that can be read to access the "updated" token?

Downside is that the sidecar keeps an access token up to date even, if no new requests are planned.

iwilltry42:
@thockin & @stp-ip I totally understand your (security) concerns, especially since the jwt-go module is not that actively maintained (also see recent Twitter activity on this topic).
But maybe, we get jwt support in stdlib: https://twitter.com/FiloSottile/status/1315803614657556481 :)
Closing this for now, as we're using this in internal clusters that don't have external attack surface. Maybe at some point I have some time to set it up as an additional sidecar.
Thanks for your input :+1:

vladComan0:
Is there any new initiative to add functionality for authentication using GitHub apps? Our organisation also disabled service accounts and now uses exclusively Github apps.

thockin:
I don't object to trying to solve this, but I don't have enough context/time to set up a test case and implemement it right now - PRs are welcome, even if it is a sekelton which I could use to build that context.

PR #285, opened 2020-09-23
add an option to run the command when the repository is updated

unknowingknow:
This PR adds an option to run the command when the repository is updated.

`sync-hook-command` specifies the name of the command (included the argument) to be executed when the repository is updated.
This command will be executed with the directory of the updated repository.

This helps when you need to run a specific command after repository has cloned, such as when generating documentation based on code managed by the git repository.


README.md
```diff
@@ -92,6 +92,7 @@ docker run -d \
 | GIT_SYNC_ONE_TIME               | `--one-time`               | exit after the first sync                                                                                                              | false                         |
 | GIT_SYNC_MAX_SYNC_FAILURES      | `--max-sync-failures`      | the number of consecutive failures allowed before aborting (the first sync must succeed, -1 will retry forever after the initial sync) | 0                             |
 | GIT_SYNC_PERMISSIONS            | `--change-permissions`     | the file permissions to apply to the checked-out files (0 will not change permissions at all)                                          | 0                             |
+| GIT_SYNC_HOOK_COMMAND           | `--sync-hook-command`      | the command executed after cloning the new hash of repote repository                                                                   | ""                            |
```
thockin, line 95: typo: "repote" -> "the remote"

s/cloning the/syncing a

cmd/git-sync/main.go
```diff
@@ -70,6 +70,8 @@ var flMaxSyncFailures = flag.Int("max-sync-failures", envInt("GIT_SYNC_MAX_SYNC_
 	"the number of consecutive failures allowed before aborting (the first sync must succeed, -1 will retry forever after the initial sync)")
 var flChmod = flag.Int("change-permissions", envInt("GIT_SYNC_PERMISSIONS", 0),
 	"the file permissions to apply to the checked-out files (0 will not change permissions at all)")
+var flSyncHookCommand = flag.String("sync-hook-command", envString("GIT_SYNC_HOOK_COMMAND", ""),
+	"the command executed after cloning the new hash of repote repository")
```
thockin, line 74: same typos

cmd/git-sync/main.go
```diff
@@ -653,6 +655,23 @@ func addWorktreeAndSwap(ctx context.Context, gitRoot, dest, branch, rev string,
 		}
 	}
 
+	// Execute the command, if requested.
+	if *flSyncHookCommand != "" {
+		log.V(0).Info("executing command for git sync hooks", "command", *flSyncHookCommand)
+		splitData := strings.Fields(*flSyncHookCommand)
```
thockin, line 661: Do we really need to allow args?  This introduces ambiguity around names and quoting.  E.g. if I set flSyncHookCommand to `command "foo bar"` that gets split into `[command, "foo, bar"]` which is pretty clearly not what was wanted.  It might also seem reasonable to try globs, e.g. "command *.md" which is also not going to work. So we either define this as "a string passed to `sh -c`" (and spec which shell) or we spec it as a command name with no args, I think.

thockin: We also need to document that it is subject to the sync timeout and that it will extend the period between syncs.

cmd/git-sync/main.go
thockin: Also document that it runs with the repo as its working directory.

cmd/git-sync/main.go
unknowingknow: Changed to take a command name with no arguments to make the option name intuitive.

unknowingknow:
@thockin 
Could you please check?

thockin (commented):
Huh.  I reviewed this and I know I sent a comment, but it's not here....

Can you write a testcase for this in test_e2e ?  Specifically something like a tiny script that writes to a file to prove that it ran, and then verify the file is correct?

unknowingknow:
@thockin 
Added a testcase. Please recheck.

thockin (commented):
Thanks!

/lgtm
/approve

johnchenghk01:
How to see the stdout and stderr of the command ran?

thockin:
If the command failed, it should be logged - look for `Run(cmd): error
{ stdout: "stdout", stderr: "stderr" }`

If you want logs even when it doesn't fail, set your log level (-v) to
6 or higher.

PR #289, opened 2020-09-25
Fix webhook regression: allow -1 to ignore status

### LINKED ISSUE #288: -webhook-success-status option cannot be disabled in v3.1.7
It seems there are multiple issues with the `-webhook-success-status` option:

1. the input validation requires the value to be > 0
2. the default value is that of the env variable `GIT_SYNC_WEBHOOK_SUCCESS_STATUS` i.e 200
3. the option help still recommends setting this to -1 to disable

I'm not sure if I am missing something here but it looks like this option cannot be disabled in v3.1.7 to make webhook calls fire-and-forget.

thockin:
Fixes #288

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -294,8 +294,8 @@ func main() {
 	}
 
 	if *flWebhookURL != "" {
-		if *flWebhookStatusSuccess <= 0 {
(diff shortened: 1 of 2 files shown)


thockin:
I am going to self approve to start a build

stp-ip:
Retroactive lgtm ;)

PR #291, opened 2020-09-28
logtostderr defaults to true: don't set it in test

thockin:
Cleanup I saw while trying something else.


stp-ip (approved):
/approve
/lgtm

PR #293, opened 2020-10-06
Fix line breaks in README

stp-ip:


thockin (commented):
Thanks!

/lgtm
/approve

Issue #284, opened 2020-09-13
Discussion; Git-sync v4


thockin:
We have accumulated a bunch of things that I think are weird, but can't easily be fixed without breaking people.  Some of how git-sync works (e.g. flags) shows how I didn't (or don't) really understand git well.  

I started a branch to do a better job of pulling (using git mirror) but I realized that was still awkward, especially when people want to economize time spent syncing (e.g. depth 1).

I started a new branch to see if I could find a different way to do it.  I realized we can use git fetch more effectively, which can pretty much guarantee only transferring what is required, but that seems to depend on flags on the git server when used with SHAs (though github seems to work).

So, TL;DR there are a bunch of things I would like to fix, which require breaking changes.  I'm proposing a v3 maintenance branch and a v4 main branch, with breaking changes, such as:

- rename 'dest' flag
- change all time-related flags to Durations
- rename 'wait' flag
- unify branch and rev related flags into a single one
- focus on being as lightweight as possible when pulling

We can discuss merits of each idea, but the proposal stands.

Objections?

thockin:
Could also use this to switch to POSIX flags

stp-ip:
On my radar.

thockin:
I have a patch brewing to switch to pflag and add a `--man` flag to dump
better docs.

stp-ip:
v3 and v4 branches sound good.

Renaming and flag simplification ideas sound good.
Do we want to deprecate the old flags aka warn about them breaking within v4 or do a clean cut?
"--man" I assume to work around pflags default help command or a full man page like doc generation?

Will read up on git fetch in the interim.

thockin:
I'm working on v4 exclusively (when I get time) now.

PR #298, opened 2020-10-31
Add help and manual flags, use pflag

thockin:
THIS IS A BREAKING CHANGE, beginning the v4.x dev cycle.

Switch flags implementation to use pflag.  This means that long flags
like `-username` must now use 2 dashes: `--username`.

The `-v` flag (verbose) used to accept `-v` or `--v`.  Now it only
accepts `-v.

The `--help` and `-h` flags are new.

The `--man` flag is new (print a man-page like help message).

Several glog flags which used to be exposed (e.g. --logtostderr) are no
longer exposed.  Logs always go to stderr.

xref #284


stp-ip (approved):
/lgtm

Is the man page generated or manually curated?

thockin:
The docs are manually managed

Issue #297, opened 2020-10-30
--timeout arg expects type int whereas all other time arguments are type duration


vanugrah:
Hello! 👋 

Firstly, absolutely love this project. Kudos! With that out of the way, I noticed an inconsistency between time format specification across the different command line arguments. Specifically, the following arguments all expect a duration type: 

- `--wait`
- `--webhook-timeout`
- `--webhook-backoff`

The only time argument that defies this convention is `--timeout` which expects an int type instead of a duration. Not sure if this is an intended inconsistency so I thought I'd bring it up.

thockin:
Fixing it is a breaking change, so I plan to do that in a v4 release.
In fact, I have it fixed in my working branch :)

vanugrah:
Splendid! Glad you are aware of this. I just noticed that the `--wait` property is of type float instead of duration as well. Is that also being handled in the v4 release?

thockin:
https://github.com/thockin/git-sync/commit/498e790405b7fac9b9970e29ae1c1f48d2e2f751

I need to get back to this and do it properly.

I'll close this issue now, but thanks for raising it.

PR #299, opened 2020-10-31
Change time-related flags to durations

thockin:
Builds on #298 

Add '--period' to replace '--wait', which is now obsolete.
    
Add '--sync-timeout' to replace '--timeout', which is now obsolete.
    
Both of these new flags take a Go-style time string, rather than a bare
number. For example "1s" for 1 second or "1m" for one minute.
    
The old flags have been kept and will take precedence if specified.

xref #297

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -66,10 +66,10 @@ var flRoot = pflag.String("root", envString("GIT_SYNC_ROOT", envString("HOME", "
 	"the root directory for git-sync operations, under which --dest will be created")
 var flDest = pflag.String("dest", envString("GIT_SYNC_DEST", ""),
 	"the name of (a symlink to) a directory in which to check-out files under --root (defaults to the leaf dir of --repo)")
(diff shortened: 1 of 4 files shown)


cmd/git-sync/main.go
```diff
@@ -1145,11 +1160,13 @@ OPTIONS
             An optional command to be executed after syncing a new hash of the
             remote repository.  This command does not take any arguments and
             executes with the synced repo as its working directory.  The
-            execution is subject to the overall --timeout flag and will extend
-            the period between syncs attempts.
+            execution is subject to the overall --sync-timeout flag and will
+            extend the effective period between syncs attempts.
```
stp-ip, line 1164:
```suggestion
            extend the effective period between sync attempts.
```

stp-ip (changes requested):
small nit

/approve

cmd/git-sync/main.go
thockin: ACK

thockin:
Fixed, rebased and repushed.  PTAL

PR #300, opened 2020-11-07
Change some flags (breaking)

thockin:
* Make the --root flag required, no default (breaking)
* Don't allow --dest to start with a dot (breaking)
* Rename the --dest flag to --link (old flag kept for compat)

See commit messages for more details.

PR #302, opened 2020-11-08
Normalize the root path

thockin:
This makes sure there's never any confusion about whether it is an
absolute path or has symlinks or whatever.  Add e2e cases to cover.

Builds on #300


stp-ip (approved):
/lgtm

PR #303, opened 2020-11-08
Clarify logging vs printf for fatal messages

thockin:
Also log when HTTP serving fails.

Builds on #300


stp-ip (approved):
Not merging due to the conflict, but lgtm otherwise.
/hold
/lgtm

thockin:
Resolved

PR #304, opened 2020-11-08
Add a "main struct" with methods

thockin:
Broken down by commit for easier reviewing.  This uncovered at least one real bug.  This is building to some deeper code re-organization.

Builds on #300, #302, and #303


stp-ip (approved):
Great refactor. Thanks for splitting it up into the separated commits. Fascinating to move through each individually.

/lgtm

PR #307, opened 2020-11-18
Add the `--root` flag into the Usage section

haiyanmeng:


thockin:
Thanks!  Please note that the master branch is currently in a "breaking dev" cycle and might be unstable.  There is a stable release branch if that is what you want.

thockin:
Thanks!

/lgtm
/approve

Issue #320, opened 2021-01-06
Can't pass octals to --change-permissions (GIT_SYNC_PERMISSIONS)


akloeckner:
Hi there,

Thanks for this nice piece of software!

I'm using `git-sync` (`latest` = `v3.2.1`) to clone a repo into a named volume and access it from another container for some post-processing. I'm struggling with the permissions however. (I'm new to containers, so that might just be me.) 

As `git-sync` creates everything as the `65533` user but my other container obviously does not, I thought I'd set the `GIT_SYNC_PERMISSIONS` variable to `777`. However, that actually changes the permissions to something seemingly random.

After some trial and error, I figured out that the octal `777` was apparently converted to a decimal number and passed on to `chmod`. But `chmod` expects an octal number. So, I tried the decimal equivalent of `GIT_SYNC_PERMISSIONS='511'`, which actually worked:
```
$ /git-sync
I0106 00:44:12.602504      12 main.go:430]  "level"=0 "msg"="starting up"  "args"=["/git-sync"] "pid"=12
[...cloning logs...]
I0106 00:44:26.054483      12 main.go:652]  "level"=0 "msg"="changing file permissions"  "mode"="0777"

$ ls -la
total 20
drwxrwxrwt 5 root     root  4096 Jan  6 00:44 .
drwxr-xr-x 1 root     root  4096 Jan  6 00:43 ..
drwxr-xr-x 9 git-sync 65533 4096 Jan  6 00:44 .git
drwx------ 2 git-sync 65533 4096 Jan  6 00:44 .ssh
lrwxrwxrwx 1 git-sync 65533   44 Jan  6 00:44 deploy -> rev-874f48bce4069af461bc2f23e785d7ffa722ea96
drwxrwxrwx 7 git-sync 65533 4096 Jan  6 00:44 rev-874f48bce4069af461bc2f23e785d7ffa722ea96

$ set
GIT_KNOWN_HOSTS='false'
GIT_SSH_KEY_FILE='/etc/git-secret/id_rsa_docker'
GIT_SYNC_BRANCH='deploy'
GIT_SYNC_DEST='deploy'
GIT_SYNC_PERMISSIONS='511'
GIT_SYNC_REPO='[obscured]'
GIT_SYNC_ROOT='/tmp'
GIT_SYNC_SSH='true'
GIT_SYNC_TIMEOUT='120'
GIT_SYNC_WAIT='30'
HOME='/tmp'
HOSTNAME='c494b3dcc487'
[...]

PR #322, opened 2021-01-06
Allow octal and hex values for int flags

thockin:
This is specifcally relevant to --change-permissions.

xref #320


stp-ip (approved):
/lgtm

PR #328, opened 2021-01-21
Add an example pod YAML

thockin:


stp-ip (approved):
/lgtm

Note: v4 is not available yet and therefore makes this example non workable for now.

PR #329, opened 2021-01-21
Default --root flag in container

### LINKED ISSUE #325: Permission denied on /tmp/git
When using git-sync in the recommended way, by bind-mounting the destination folder onto `/tmp/git`, I get the following error:

```
E0116 23:29:01.393979      11 main.go:455]  "msg"="too many failures, aborting" "error"="Run(git clone --no-checkout -b master ssh://[...].git /tmp/git): exit status 1: { stdout: \"\", stderr: \"Cloning into '/tmp/git'...\\n/tmp/git/.git: Permission denied\\n\" }"  "failCount"=0
```

This is the case for an empty bind-mount and also if I use a volume.

I have read https://github.com/kubernetes/git-sync/issues/245#issuecomment-642379627 and the workaround there also works for me: Mount the volume onto `/tmp` directly. 

So, I wondered why this is...

Loading the fresh image with only a `sleep` command lets me inspect the initial contents of `tmp`:
```
$ ls -la
total 8
drwxrwxrwt 1 root root 4096 Jan  5 17:18 .
drwxr-xr-x 1 root root 4096 Jan 16 23:48 ..
```

As we can see, `/tmp` is world-writeable, so the `git-sync` user can create his own `git` directory here and work with it. This is why mounting a new volume onto `/tmp` works as expected:

```
$ ls -la /tmp
total 16
drwxrwxrwt 4 root     root  4096 Jan 16 23:45 .
drwxr-xr-x 1 root     root  4096 Jan 16 23:45 ..
drwx------ 2 git-sync 65533 4096 Jan 16 23:45 .ssh
drwxr-xr-x 4 git-sync 65533 4096 Jan 16 23:45 git
```

**CAVEAT**: The checkedout folder structure will be `VOLUME/git/SYMLINK` in this case. If I want to get rid of the `git` folder inbetween, e.g. for mounting the volume into an `nginx`container, I end up with a `SYMLINK` inside `/tmp`. Since `/tmp` has the `t` flag set, I can not follow this link as `root` user, such as when I'm in a standard `nginx` container...

If I mount an empty volume onto `/tmp/git`, it will not be writeable by `git-sync`:
```
$ ls -la /tmp
total 12
drwxrwxrwt 1 root root 4096 Jan 16 23:53 .
drwxr-xr-x 1 root root 4096 Jan 16 23:53 ..
drwxr-xr-x 2 root root 4096 Jan 16 23:52 git
```

Apparently, the `/tmp/git` folder is created as root by docker itself, when mounting the volume.

**Suggested improvement:**

This is why I would suggest to create the `git` folder from the docs already in the image and make it be owned by `git-sync`. This would also address the caveat mentioned above, because the new `git` folder would ideally not have the `t` flag enabled...

**Note**: This will not fix the cause you lined out in #245. If we bind-mount a non-existing folder from the host system (even onto `/tmp` as in the workaround), the permission will still be denied by the host system. But as you said, there is nothing we can do about that.

What do you think?

thockin:
This makes the container easier to use and allows docker volumes to be
used in a more obvious way.

Fixes #325

### DIFF
--- a/Dockerfile.in
+++ b/Dockerfile.in
@@ -12,6 +12,38 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# HOW TO USE THIS CONTAINER:
+#
+# For most users, the simplest way to use this container is to mount a volume
+# on /tmp/git.  The only commandline argument (or env var) that is really
+# required is `--repo` ($GIT_SYNC_REPO).  Everything else is optional (run this
+# with `--man` for details).
+#
+# This container will run as UID:GID 65533:65533 by default, and unless you
+# change that, you do not need to think about permissions much.  If you run
+# into permissions problems, this might help:
+#
+#  - User does not mount a volume
+#    => should work, but limited utility
+#
+#  - User mounts a new docker volume on /tmp/git
+#    => should work
+#
+#  - User mounts an existing docker volume on /tmp/git
+#    => if the volume already exists with compatible permissions it should work
+#    => if the volume already exists with different permissions you can either
+#       set the container UID or GID(s) or you can chown the volume
+#
+#  - User mounts an existing dir on /tmp/git
+#    => set container UID or GID(s) to be able to access that dir
+#
+#  - User sets a different UID and git-sync GID
+#    => should work
+#
+#  - User sets a different GID
+#    => either add the git-sync GID or else set --root, mount a volume,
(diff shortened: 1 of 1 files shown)


stp-ip (approved):
Went through the referenced issue and see the reasoning you went through no objections aka

/lgtm

PR #330, opened 2021-01-21
Default the git-sync root dir in container

thockin:
This should make the container image easier to use, without defaulting the root dir in non-container usage.

xref #325


stp-ip (approved):
/lgtm

Issue #326, opened 2021-01-20
Proposal: Generate an error file


nan-yu:
# Objective
The goal of this feature is to export the git-sync errors to a file which can be shared by other sidecar containers.

# Motivation
The current git-sync process outputs the error information to standard out, which is inaccessible from outside the container. Users have to dump the logs using `kubectl logs` in order to check the error details in the git-sync process. This proposal provides users the capability to check the errors directly from other sidecar containers.

# Goals
- Share the most recent git-sync error with other sidecar containers.

# Non-Goals
- This proposal does not intend to redirect all logging information to a file.
- The error file only keeps the most recent error. It is not supposed to be used as the error history.

# High-level Design
There will be a new flag to indicate whether to generate the error file. If the flag is turned on, the error file may or may not exist depending on the sync status.
If sync succeeds, the file will be removed. Otherwise, the file will be created or overwritten with the most recent error message.
To guarantee atomic update to the error file, we always write to a temp file first and then rename that to the error file. This enables the client to know the file has been updated even if the contents are the same.
The client will poll on the error file to get the error message on failures.

# Implementation details
## Add a flag for `--error-file`
```
var flErrorFile = flag.String("error-file", envString("GIT_SYNC_ERROR_FILE", ""), "the path to the error file where to dump the most recent error details")
```

### 1. Before calling `os.Exit()` with a non-zero code
There are some common steps if the process exits abnormally with a non-zero code: output the error to standard error, print the usage information (optional), export the error to the error file and exit the process. Therefore, we extract these steps into a function.
```
// handleError prints the error to the standard error, prints the usage if the `printUsage` flag is true,
// exports the error to the error file and exits the process with the exit code.
func handleError(exitCode int, printUsage bool, format string, a ...interface{}) {
	fmt.Fprintf(os.Stderr, format, a...)
	if printUsage {
		flag.Usage()
	}
	exportError(fmt.Sprintf(format, a...))
	os.Exit(exitCode)
}
```

Below are the cases when the process exits with a non-zero code:
- [Flag initialization errors](https://github.com/kubernetes/git-sync/blob/release-3.x/cmd/git-sync/main.go#L253)
  If any flag is invalid, export the error before exiting by calling `handleError(1, true|false, "ERROR: can't ....: %v\n", err)`

[...]

PR #331, opened 2021-01-22
Export the error details to an error file

nan-yu:
The current git-sync process outputs the error information to standard
out, which is inaccessible from outside the container. Users have to
dump the logs using kubectl logs in order to check the error details in
the git-sync process. This commit exports the error details to a file,
which provides users the capability to check the errors directly from
other sidecar containers.

proposal: https://github.com/kubernetes/git-sync/issues/326

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -21,6 +21,7 @@ package main // import "k8s.io/git-sync/cmd/git-sync"
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"flag"
 	"fmt"
 	"io/ioutil"
@@ -37,6 +38,7 @@ import (
 	"time"
 
 	"github.com/go-logr/glogr"
+	"github.com/go-logr/logr"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"k8s.io/git-sync/pkg/pid1"
@@ -60,6 +62,8 @@ var flRoot = flag.String("root", envString("GIT_SYNC_ROOT", envString("HOME", ""
 	"the root directory for git-sync operations, under which --dest will be created")
 var flDest = flag.String("dest", envString("GIT_SYNC_DEST", ""),
 	"the name of (a symlink to) a directory in which to check-out files under --root (defaults to the leaf dir of --repo)")
+var flErrorFile = flag.String("error-file", envString("GIT_SYNC_ERROR_FILE", ""),
+	"the name of a file into which errors will be written under --root (defaults to \"\", disabling error reporting)")
 var flWait = flag.Float64("wait", envFloat("GIT_SYNC_WAIT", 1),
 	"the number of seconds between syncs")
 var flSyncTimeout = flag.Int("timeout", envInt("GIT_SYNC_TIMEOUT", 120),
@@ -119,7 +123,7 @@ var flHTTPMetrics = flag.Bool("http-metrics", envBool("GIT_SYNC_HTTP_METRICS", t
 var flHTTPprof = flag.Bool("http-pprof", envBool("GIT_SYNC_HTTP_PPROF", false),
 	"enable the pprof debug endpoints on git-sync's HTTP endpoint")
 
-var log = glogr.New()
+var log *customLogger
 
 // Total pull/error, summary on pull duration
 var (
@@ -155,6 +159,90 @@ const (
 	submodulesOff       = "off"
 )
 
+type customLogger struct {
+	logr.Logger
+	errorFile string
+}
+
+func (l customLogger) Error(err error, msg string, kvList ...interface{}) {
+	l.Logger.Error(err, msg, kvList...)
+	if l.errorFile == "" {
+		return
+	}
+	payload := struct {
+		Msg  string
+		Err  string
+		Args map[string]interface{}
+	}{
+		Msg:  msg,
+		Err:  err.Error(),
+		Args: map[string]interface{}{},
+	}
+	if len(kvList)%2 != 0 {
+		kvList = append(kvList, "<no-value>")
+	}
+	for i := 0; i < len(kvList); i += 2 {
+		k, ok := kvList[i].(string)
+		if !ok {
+			k = fmt.Sprintf("%v", kvList[i])
+		}
+		payload.Args[k] = kvList[i+1]
+	}
+	jb, err := json.Marshal(payload)
+	if err != nil {
+		l.Logger.Error(err, "can't encode error payload")
+		content := fmt.Sprintf("%v", err)
+		l.writeContent([]byte(content))
+	} else {
+		l.writeContent(jb)
+	}
+}
+
+// exportError exports the error to the error file if --export-error is enabled.
+func (l *customLogger) exportError(content string) {
+	if l.errorFile == "" {
+		return
+	}
+	l.writeContent([]byte(content))
+}
+
+// writeContent writes the error content to the error file.
+func (l *customLogger) writeContent(content []byte) {
+	tmpFile, err := ioutil.TempFile(*flRoot, "tmp-err-")
+	if err != nil {
+		l.Logger.Error(err, "can't create temporary error-file", "directory", *flRoot, "prefix", "tmp-err-")
+		return
+	}
+	defer func() {
+		if err := tmpFile.Close(); err != nil {
+			l.Logger.Error(err, "can't close temporary error-file", "filename", tmpFile.Name())
+		}
+	}()
+
+	if _, err = tmpFile.Write(content); err != nil {
+		l.Logger.Error(err, "can't write to temporary error-file", "filename", tmpFile.Name())
+		return
+	}
+
+	if err := os.Rename(tmpFile.Name(), l.errorFile); err != nil {
+		l.Logger.Error(err, "can't rename to error-file", "temp-file", tmpFile.Name(), "error-file", l.errorFile)
+		return
+	}
+}
+
+// deleteErrorFile deletes the error file.
+func (l *customLogger) deleteErrorFile() {
+	if l.errorFile == "" {
+		return
+	}
+	if err := os.Remove(l.errorFile); err != nil {
+		if os.IsNotExist(err) {
+			return
+		}
+		l.Logger.Error(err, "can't delete the error-file", "filename", l.errorFile)
+	}
+}
+
 func init() {
 	prometheus.MustRegister(syncDuration)
 	prometheus.MustRegister(syncCount)
@@ -184,7 +272,7 @@ func envInt(key string, def int) int {
 	if env := os.Getenv(key); env != "" {
 		val, err := strconv.ParseInt(env, 0, 0)
 		if err != nil {
-			log.Error(err, "invalid env value, using default", "key", key, "val", os.Getenv(key), "default", def)
+			fmt.Fprintf(os.Stderr, "WARNING: invalid env value (%v): using default, key=%s, val=%q, default=%d\n", err, key, env, def)
 			return def
 		}
 		return int(val)
@@ -196,7 +284,7 @@ func envFloat(key string, def float64) float64 {
 	if env := os.Getenv(key); env != "" {
 		val, err := strconv.ParseFloat(env, 64)
 		if err != nil {
-			log.Error(err, "invalid env value, using default", "key", key, "val", os.Getenv(key), "default", def)
+			fmt.Fprintf(os.Stderr, "WARNING: invalid env value (%v): using default, key=%s, val=%q, default=%f\n", err, key, env, def)
 			return def
 		}
 		return val
@@ -208,7 +296,7 @@ func envDuration(key string, def time.Duration) time.Duration {
 	if env := os.Getenv(key); env != "" {
 		val, err := time.ParseDuration(env)
 		if err != nil {
-			log.Error(err, "invalid env value, using default", "key", key, "val", os.Getenv(key), "default", def)
+			fmt.Fprintf(os.Stderr, "WARNING: invalid env value (%v): using default, key=%s, val=%q, default=%d\n", err, key, env, def)
 			return def
 		}
 		return val
@@ -220,8 +308,7 @@ func setFlagDefaults() {
 	// Force logging to stderr (from glog).
 	stderrFlag := flag.Lookup("logtostderr")
 	if stderrFlag == nil {
-		fmt.Fprintf(os.Stderr, "ERROR: can't find flag 'logtostderr'\n")
-		os.Exit(1)
+		handleError(false, "ERROR: can't find flag 'logtostderr'")
 	}
 	stderrFlag.Value.Set("true")
 }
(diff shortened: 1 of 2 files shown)


nan-yu:
FYI @janetkuo @haiyanmeng @ekitson

cmd/git-sync/main.go
```diff
@@ -60,6 +61,8 @@ var flRoot = flag.String("root", envString("GIT_SYNC_ROOT", envString("HOME", ""
 	"the root directory for git-sync operations, under which --dest will be created")
 var flDest = flag.String("dest", envString("GIT_SYNC_DEST", ""),
 	"the name of (a symlink to) a directory in which to check-out files under --root (defaults to the leaf dir of --repo)")
+var flErrorFile = flag.String("error-file", envString("GIT_SYNC_ERROR_FILE", ""),
+	"the path to the error file where to dump the most recent error details")
```
janetkuo, line 65: When the error file path is not provided, would it throw an error for not being able to create an error file? If so, could we make this optional? i.e. if it's not specified, skip creating the error file

nan-yu: Thanks. Updated `exportError` function to check if the file path is provided.

cmd/git-sync/main.go
```diff
@@ -460,11 +430,13 @@ func main() {
 			if *flMaxSyncFailures != -1 && failCount >= *flMaxSyncFailures {
 				// Exit after too many retries, maybe the error is not recoverable.
 				log.Error(err, "too many failures, aborting", "failCount", failCount)
+				exportError(err.Error())
```
thockin, line 433: In these cases - is the Error() enough context?  This could come as "File does not exist", which is not super helpful.

I think this is worth a bit more scaffolding.  What do you think of something like:

```
type customLogger struct {
    logr.Logger
    errorFile string
}

func (l customLogger) Error(err error, msg string, kvList ...interface{}) {
    l.Logger.Error(err, msg, kvList...)
    if l.errorFile != "" {
        l.writeErrorFile(err, msg, kvList...)
    }
}
```

and instead of `var log = glogr.New()` at global scope:

```
var log customLogger
```

and in main():

```
log = customLogger{ glogr.New(), *flErrorFile }
```

Then you can totally encapsulate *flErrorFile in `main()` and `writeErrorFile()` and `deleteErrorFile()` can be methods on `customLogger`.

Within `writeErrorFile()` you can unpack the kvList into something like:

```
    payload := struct {
        Msg string
        Err string
        Args map[string]interface{}
    } {
        Msg: msg,
        Err: err.Error(),
        Args: map[string]interface{}{}
    }
    for i := 0; i < len(kvList); i++ {
        // unpack into Args
    }
  }
```

Then when when you write to the file:

```
jb, err := json.Marshall(payload)
```

So we write something with a defined structure to the file.

thockin (commented):
We'll also need a test_e2e.sh block that proves that errors are written (you can pass bad flags to force it) and that the file is cleaned up when the flags are good.

cmd/git-sync/main.go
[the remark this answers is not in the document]
nan-yu: I was thinking to use it as a relative path under --root. After a second thought, we can change the flag to be a boolean flag and use a fixed path.
Users only need to turn the flag on and off. We'll specify the path in the description.

cmd/git-sync/main.go
```diff
@@ -220,56 +223,50 @@ func setFlagDefaults() {
...
 	if stderrFlag == nil {
-		fmt.Fprintf(os.Stderr, "ERROR: can't find flag 'logtostderr'\n")
-		os.Exit(1)
+		handleError(1, false, "ERROR: can't find flag 'logtostderr'\n")
 	}
 	stderrFlag.Value.Set("true")
 }
 
 func main() {
+	if *flVer {
```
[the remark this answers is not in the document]
nan-yu: I'm inclined to run the pid1 logic first for the following reasons:
- It is unlikely that pid1 should fail.
- If pid1 does error and we restructure as above, there is still chance that the pid1 error will be deleted if the child process is not terminated and successfully synced.
- The error file only reflects the sync errors.

cmd/git-sync/main.go
```diff
@@ -517,6 +493,81 @@ func sleepForever() {
 	os.Exit(0)
 }
 
+// validateErrorPath validates if the parent directory of `--error-file` exits and creates one if not exits.
+// It also checks if the parent directory has the read and write permission.
+func validateErrorPath(errPath string) {
+	errDir := filepath.Dir(errPath)
+	if errDirInfo, err := os.Stat(errDir); err != nil {
+		fmt.Printf("The error parent path %s doesn't exist. Attempt to create it\n", errDir)
```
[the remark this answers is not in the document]
nan-yu: After changing the `--export-error` flag to boolean, there's no need to validate the error path.

cmd/git-sync/main.go
```diff
@@ -517,6 +493,81 @@ func sleepForever() {
...
+		os.Exit(1)
+	} else if err = syscall.Access(errDir, syscall.O_RDWR); err != nil {
+		fmt.Fprintf(os.Stderr, "ERROR: can't access %s: %v\n", errDir, err)
+		os.Exit(1)
+	}
+}
+
+// exportError writes the error content to the error file.
+func exportError(content string) {
+	if *flErrorFile != "" {
```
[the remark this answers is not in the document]
nan-yu: DONE

cmd/git-sync/main.go
```diff
@@ -517,6 +493,81 @@ func sleepForever() {
...
+		tmpFile, err := ioutil.TempFile(os.TempDir(), "err-")
```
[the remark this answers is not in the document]
nan-yu: DONE. Both the actual error file and the temp file will be under the same --root dir.

cmd/git-sync/main.go
```diff
@@ -517,6 +493,81 @@ func sleepForever() {
...
+		if err != nil {
+			fmt.Fprintf(os.Stderr,"Cannot create temporary file: %v\n", err)
+			os.Exit(1)
+		}
+
+		if _, err = tmpFile.WriteString(content); err != nil {
+			fmt.Fprintf(os.Stderr, "ERROR: can't write to temporary file: %v\n", err)
+			os.Exit(1)
+		}
+		defer func() {
```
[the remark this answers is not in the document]
nan-yu: Thanks! Done

cmd/git-sync/main.go
```diff
@@ -517,6 +493,81 @@ func sleepForever() {
...
+			fmt.Fprintf(os.Stderr, "ERROR: can't rename to error file: %v\n", err)
+			os.Exit(1)
+		}
+	}
+}
+
+// deleteErrorFile deletes the error file.
+func deleteErrorFile() {
+	if *flErrorFile != "" {
+		if _, err := os.Stat(*flErrorFile); err != nil {
```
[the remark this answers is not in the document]
nan-yu: DONE

cmd/git-sync/main.go
nan-yu: Thanks for the suggestion. I like the defined structure in the error file where we can have additional information.
DONE

cmd/git-sync/main.go
```diff
@@ -517,6 +493,81 @@ func sleepForever() {
...
+		if err := os.Rename(tmpFile.Name(), *flErrorFile); err != nil {
+			fmt.Fprintf(os.Stderr, "ERROR: can't rename to error file: %v\n", err)
+			os.Exit(1)
+		}
+	}
+}
+
+// deleteErrorFile deletes the error file.
+func deleteErrorFile() {
+	if *flErrorFile != "" {
```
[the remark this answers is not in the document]
nan-yu: DOEN

cmd/git-sync/main.go
```diff
@@ -517,6 +493,81 @@ func sleepForever() {
...
+		os.Exit(1)
+	}
+}
+
+// exportError writes the error content to the error file.
+func exportError(content string) {
+	if *flErrorFile != "" {
+		tmpFile, err := ioutil.TempFile(os.TempDir(), "err-")
+		if err != nil {
+			fmt.Fprintf(os.Stderr,"Cannot create temporary file: %v\n", err)
```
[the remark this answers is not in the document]
nan-yu: DONE

cmd/git-sync/main.go
```diff
@@ -155,6 +159,93 @@ const (
...
+func (l *customLogger) exportError(content string) {
+	if !*flExportError {
+		return
+	}
+	l.writeContent([]byte(content))
+}
+
+// writeContent writes the error content to the error file.
+func (l *customLogger) writeContent(content []byte) {
+	tmpFile, err := ioutil.TempFile(*flRoot, "tmp-err-")
```
thockin, line 216: don't read global flags - stick the root into the logger instance at creation.

cmd/git-sync/main.go
```diff
@@ -155,6 +159,93 @@ const (
...
+	if err != nil {
+		fmt.Fprintf(os.Stderr,"ERROR: can't create temporary file: %v\n", err)
+		os.Exit(1)
```
thockin, line 219: no exit allowed.  Return the error and log it through the underlying logger

cmd/git-sync/main.go
```diff
@@ -155,6 +159,93 @@ const (
...
+	}
+
+	if _, err = tmpFile.Write(content); err != nil {
+		fmt.Fprintf(os.Stderr, "ERROR: can't write to temporary file: %v\n", err)
+		os.Exit(1)
+	}
+	if err := tmpFile.Close(); err != nil {
```
thockin, line 226: stick this in a `defer` right after it was opened (before trying to write), or you will leak FDs.

cmd/git-sync/main.go
```diff
@@ -184,7 +275,7 @@ func envInt(key string, def int) int {
 	if env := os.Getenv(key); env != "" {
 		val, err := strconv.ParseInt(env, 0, 0)
 		if err != nil {
-			log.Error(err, "invalid env value, using default", "key", key, "val", os.Getenv(key), "default", def)
+			fmt.Fprintf(os.Stderr, "invalid env value: %v, using default, key=%q, val=%q, default=%d\n", err, key, os.Getenv(key), def)
```
thockin, line 278: This is a good catch - it would log before flag parsing (bad).

How about:

```
fmt.Fprintf(os.Stderr, "WARNING: invalid env value (%v): using default, key=%s, val=%q, default=%d\n", err, key, env, def)
```

thockin (commented):
Thanks for the test!  Mostly small nits to pick. Looks pretty good!

cmd/git-sync/main.go
```diff
@@ -60,6 +62,8 @@ var flRoot = flag.String("root", envString("GIT_SYNC_ROOT", envString("HOME", ""
 	"the root directory for git-sync operations, under which --dest will be created")
 var flDest = flag.String("dest", envString("GIT_SYNC_DEST", ""),
 	"the name of (a symlink to) a directory in which to check-out files under --root (defaults to the leaf dir of --repo)")
+var flExportError = flag.Bool("export-error", envBool("GIT_EXPORT_ERROR", false),
```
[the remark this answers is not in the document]
nan-yu: Agreed, especially both dest and error are under --root.
DONE

cmd/git-sync/main.go
```diff
@@ -60,6 +62,8 @@ var flRoot = flag.String("root", envString("GIT_SYNC_ROOT", envString("HOME", ""
...
+	fmt.Sprintf("export the error details. If enabled, the most recent error details will be dumped to %q under --root", errorFile))
```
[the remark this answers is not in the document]
nan-yu: DONE

cmd/git-sync/main.go
```diff
@@ -155,6 +159,93 @@ const (
...
+)
+
+type customLogger struct {
+	logr.Logger
+	errorFile string
+}
+
+func (l customLogger) Error(err error, msg string, kvList ...interface{}) {
+	l.Logger.Error(err, msg, kvList...)
+	if !*flExportError {
```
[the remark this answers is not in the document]
nan-yu: DONE

cmd/git-sync/main.go
```diff
@@ -155,6 +159,93 @@ const (
...
+		Msg string
+		Err string
+		Args map[string]interface{}
+	} {
+		Msg: msg,
+		Err: err.Error(),
+		Args: map[string]interface{}{},
+	}
+	for i := 0; i < len(kvList); i += 2 {
+		k, ok := kvList[i].(string)
```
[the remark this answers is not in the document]
nan-yu: DONE

cmd/git-sync/main.go
```diff
@@ -155,6 +159,93 @@ const (
...
+		if !ok {
```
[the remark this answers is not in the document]
nan-yu: SG. DONE

cmd/git-sync/main.go
```diff
@@ -155,6 +159,93 @@ const (
...
+		var v interface{}
+		if i+1 < len(kvList) {
+			v = kvList[i+1]
+		}
+		payload.Args[k] = v
+	}
+	jb, err := json.Marshal(payload)
+	if err != nil {
+		fmt.Fprintf(os.Stderr,"ERROR: can't encode payload: %v\n", err)
+		os.Exit(1)
```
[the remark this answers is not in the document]
nan-yu: DONE

cmd/git-sync/main.go
nan-yu: DONE

cmd/git-sync/main.go
nan-yu: DONE

nan-yu:
Thanks for reviewing the PR! All comments have been addressed. Please take another look when you get a chance.

cmd/git-sync/main.go
```diff
@@ -60,6 +62,8 @@ var flRoot = flag.String("root", envString("GIT_SYNC_ROOT", envString("HOME", ""
 	"the root directory for git-sync operations, under which --dest will be created")
 var flDest = flag.String("dest", envString("GIT_SYNC_DEST", ""),
 	"the name of (a symlink to) a directory in which to check-out files under --root (defaults to the leaf dir of --repo)")
+var flError = flag.String("error", envString("GIT_SYNC_ERROR", ""),
```
thockin, line 65: how about flErrorFile and --error-file and GIT_SYNC_ERROR_FILE

cmd/git-sync/main.go
```diff
@@ -155,6 +159,91 @@ const (
...
+		return
+	}
+	l.writeContent([]byte(content))
+}
+
+// writeContent writes the error content to the error file.
+func (l *customLogger) writeContent(content []byte) {
+	tmpFile, err := ioutil.TempFile(*flRoot, "tmp-err-")
+	if err != nil {
+		l.Logger.Error(err, "can't create temporary file")
```
thockin, line 213: "...temporary error-file", "filename", tmpFile.Name()

Same in other errors

cmd/git-sync/main.go
```diff
@@ -155,6 +159,91 @@ const (
...
+		return
+	}
+	defer func() {
+		if err := tmpFile.Close(); err != nil {
+			l.Logger.Error(err, "can't close temporary file")
+		}
+	}()
+
+	if _, err = tmpFile.Write(content); err != nil {
+		l.Logger.Error(err, "can't create write to temporary file")
```
thockin, line 223: s/create write/write

cmd/git-sync/main.go
```diff
@@ -155,6 +159,91 @@ const (
...
+		return
+	}
+
+	if err := os.Rename(tmpFile.Name(), l.errorFile); err != nil {
+		l.Logger.Error(err, "can't rename to error file")
```
thockin, line 228: include old and new names

cmd/git-sync/main.go
```diff
@@ -155,6 +159,91 @@ const (
...
+func (l *customLogger) deleteErrorFile() {
+	if l.errorFile == "" {
+		return
+	}
+	if err := os.Remove(l.errorFile); err != nil {
+		if os.IsNotExist(err) {
+			return
+		}
+		fmt.Fprintf(os.Stderr, "ERROR: can't delete the error file: %v\n", err)
+		os.Exit(1)
```
thockin, line 243: Don't exit, please :)

log it to the underlying logger and move on

thockin (commented):
super nits.  So sorry.  if not for the exit I would have done this in followup.

cmd/git-sync/main.go
nan-yu: Sounds good. Originally, I thought --error is consistent with --root and --dest, but both of --root and --dest are directories and --error seems a little bit vague.
DONE

cmd/git-sync/main.go
nan-yu: Sorry, I missed that! Done

nan-yu:
> super nits. So sorry. if not for the exit I would have done this in followup.

no worries. I appreciate your time and I like the way how it looks right now.

thockin (commented):
Now, do you want to port it to the master (v4) branch, or do you want me to?

/lgtm
/approve

nan-yu:
> Now, do you want to port it to the master (v4) branch, or do you want me to?

I can port it to the master branch and will send another PR later.

A quick question for the release. When do we plan to make a new v3 release?

thockin:
I was waiting for this to merge.  I MIGHT be able to kick that off today,
otherwise 2nd half next week ?

nan-yu:
PR for the master branch: https://github.com/kubernetes/git-sync/pull/375

PR #332, opened 2021-01-23
Use docker to run helper servers in e2e

thockin:
Reliably cleaning up leftover things like `nc` processes is surprisingly
difficult in pure shell.  e2e passes on the master branch now.


stp-ip (approved):
/lgtm

PR #341, opened 2021-03-11
Add --git-config flag

### LINKED ISSUE #235: Support using custom ca(self-signed) for interacting with git

### LINKED ISSUE #340: RPC failed; curl 56 GnuTLS recv error
stdout: \"\", stderr: \"Cloning into '/tmp/git'...\\nerror: RPC failed; curl 56 GnuTLS recv error (-54): Error in the pull function.\\nfatal: the remote end hung up unexpectedly\\nfatal: early EOF\\nfatal: index-pack failed\\n\

According to https://stackoverflow.com/questions/38378914/how-to-fix-git-error-rpc-failed-curl-56-gnutls
i would need to git config --global http.postBuffer 1048576000 or
"Rebuilding git with openssl instead of gnutls"

thockin:
This allows arbitrary git configs to be passed in.  For example:

`git config --global http.postBuffer 1048576000`

This flag takes a comma-separated list of `key:val` pairs.  The key part
is passed to `git config` and must be a valid gitconfig section header
and variable name.  The val part can be either a quoted or unquoted
value.  For all values the following escape sequences are supported:
   * `\n` => [newline]
   * `\t` => [tab]
   * `\"` => `"`
   * `\,` => `,`
   * `\\` => `\`

Within unquoted values, commas MUST be escaped. Within quoted values,
commas MAY be escaped, but are not required to be.  Any other escape
sequence is an error.

Example:

`--git-config=foo.one:val1,foo.two:"quoted val",foo.three:12345`

This commit exposed a bug in runCommand() which modified its args when
they had an embedded space.

Fixes #235 
Fixes #340 
Replaces #240 (thanks to @onesolpark for starting it)


stp-ip (approved):
/lgtm

PR #342, opened 2021-03-11
Add --git-config flag

thockin:
This allows arbitrary git configs to be passed in.  For example:

`git config --global http.postBuffer 1048576000`

This flag takes a comma-separated list of `key:val` pairs.  The key part
is passed to `git config` and must be a valid gitconfig section header
and variable name.  The val part can be either a quoted or unquoted
value.  For all values the following escape sequences are supported:
   * `\n` => [newline]
   * `\t` => [tab]
   * `\"` => `"`
   * `\,` => `,`
   * `\\` => `\`

Within unquoted values, commas MUST be escaped. Within quoted values,
commas MAY be escaped, but are not required to be.  Any other escape
sequence is an error.

Example:

`--git-config=foo.one:val1,foo.two:"quoted val",foo.three:12345`

This commit exposed a bug in runCommand() which modified its args when
they had an embedded space.

Replaces #240 (thanks to @onesolpark for starting it)


stp-ip:
Ran into an issue. Investigating.

thockin:
I think tests on master (v4) are unhappy.  I can tackle that independently.

/hold

stp-ip:
Yeah. I missed that in the review #351 as far as I can see. It's due to the file/dir being ambigious to the revision as we removed the "rev-".
Will check it out and overall shouldn't be a blocker here once I boiled down the issue I guess.

thockin:
I have a trivial fix - coming momentarily.

thockin:
https://github.com/kubernetes/git-sync/pull/353

PR #344, opened 2021-03-11
Fix exit non-zero exit codes when running as pid1

thockin:
(v4 branch)

Prior to this we would swallow the exit code and always exit(0).

### DIFF
--- a/pkg/pid1/pid1.go
+++ b/pkg/pid1/pid1.go
@@ -11,54 +11,67 @@ import (
 // ReRun converts the current process into a bare-bones init, runs the current
 // commandline as a child process, and waits for it to complete.  The new child
 // process shares stdin/stdout/stderr with the parent.  When the child exits,
-// this will return the same value as exec.Command.Run(). If there is an error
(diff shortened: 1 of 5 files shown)


pkg/pid1/pid1.go
```diff
@@ -11,60 +11,73 @@ import (
...
 		}
 	}
+	return 0, fmt.Errorf("signal handler terminated unexpectedly")
 }
 
-// sigchld handles a SIGCHLD.  This will return true when firstborn exits.  In
-// case of truly unknown errors it will panic.
-func sigchld(firstborn int) bool {
+// sigchld handles a SIGCHLD.  This will return true only when firstborn exits.
+// For any other process thihs will return false and a 0 status.
```
stp-ip, line 63:
```suggestion
// For any other process this will return false and a 0 status.
```

stp-ip (approved):
/approve

Small nit/typo here as well

thockin:
Typo fixed

PR #346, opened 2021-03-12
Update git from backports

thockin:
Some bugs have been fixed that impact some users.

Fixes #334

PR #350, opened 2021-03-12
Change the symlink targets to just the SHA

### LINKED ISSUE #345: Define readlink() as API to know current rev

thockin:
This allows users to call readlink() on the link and learn the current
checked out SHA.

Fixes #345 

Depends on #346

### DIFF
--- a/Dockerfile.in
+++ b/Dockerfile.in
@@ -46,14 +46,16 @@
 
 FROM {ARG_FROM}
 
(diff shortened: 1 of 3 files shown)


stp-ip:
/approve
/lgtm

Small nit that someone might be relying on the underlying structure of the current link, but as we never documented the structure until now I'd say it is more akin to using an internal API, which is subject to breaking.

PR #351, opened 2021-03-12
Change the symlink targets to just the SHA (v4)

### LINKED ISSUE #345: Define readlink() as API to know current rev

thockin:
This allows users to call readlink() on the link and learn the current
checked out SHA.

Fixes #345 

Depends on #346


stp-ip (approved):
/approve
/lgtm

PR #353, opened 2021-03-15
Fix tests on master - reset needs "--" arg

thockin:
Now that the worktree dirs are the same name as the hash, `git reset`
doesn't know if we are asking for reset-to-hash or reset-the-dir.
Adding "--" makes that unambiguous.

Tests pass now.


stp-ip:
Took a look into other potential ambitiousness issues as well, but seems this is the only one. So all good.

/lgtm

PR #358, opened 2021-03-18
Add third-party licenses to the docker image

nan-yu:


Makefile
```diff
@@ -119,8 +119,17 @@ $(OUTBIN): .go/$(OUTBIN).stamp
 # Used to track state in hidden files.
 DOTFILE_IMAGE = $(subst /,_,$(IMAGE))-$(TAG)
 
+THIRD_PARTY_DIR=.third-party
+
+third-party-licenses:
+	rm -rf $(THIRD_PARTY_DIR)
+	go install github.com/google/go-licenses
```
thockin, line 126: When I run this I get:

```
$ go install github.com/google/go-licenses
cannot find package "." in:
	/home/thockin/src/go/src/k8s.io/git-sync/vendor/github.com/google/go-licenses
```

Go modules baffle me.  If I run with `-mod=mod` it installs.  But then it modifies all the modules metadata, but doesn't actually vendor the code.  `go mod vendor` ...doesn't vendor.  `go mod tidy` cleans up but only partway.

Looking around, it seems this is a general Go problem still.  Playing it seems like this is the way:

1)  Write a new file `go.tools.mod`:
```
module k8s.io/git-sync

go 1.15
```

2) Run `go get -modfile=go.tools.mod github.com/google/go-licenses`.  This updates the new file.

3) Add go.tools.sum to `.gitignore`

4) The rest of this PR, except step 2 instead of `go install`

5) Commit

This still stinks because it will self-update any time that repo changes, causing changes to a pretty meaningless file.

We could just .gitignore go.tools.mod, but then it could explode later and be totally opaque.

We could just `git reset` the go.tools.mod file after `go get`.  Then it's clean.  But it still writes to GOPATH.

Yucccccck

How about:

```
export GOPATH="$(pwd)/.gopath.tools"
export GOBIN="$(pwd)/bin/tools"
mkdir -p "$GOBIN"
go get -modfile=go.tools.mod github.com/google/go-licenses
```

The call it from `./bin/tools/go-licenses`.  In `make clean` do `chmod -Rf u+w .gopath.tools` and then `rm -rf` it?  With .gitignore of `go.tools.sum` and `.gopath.tools`, we should keep the mess contained..

thockin: I'm asking go experts for guidance

nan-yu: Updated the `third-party-license` target to use `go-license` binary from `./bin/tools` directory.

thockin: Let me pause this while I talk to some go maintainers about best-practices.

thockin:
Replaced by #365 and #366

boypssy09:
Enviado desde mi iPhone

boypssy09:
Enviado desde mi iPhone

boypssy09:
Cuando quieres venir a comer Culo a casa y me puedes sorprender por el patio estoy en la espera 

Enviado desde mi iPhone

boypssy09 (commented):
[7876793896 let's get together now rear his open door for you asp come in for I welcome you now asp ]()

boypssy09:
> 

[where are you now ]()

PR #360, opened 2021-03-18
Don't require a TTY to build/test

### LINKED ISSUE #359: Builds require TTY again?
Builds via the Makefile currently require a TTY again, although it appears that this wasn't always the case (https://github.com/kubernetes/git-sync/pull/87).

Given this issue has been around awhile, I guess it doesn't affect too many folks, but it would be great to re-apply this no-TTY change so current branches can be rebuilt by CI systems that don't allocate one.

thockin:
Fixes #359


stp-ip (approved):
/lgtm

PR #365, opened 2021-03-30
Copy all licenses into the container image (v3)

thockin:
As per discussion with Go team.  it's hacky but it works.
    
    create tools.go
    git add tools.go
    go mod vendor
    git add vendor
    go mod tidy
    go mod vendor # again

Replaces #358


stp-ip:
Not fully on board with checking in the vendor folder, but ack.
/lgtm

PR #366, opened 2021-03-30
Copy all licenses into the container image (v4)

thockin:
As per discussion with Go team.  it's hacky but it works.
    
    create tools.go
    git add tools.go
    go mod vendor
    git add vendor
    go mod tidy
    go mod vendor # again

Replaces #358


stp-ip:
Not fully on board with checking in the vendor folder, but ack.
/lgtm

thockin:
I trust nobody when it comes to code availability.  It seems extreme, but
vendoring for the binary build (as opposed to a library repo) feels better
to me?

stp-ip:
I would have fully agreed before proxy/mirror. Now it's less of a concern and am ok with vendoring as well, but wanted to give context to the ack.

Issue #369, opened 2021-04-02
Move the git-sync-hook executed after the symlink


yogyang:
Hi, guys, wonder could we move the git-sync-hook-command after the symlink.
https://github.com/kubernetes/git-sync/blob/32828834a7a0f6c77d7ffb565a071afddce536b8/cmd/git-sync/main.go#L666-L676

The scenario is I want to put some scripts in my remote repo, and I would like to execute on_release.sh when git-sync done.
So the hook command would be like `sh /{dist}/script/on_release.sh`,  with the symlink not done yet, I can not set the correct script path.

thockin:
The exec is run in the worktree root - is that not sufficient?

Anyway, it seems like a reasonable suggestion.

PR #371, opened 2021-04-06
Move sync-hook to after symlink flip

thockin:
xref #369

### DIFF
--- a/.gitallowed
+++ b/.gitallowed
@@ -1,0 +1,1 @@
+xxxyyyzzz
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -663,32 +663,42 @@ func addWorktreeAndSwap(ctx context.Context, gitRoot, dest, branch, rev string, 
 		}
 	}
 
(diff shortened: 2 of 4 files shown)


.gitallowed
```diff
@@ -0,0 +1 @@
+xxxyyyzzz
```
stp-ip, line 1: Can you give context here?

stp-ip: Why do we need this checked in?

stp-ip: Generally would say this is more a workstation config than a repository specific one similar to editor configs, which I usually agree shouldn't be checked in.
Any reason it should be checked in besides making live easier?
(Should that be enough of a reason. Actually not sure on our stance here as this relates to easier maintainer life.)

thockin:
I would expect anyone who uses git secrets will eventually hit this
warning?  It's not that "xxxyyyzzz" is a safe password for all repos, but
IN THIS REPO that's the example given..

stp-ip:
Fair point.
/approve

PR #375, opened 2021-04-09
Export the error details to an error file

nan-yu:
The current git-sync process outputs the error information to standard
out, which is inaccessible from outside the container. Users have to
dump the logs using kubectl logs in order to check the error details in
the git-sync process. This commit exports the error details to a file,
which provides users the capability to check the errors directly from
other sidecar containers.

proposal: https://github.com/kubernetes/git-sync/issues/326


thockin (commented):
Can you please also update the README and add this flag to the "manual" string at the end of main.go ?

nan-yu:
> Can you please also update the README and add this flag to the "manual" string at the end of main.go ?

DONE

thockin (commented):
Thanks!

/lgtm
/approve

PR #376, opened 2021-04-14
Update to latest base image debian-base:buster-v1.6.0

amccague:
Uses the latest promoted debian base image: https://github.com/kubernetes/k8s.io/pull/1891

Overcomes two published CVEs:
* https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-24032
* https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-24031

My organisation cares about published and fixed vulnerabilities and we use this tool extensively, thank you for it :)

PR #377, opened 2021-04-15
Update to latest base image debian-base:buster-v1.6.0 (v4)

thockin:
Same as #376 but on master branch


stp-ip:
Retroactive lgtm

PR #378, opened 2021-04-16
Fix licenses for other arch'es

thockin:


thockin:
self-merge to make a release

stp-ip:
Retroactive lgtm

PR #385, opened 2021-05-21
Create the root directory if it doesn't exist

nan-yu:
The `git clone` command will create the root directory if it doesn't
exist, but if `git clone` fails, the root directory needs to be present
so that we can write the error to a file under the directory.


thockin (commented):
Thanks!

/lgtm
/approve

PR #389, opened 2021-05-24
Grant read access for the error file to all users

nan-yu:


thockin:
Super-nit - 0644 - no need for u+x

nan-yu:
> Super-nit - 0644 - no need for u+x

DONE

thockin (commented):
Thanks!

/lgtm
/approve

Issue #395, opened 2021-05-27
Upgraded libcurl3-gnutls is broken (fails to clone repo in >= 3.3.1)


shaglund:
I tried to upgrade git-sync from 3.2.1 to 3.3.1 but clone fails with unable to access repo "Failed sending HTTP2 data".

```
INFO: detected pid 1, running init handler
I0527 10:18:13.784232      13 main.go:504]  "level"=0 "msg"="starting up"  "args"=["/git-sync","--repo=https://github.com/kubernetes/git-sync.git","--branch=master"] "pid"=13
I0527 10:18:13.784405      13 main.go:831]  "level"=0 "msg"="cloning repo"  "origin"="https://github.com/kubernetes/git-sync.git" "path"="/tmp/git"
E0527 10:18:13.888791      13 main.go:172]  "msg"="too many failures, aborting" "error"="Run(git clone --no-checkout -b master https://github.com/kubernetes/git-sync.git /tmp/git): exit status 128: { stdout: \"\", stderr: \"Cloning into '/tmp/git'...\\nfatal: unable to access 'https://github.com/kubernetes/git-sync.git/': Failed sending HTTP2 data\\n\" }"  "failCount"=0
```

The issue seems to be related to libcurl in debian, see https://superuser.com/q/1642858

thockin:
Ooh fun.  I'll look into it, but I suggest you stay on 3.2.1 until we get a proper fix from upstream.

thockin:
It looks like I can manually downgrade that lib.  Looking into details.

thockin:
REMINDER: revert #398 and #399 once upstream bug is fixed.

@shaglund - do you have a local repro we could build into our e2e?

shaglund:
@thockin the log above I got from `docker run --rm k8s.gcr.io/git-sync/git-sync:v3.3.1 --repo=https://github.com/kubernetes/git-sync`

If that is what you mean by local repro?

thockin:
I guess I really meant "someone should find a way that reproduces against a
local test server, so this never slips thru again".  Our e2e has a lot of
cases, but it assumes file:// for most.  I'm SWAMPED right now, so any help
anyone can give is appreciated.

thockin:
Fixed in v3.3.2, pending promotion of that image

PR #398, opened 2021-05-27
Downgrade libcurl to work around HTTP bug (v3)

thockin:
xref https://github.com/kubernetes/git-sync/issues/395

This also cleans up the image build process to result in less layers


thockin:
FTR: this blocks a 3.3.2 release, which fixes a totally broken 3.3.1

stp-ip (approved):
/lgtm

PR #399, opened 2021-05-27
Downgrade libcurl to work around HTTP bug (v4)

thockin:
xref https://github.com/kubernetes/git-sync/issues/395

This also cleans up the image build process to result in less layers


stp-ip (approved):
/lgtm

PR #400, opened 2021-05-29
Change libcurl workaround to be less precise (v3)

thockin:
This should work across platforms more cleanly.

### DIFF
--- a/Dockerfile.in
+++ b/Dockerfile.in
@@ -57,15 +57,11 @@ RUN apt-get -y install --no-install-recommends \
         ca-certificates \
         coreutils \
         socat \
-        openssh-client
+        openssh-client \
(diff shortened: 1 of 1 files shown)


thockin:
self-merge to cut a release

stp-ip:
Not sure I get the reasoning here. Wasn't the point downgrading libcurl?

thockin:
The way it was set up before, we install the backports git, which
dragged the backported libcurl.  Then we downgraded it to stable.

The way it is now, we install the stable git, which drags in stable
libcurl, then upgrade ONLY git.  Same net result, but we don't have to
specify a particular version or architecture.

stp-ip:
Thanks for clarifying

PR #401, opened 2021-05-29
Change libcurl workaround to be less precise (v4)

thockin:
This should work across platforms more cleanly.


thockin:
self-merge to cut a release

PR #408, opened 2021-06-15
Improve error message for askpass.

briantkennedy:
When endpoint returns non-200 status, include the body in the error
message since it can contain useful information for debugging.  Also
defer closing the response body ReadCloser as this may have leaked in
the past.

PR for release-3.x branch.


thockin (commented):
Thanks!

/lgtm
/approve

PR #410, opened 2021-06-18
Add a git log after a git fetch

barney-s:
Adding a git log after fetch to record the commit that was fetched recently.
This would be useful to debug hard to reproduce potential race conditions b/w git ls-remote and git fetch.


thockin (commented):
Is this still a WIP?  I don't see what it's doing...

barney-s:
Dont need this anymore.

Issue #412, opened 2021-06-18
add validation after git fetch


Liujingfang1:
When the git server has some caching issue, `git ls-remote` and `git fetch` may get inconsistent commits. Later this leads failure in the command `git reset --hard <COMMIT>`.

The failure can be avoid by adding a validation after `git fetch`. If the local head commit is different from the result of `git ls-remote`, it should retry or wait for the next sync.

thockin:
fixed

PR #417, opened 2021-06-22
Some collected cleanups on v4 

thockin:
Prepping to try to push ahead on the big rework


thockin:
merging so I can port #412 to v4

PR #418, opened 2021-06-22
Clean up worktree dirs during sync loop

thockin:
Avoid "already exists" error.  This is a port of #412 but I changed
the e2e to use `docker pause`.

### DIFF
--- a/test_e2e.sh
+++ b/test_e2e.sh
@@ -425,6 +425,54 @@ assert_file_eq "$ROOT"/link/file "$TESTCASE 1"
 pass
 
 ##############################################
+# Test worktree-cleanup
+##############################################
+testcase "worktree-cleanup"
+
+echo "$TESTCASE" > "$REPO"/file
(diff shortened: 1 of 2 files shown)


test_e2e.sh
```diff
@@ -424,6 +424,54 @@ assert_file_eq "$ROOT"/link/file "$TESTCASE 1"
...
 
+##############################################
+# Test worktree-cleanup
+##############################################
+testcase "worktree-cleanup"
+
+echo "$TESTCASE" > "$REPO"/file
+git -C "$REPO" commit -qam "$TESTCASE"
+GIT_SYNC \
+    --period=100ms \
```
barney-s, line 435: This needs to be --wait

barney-s:
@thockin 
Sorry pulled the trigger too early.
The parameter "--period" needs to be "--wait" ?

thockin:
On v4 it is --period

PR #434, opened 2021-07-03
Remove broken analytics links

xinydev:
Remove broken analytics links, same as https://github.com/kubernetes/kubernetes/issues/93949

/kind cleanup


thockin (commented):
Thanks!

/lgtm
/approve

PR #435, opened 2021-07-05
Add option to read passwd from file (v3 branch)

thockin:
This is a port of #431.

A new flag `--password-file` is added. This allows git-sync to read
password from file and this is considered as safer than reading from
env or flag directly.

Few more checks are added as well:

1. `--password` and `--password-file` can't be specified at the same
time.
2. If `--username` is specified, then one of `--password` or
`--password-file` must be specified.


stp-ip (approved):
/lgtm

PR #437, opened 2021-07-08
Add syncHookCommand error metric

eminugurkenar:
added following metric

```
#HELP git_sync_hook_command_last_error Whether the last sync hook command resulted in an error (1 for error, 0 for success).
#TYPE git_sync_hook_command_last_error gauge
git_sync_hook_command_last_error 1
```

use case:
if you run git-sync with options `--sync-hook-command` and `--max-sync-failures=-1`, the following can happen

- syncs the repo 
- runs syncHookCommand and it fails
- it increments git_sync_count_total metric for error , so far good.
- but as hash of files on disk and repo is same , on next-run it does not retry to sync repo again and so it does not retry to run syncHookCommand , 
- if you create your alert on rate of this counter ,incase of command failure, after sometime your alert gets resolved, because it does not keep increasing error as there is no error now
- if you create your alert to just check whether git_sync_count_total{error} is bigger than 0 or not, after fixing the problem on command you have to restart the process , in order to resolve alert, which is not ideal.

so , added this metric to keep track last status of syncHookCommand.


thockin:
As an alternative - would it make more sense to keep track of the last hash we notified on and keep retrying notifications (of all flavors) until they each succeed?  Webhook notification already (kind of) does this.

This whole subsystem grew piecemeal.  Maybe we should just retool it completely, make them work correctly and similarly, and add the right metrics there.

That probably includes sending notifications at startup (or checkpointing state, yick).

?

eminugurkenar:
* would it make more sense to keep track of the last hash we notified on and keep retrying notifications

for some cases retrying would not change anything i think, imagine you have a some set of config filse in the repo and with syncHookCommand you render this config file to an other format , in this case if something wrong with config files in repo , this command will always fail , and retrying will not solve it as those config files in repo is still in bad state.

but ofc , retrying would solve some cases , syncHookCommand could be depended on an other service/process ,and if tool retries, eventually , it can get succeed.

and having same behaviour (imp) for syncHookCommand and webhook would be usefull for long term.

but for the current situation , i think the only possible option is at least having a metric like this to indicate current state of syncHookCommand.

thockin:
A few thoughts:

We already retry webhook, so this "problem" exists - it is just
inconsistent between webhook and exec.

If we are retrying then a counter metric that indicates how many times it
has failed is appropriate (as opposed to "failed last time").

I'm all for adding metrics, I just don't want to add hacky metrics because
we didn't think it through :)

eminugurkenar:
Yes i agree , i just thought maybe this "quick" solution is enough for this release, but for long-term, retry and a counter metric is actually right direction.

currently syncHookCommand is part of the syncRepo (called from) function , so would you separate it from syncRepo , and have syncHookCommand in a gorutine like webhook and only notify it if hash changed and ofc keep retrying? 

if yes i think i can try this ?

eminugurkenar:
I did some changes according to what would be good fit for your idea

thockin:
I really like where this is headed, thanks.  I will not have time for an in-depth review until next week earliest, but I wanted to let you know I read it over and I like it.

cmd/git-sync/cmdhook_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+package main
+
+import (
+	"testing"
+	"time"
+)
+
+func TestCmdhookDo(t *testing.T) {
+	t.Run("test not zero return code", func(t *testing.T) {
```
thockin, line 25: can you also add a "true" case? :)

thockin: and a `sleep 2` case proving the timeout ?

cmd/git-sync/hook_test.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+const (
+	hash1 = "1111111111111111111111111111111111111111"
+	hash2 = "2222222222222222222222222222222222222222"
+)
+
+func TestHookData(t *testing.T) {
+	t.Run("hook consumes first hash value", func(t *testing.T) {
+		whd := NewHookData()
```
thockin, line 31: `whd` was "webhook data" -> should be `hd` or `data` now ?

cmd/git-sync/main.go
```diff
@@ -79,6 +79,10 @@ var flChmod = flag.Int("change-permissions", envInt("GIT_SYNC_PERMISSIONS", 0),
 var flSyncHookCommand = flag.String("sync-hook-command", envString("GIT_SYNC_HOOK_COMMAND", ""),
 	"the command executed with the syncing repository as its working directory after syncing a new hash of the remote repository. "+
 		"it is subject to the sync time out and will extend period between syncs. (doesn't support the command arguments)")
+var flSynchookCommandTimeout = flag.Duration("sync-hook-command-timeout", envDuration("GIT_SYNC_HOOK_COMMAND_TIMEOUT", time.Second),
```
thockin, line 82: because this had no timeout before, I think the default needs to be more like 30 seconds

cmd/git-sync/main.go
```diff
@@ -144,6 +148,11 @@ var (
 		Help: "How many git syncs completed, partitioned by state (success, error, noop)",
 	}, []string{"status"})
 
+	hookRunCount = prometheus.NewCounterVec(prometheus.CounterOpts{
```
thockin, line 151: This is combined Webhook and Cmdhook, right?  The thinking, I guess, is that most people won't do both?

thockin: The help should b e clear that this is across all hook types

cmd/git-sync/main.go
```diff
@@ -1075,7 +1101,6 @@ func runCommand(ctx context.Context, cwd, command string, args ...string) (strin
 
 func runCommandWithStdin(ctx context.Context, cwd, stdin, command string, args ...string) (string, error) {
 	cmdStr := cmdForLog(command, args...)
-	log.V(5).Info("running command", "cwd", cwd, "cmd", cmdStr)
```
thockin, line 1078: these logs are useful, please don't remove them :)

cmd/git-sync/main.go
```diff
@@ -1096,7 +1121,6 @@ func runCommandWithStdin(ctx context.Context, cwd, stdin, command string, args .
 	if err != nil {
 		return "", fmt.Errorf("Run(%s): %w: { stdout: %q, stderr: %q }", cmdStr, err, stdout, stderr)
 	}
-	log.V(6).Info("command result", "stdout", stdout, "stderr", stderr)
```
thockin, line 1099: same

cmd/git-sync/hook.go
```diff
@@ -0,0 +1,115 @@
...
+	// Holds the data as it crosses from producer to consumer.
+	Data *hookData
+}
+
+func (r *HookRunner) Send(hash string) {
+	r.Data.send(hash)
+}
+
+// Wait for trigger events from the channel, and run hook when triggered
+func (r *HookRunner) run() {
```
thockin, line 85: can we make this `Run()` ?  It's part of the exposed API.

thockin (commented):
Overall this looks great.  I wonder if we should move the hook logic to a top-level `pkg/hook` directory and move webhook and cmd under that, to emphasize the API encapsulation.

Looking at how I would port this to v4 branch, I think I will deprecate the `--sync-hook-command` and make it more congruent to webhook as `--exechook-command` - do you want to tackle that here, so it is the same?  If not, I can do it eventually.  We can still allow the old flag, but it just means the same thing as the new.

We probably also need an e2e test that demonstrates the retry logic (webhook has this, you can copy that)

cmd/git-sync/cmdhook_test.go
eminugurkenar: this is done.

cmd/git-sync/hook_test.go
eminugurkenar: it is now "hd".

cmd/git-sync/main.go
eminugurkenar: set it 30, done.

cmd/git-sync/main.go
eminugurkenar: this is actually, to be able to test runCommand , test case fails as "log" is not inject during tests ?

cmd/git-sync/hook.go
eminugurkenar: will work on this with "pkg/hook"

cmd/git-sync/main.go
eminugurkenar: yes it combines hooks under metric `git_sync_hook_run_count_total` , this has a benefit over having two metric  i guess 
you can handle both cases with just alert something like below

`rate(git_sync_hook_run_count_total{status="error"}[5m]) > 0`

but ofc we can devide it two like

```
git_sync_webhook_run_count_total
git_sync_cmdhook_run_count_total
```
---------

"How many hook runs completed, partitioned by name and state "

accually by saying "by name and state" , i thought it is enough as metrics has labels like below, but apparently not:<

```
git_sync_hook_run_count_total{name="cmdhook",status="error"} 5
git_sync_hook_run_count_total{name="webhook",status="error"} 2
```

i think we can think about description , after deciding whether we are going to devide the metric or use labels ?

thockin: I missed that it was partitioned by name.  That seems fine to me.

cmd/git-sync/main.go
thockin: ok, so we probably need to initialize `log` as `&customLogger{logr.Discard(), "", ""}`

thockin:
Let me know when this is ready to re-review. :)

cmd/git-sync/cmdhook.go
```diff
@@ -0,0 +1,47 @@
...
+	// Timeout for the command
+	Timeout time.Duration
+}
+
+func (w *Cmdhook) Name() string {
+	return "cmdhook"
+}
+
+func (c *Cmdhook) Do(hash string) error {
+	ctx, cancel := context.WithTimeout(context.Background(), c.Timeout)
```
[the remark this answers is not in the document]
eminugurkenar: now gets it as argument.

cmd/git-sync/main.go
eminugurkenar: created cmd package which is responsible of running commands and it keeps logger as struct field.
reverted logs back.

cmd/git-sync/main.go
eminugurkenar: reverted logs back.

cmd/git-sync/hook.go
eminugurkenar: created pkg/hook which exports function Run()

eminugurkenar:
* I wonder if we should move the hook logic to a top-level pkg/hook directory and move webhook and cmd under that, to emphasize the API encapsulation.

created pkg/hook also created as pkg/cmd we need to reach RunCommand/RunCommandWithStdin functions, i encapsulated this functions in pkg/cmd

for logging created one more package pkg/logging which is responsible of implementing logging functions and so we can pass this logger arround which simplify testing as well. I didnt want to initialize new customLogger everytime a command runs.

eminugurkenar:
* I think I will deprecate the --sync-hook-command and make it more congruent to webhook as --exechook-command

deprecated sync-hook-command but we still allow it , if it is set and exechook-command not set we use it otherwise we always use exechook-command.

eminugurkenar:
* We probably also need an e2e test that demonstrates the retry logic (webhook has this, you can copy that)

created a e2e testcase named "exechook-fail-retry" which checks number of lines of "/var/log/runs" similar to webhook , if it is less than 2 it fails.

Also renamed "sync_hook_command" testcase to exechook-success to make it similar to webhook-success testcase.

eminugurkenar:
@thockin i think it is ready.

cmd/git-sync/main.go
```diff
@@ -77,15 +76,22 @@ var flMaxSyncFailures = flag.Int("max-sync-failures", envInt("GIT_SYNC_MAX_SYNC_
 var flChmod = flag.Int("change-permissions", envInt("GIT_SYNC_PERMISSIONS", 0),
 	"the file permissions to apply to the checked-out files (0 will not change permissions at all)")
 var flSyncHookCommand = flag.String("sync-hook-command", envString("GIT_SYNC_HOOK_COMMAND", ""),
+	"(deprecated) the command executed with the syncing repository as its working directory after syncing a new hash of the remote repository. "+
```
thockin, line 79: The v4 branch set a convention here.  Let's do `"DEPRECATED: use --exechook-command instead"`

cmd/git-sync/main.go
```diff
@@ -77,15 +76,22 @@ var flMaxSyncFailures = flag.Int("max-sync-failures", envInt("GIT_SYNC_MAX_SYNC_
...
+		"it is subject to the sync time out and will extend period between syncs. (doesn't support the command arguments)")
+var flExechookCommand = flag.String("exechook-command", envString("GIT_EXECHOOK_COMMAND", ""),
 	"the command executed with the syncing repository as its working directory after syncing a new hash of the remote repository. "+
```
thockin, line 82: nits, but let's clean up the grammar:

"a command to be executed (without arguments, with the syncing repository as its working directory) after syncing a new hash of the remote repository.  It is subject to --timeout out and will extend period between syncs."

cmd/git-sync/main.go
```diff
@@ -129,7 +135,8 @@ var flHTTPMetrics = flag.Bool("http-metrics", envBool("GIT_SYNC_HTTP_METRICS", t
 var flHTTPprof = flag.Bool("http-pprof", envBool("GIT_SYNC_HTTP_PPROF", false),
 	"enable the pprof debug endpoints on git-sync's HTTP endpoint")
 
-var log *customLogger
+var cmdRunner *cmd.CommandRunner
+var logger *logging.Logger
```
thockin, line 139: rather than change every log-site, can we just leave it as "log"

pkg/cmd/cmd.go
```diff
@@ -0,0 +1,86 @@
...
+	"os/exec"
+	"strings"
+
+	"k8s.io/git-sync/pkg/logging"
+)
+
+// CommandRunner structure
+type CommandRunner struct {
+	// Logger
+	logger *logging.Logger
```
thockin, line 32: this should take and store a `logr.Logger`

pkg/hook/cmdhook.go
```diff
@@ -0,0 +1,71 @@
...
+	"context"
+	"path/filepath"
+	"time"
+
+	"k8s.io/git-sync/pkg/cmd"
+	"k8s.io/git-sync/pkg/logging"
+)
+
+// Cmdhook structure, implements Hook
+type Cmdhook struct {
```
thockin, line 29: I hate myself, but can you change these to Exechook (including the filename :)

pkg/cmd/cmd.go
```diff
@@ -0,0 +1,86 @@
...
+	"context"
+	"fmt"
+	"os/exec"
+	"strings"
+
+	"k8s.io/git-sync/pkg/logging"
+)
+
+// CommandRunner structure
+type CommandRunner struct {
```
thockin, line 30: Just "Runner" (to follow Go's "don't stutter" idiom)

pkg/hook/hook_test.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Kubernetes Authors All rights reserved.
+Copyright 2021 The Kubernetes Authors All rights reserved.
```
thockin, line 2: don't update the copyright please

pkg/logging/logging.go
```diff
@@ -0,0 +1,152 @@
...
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/go-logr/glogr"
+	"github.com/go-logr/logr"
+)
+
+// A logger embeds logr.Logger
+type Logger struct {
```
thockin, line 31: this can be a private type.  New() should return `logr.Logger`

pkg/logging/logging.go
```diff
@@ -0,0 +1,152 @@
...
+	log       logr.Logger
+	root      string
+	errorFile string
+}
+
+// NewLogger returns glogr implemented logr.Logger.
+func NewLogger(root string, errorFile string) *Logger {
```
thockin, line 38: Just "New"

thockin (commented):
This is an awesome PR.  Ideally it would be a small number of complete commits.  E.g.

* move logging to a package
* move cmd to a package
* refactor webhook to a generic hook subsystem
* reimplement hook command as a generic hook

I don't know if that's plausible at this point, so if not, let's just squash down to 1 commit :)

cmd/git-sync/main.go
eminugurkenar: changed description.

cmd/git-sync/main.go
eminugurkenar: updated description.

cmd/git-sync/main.go
eminugurkenar: reverted back to log.

pkg/hook/cmdhook.go
eminugurkenar: now we use Exechook on everywhere.

pkg/cmd/cmd.go
eminugurkenar: changed.

pkg/hook/hook_test.go
eminugurkenar: mistake , reverted.

pkg/logging/logging.go
eminugurkenar: it is New() now.

pkg/logging/logging.go
eminugurkenar: but looks like if we return logr.Logger we cant reach "custom Error()" and other functions (deleteErrorFile...)?

pkg/cmd/cmd.go
eminugurkenar: depends on returning logr.Logger from logging package , asked below.

eminugurkenar:
@thockin did all of them , except returning logr.Logger, if we return it we cant reach custom Error() function , as logr.Logger is not an interface. So this is why accually i wrapped it.

eminugurkenar:
Squashed all commits to one.

pkg/logging/logging.go
eminugurkenar: we can implement LogSink  interface and just call same functions again maybe but this still can't solve DeleteErrorFile() function?

thockin: I forgot about the custom methods.  That's fine.

pkg/hook/exechook.go
```diff
@@ -0,0 +1,71 @@
...
+		command:   command,
+		gitRoot:   gitroot,
+		args:      args,
+		timeout:   timeout,
+		logger:    l,
+	}
+}
+
+// Name describes hook, implements Hook.Name
+func (w *Exechook) Name() string {
```
thockin, line 57: super duper nit, and if this is the only one, I will ignore it.

The `this` name in the methods is inconsistent.  I'd go with `h`

thockin (commented):
Thanks!  I'll need to port this to the v4 branch.  That's on me. :)

/lgtm
/approve

pkg/hook/exechook.go
eminugurkenar: yep, couldnt notice it:<

Issue #404, opened 2021-06-10
Invalid reference when changing branches


dhuang:
Hello. I'm on v3.3.0 and got things working nicely with an initial value of `GIT_SYNC_BRANCH` as `main`. However if I try to update the branch to anything else on my _existing_ volume, I'm getting "invalid reference" errors.

```
INFO: detected pid 1, running init handler
I0610 01:28:06.722177      11 main.go:495]  "level"=0 "msg"="starting up"  "args"=["/git-sync"] "pid"=11
I0610 01:28:07.135477      11 main.go:888]  "level"=0 "msg"="update required"  "local"="ed4da7e65a9102d789db2674abeafee07e42ff45" "remote"="4ddcd3cfec8da15ccb5562e3bed3a572674e05af" "rev"="HEAD"
I0610 01:28:07.135509      11 main.go:666]  "level"=0 "msg"="syncing git"  "hash"="4ddcd3cfec8da15ccb5562e3bed3a572674e05af" "rev"="HEAD"
I0610 01:28:08.603194      11 main.go:687]  "level"=0 "msg"="adding worktree"  "branch"="origin/test" "path"="/git/4ddcd3cfec8da15ccb5562e3bed3a572674e05af"
E0610 01:28:08.603337      11 main.go:168]  "msg"="too many failures, aborting" "error"="Run(git worktree add /git/4ddcd3cfec8da15ccb5562e3bed3a572674e05af origin/test): exit status 128: { stdout: \"\", stderr: \"fatal: invalid reference: origin/test\\n\" }"  "failCount"=0
```

I'm certain `origin/test` is a valid and pushed branch reference. Just to double I cloned the repo from scratch and ran `git worktree add ./git/4ddcd3cfec8da15ccb5562e3bed3a572674e05af origin/test` locally with no issues. Changing `GIT_SYNC_BRANCH` back to `main` makes everything happy again.

I suspect if I start from scratch on a fresh volume, the `test` branch it'll work, but wondering if this supposed to work or if I'm doing something wrong. Thanks in advance!

kbumsik-angelswing:
I also have the same problem.
I used a persistent volume (PVC) with git-sync, but it throws the same `invalid reference` error when I restart the Pod (Container) with the different branch.

maxkochubey:
I faced with the same issue.
Currently I am using the next workaround (the part of `git-sync` container template in a pod manifest):

```
command: ["/bin/sh","-c"]
[...]

PR #445, opened 2021-09-28
use branch revision during worktree add

natbaker:
Change the worktree add call to use the revision hash, avoiding invalid reference issue when using branch name

Fixes kubernetes/git-sync#404

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -684,11 +684,6 @@ func addWorktreeAndSwap(ctx context.Context, gitRoot, dest, branch, rev string, 
 		return nil
 	}
 
-	// GC clone
-	if _, err := cmdRunner.Run(ctx, gitRoot, *flGitCmd, "gc", "--prune=all"); err != nil {
-		return err
-	}
-
 	// Make a worktree for this exact git hash.
 	worktreePath := filepath.Join(gitRoot, hash)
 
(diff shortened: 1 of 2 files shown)


thockin:
Can you help me by whipping up a case for test_e2e.sh ?  That will  make sure I understand the conditions that caused failure and that I don't regress :)

natbaker:
The scenario seems to be if the repo is initially cloned with one branch and with depth>0 then later switched to another branch you get the invalid reference error. Changing depth after the repo is cloned doesn't have any effect, since that is only used in the clone and the fetch.

I added a test that replicates this scenario, starting with e2e-branch at depth=1, then switching to another branch.

test_e2e.sh
```diff
@@ -426,6 +426,43 @@ assert_file_eq "$ROOT"/link/file "$TESTCASE 1"
...
+echo "$TESTCASE 2" > "$REPO"/file
+git -C "$REPO" commit -qam "$TESTCASE 2"
+GIT_SYNC \
+    --one-time \
+    --repo="file://$REPO" \
+    --branch=$BRANCH \
+    --rev=HEAD \
+    --root="$ROOT" \
+    --dest="link" \
+    > "$DIR"/log."$TESTCASE" 2>&1
```
thockin, line 459: nit: this needs to be `>>` on the second run

cmd/git-sync/main.go
```diff
@@ -702,12 +697,17 @@ func addWorktreeAndSwap(ctx context.Context, gitRoot, dest, branch, rev string,
 		return err
 	}
 
-	_, err := cmdRunner.Run(ctx, gitRoot, *flGitCmd, "worktree", "add", worktreePath, "origin/"+branch, "--no-checkout")
+	_, err := cmdRunner.Run(ctx, gitRoot, *flGitCmd, "worktree", "add", "-d", worktreePath, hash, "--no-checkout")
```
thockin, line 700: can you please change `-d` to `--detach`?

thockin (commented):
Thanks!  I get the problem now.  Testcase was very helpful.  Just 2  small  nits

test_e2e.sh
natbaker: Ah, yep, good call.

test_e2e.sh
```diff
@@ -456,7 +456,7 @@ GIT_SYNC \
     --rev=HEAD \
     --root="$ROOT" \
     --dest="link" \
-    > "$DIR"/log."$TESTCASE" 2>&1
+    > "$DIR"/log."$TESTCASE" 2>>&1
```
thockin, line 459: sorry I meant `>> "$DIR"/log."$TESTCASE" 2>&1`

natbaker: 🤦 I knew I went too fast

thockin (commented):
Thanks!

/lgtm
/approve

PR #446, opened 2021-10-04
use branch revision during worktree add (v4)

thockin:
V4 branch port of #445


thockin:
cc @natbaker FYI - porting to v4 dev branch

PR #449, opened 2021-10-12
Fix password-file flag declaration

and-1:
This PR fix declaration of password-file flag


thockin (commented):
Thanks!

/lgtm
/approve

vanugrah:
HI there  - is there a plan to release a new patch with the password-file arg?

thockin:
Looking at a release now

PR #451, opened 2021-11-02
v4: use repo default branch instead of master

### LINKED ISSUE #444: Error cloning repositories without branch "master"
Hi!

I have an issue when flag GIT_SYNC_BRANCH is not set. This component requires know an existing branch of a repository or the repo must have a branch named "master".

Nowaways, new projects are created with main branch instead of master, and when we use this tool it throws an error.

I reviewed the repo and in this function

// CloneRepo does an initial clone of the git repo.
func (git *repoSync) CloneRepo(ctx context.Context) error {
	args := []string{"clone", "--no-checkout", "-b", git.branch}

and this line

_, err := git.run.Run(ctx, git.root, git.cmd, "worktree", "add", worktreePath, "origin/"+git.branch, "--no-checkout")

we can make git.branch optional and also change default value from "master" to "" in GIT_SYNC_BRANCH param.

I did some tests with modified code and it worked fine.

Let me know your thoughts.

Thank you in advance!

thockin:
This is ONLY on v4 branch - it's a breaking change for v3.

Fixes #444

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,6 @@ docker run -d \
     -u$(id -u):$(id -g) \
     registry/git-sync:tag \
         --repo=https://github.com/kubernetes/git-sync \
-        --branch=master \
         --root=/tmp/git/root \
         --period=30s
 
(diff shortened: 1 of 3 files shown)


stp-ip (approved):
/lgtm

PR #455, opened 2021-11-24
v4: Replace glogr with funcr

thockin:
No difference in flags.  Log output will now be JSON.

example:

```
{"logger":"","ts":"2021-11-24 11:19:46.340238","caller":{"file":"main.go","line":1031},"level":0,"msg":"cloning repo","origin":"https://github.com/thockin/git-sync","path":"/tmp/gt"}
```


stp-ip (approved):
/lgtm

PR #461, opened 2021-12-16
Avoid writing to /etc/passwd unless needed

mac-chaffee:
Some users of git-sync such as airflow allow users to set the runAsUser/runAsGroup, and as a result they always have to set GIT_SYNC_ADD_USER to true: https://github.com/airflow-helm/charts/blob/e1d49498426add959350ab8efacead8f96400759/charts/airflow/templates/_helpers/pods.tpl#L198-L226

But if the you don't override the default runAsUser/runAsGroup (65533) but you do have `GIT_SYNC_ADD_USER=true`, git-sync will write a duplicate entry to /etc/passwd.

I don't think duplicate entries are too much of a problem, but I mainly want this feature because falco spams me with alerts if a pod opens a file under /etc/ for writing :) https://github.com/falcosecurity/falco/blob/f035829ca2881b70a4d13e22abb5f8eb828d3a5f/rules/falco_rules.yaml#L1249

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -30,6 +30,7 @@ import (
 	"os"
 	"os/exec"
 	"os/signal"
+	"os/user"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


thockin:
Rather than coding the UID/GID in another place, can this be done in terms of https://pkg.go.dev/os/user#LookupId ? If that succeeds, we do not need to write passwd.

thockin (commented):
Thanks!

/lgtm
/approve

thockin (commented):
/hold

cmd/git-sync/main.go
```diff
@@ -164,6 +165,12 @@ const (
 // initTimeout is a timeout for initialization, like git credentials setup.
 const initTimeout = time.Second * 30
 
+// defaultUID is the default user ID as defined in the Dockerfile
+const defaultUID = 65533
```
thockin, line 169: these can go away now

thockin (commented):
I wonder if there's a way to test this...

thockin:
I'm running some manual tests. just to be 100%

Issue #463, opened 2022-01-06
GIT_SYNC_EXECHOOK_COMMAND + GIT_SYNC_ONE_TIME


ChrisERo:
Is a command specified by the EXECHOOK_COMMAND environment variable **always** supposed to execute after git sync pulls/clones a git repo? The documentation suggests this, but when GIT_SYNC_ONE_TIME is true, I do not see where this is enforced. It seems possible that the main thread could execute os.Exit before the exechook/consumer thread processes the "sync event" and executes the specified command.

I looked at version 3.3.4 and it looks like it does not have such a race condition, as the exechook is run by the same main thread.

Forgive me if this is not the correct venue to post this question/issue; it's my first "contribution" to an open-source project.

thockin:
Good question.  That intersection is not well defined.  Is it useful to
make such a guarantee?  If so, we need a way to signal back from the  exec
and webhook runners that they are done.  Or maybe signal "stop" to the
runners and they signal "done" back.  On the one-time path, before exit,
stop the runners.

The alternative would be to make those flags mutually-exclusive, which is a
breaking change, so at best we would just warn and not set up the runners.

PRs or simpler ideas welcome!

ChrisERo:
Hi @thockin, thanks for the prompt response!!

I would find this guarantee to be useful. In my personal use-case, I only need git-sync to perform a one-time 'git clone' on a git tag and store the source code in a specific, preexisting folder. Ideally, git-sync would be an init-container in this scenario, explaining my dependence on **GIT_SYNC_ONE_TIME**. To put the sourcecode in the right location, I am  relying on a shell script inside git-sync container's volume, referenced by **GIT_SYNC_EXECHOOK_COMMAND**.

I would be happy to take a crack at drafting this PR. I see that in the [previous release](https://github.com/kubernetes/git-sync/blob/v3.3.4/cmd/git-sync/main.go#L831), the sync-hook was executed in the same thread that terminates if the "ONE_TIME" flag is true. Is this an acceptable solution: removing the asynchrony involved with exec/sync hook?

thockin:
If we make this guarantee for exec, we should make it for webhook, too.
But I don't think we should serialize those, so the simple solution you
propose isn't really ideal.

ChrisERo:
Thanks for the insight!

I created the following PR, https://github.com/kubernetes/git-sync/pull/466, the code hasn't been tested, but I want to make sure this is the general strategy you were looking for before proceeding.

[...]

PR #466, opened 2022-01-08
fix: resolve issue #463

ChrisERo:
Resolved original issue by introducing a boolean chanel by which exechook runner can communicate with main thread.

Then introduced and used webhook executed-at-least-once chanel and added documentation explaining sections of of code only executed when git-sync pulls for first time.

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -174,8 +174,10 @@ var (
 	}, []string{"status"})
 )
 
-// Channels for ensuring hooks execute at least once before terminating due to GIT_SYNC_ONCE
-var execHookChannel, webHookChannel chan bool  // initialising as null to promptly catch logical errors and to avoid unneeded object creation
+// Channels for ensuring hooks execute at least once before terminating.
+// Should be nil if and only if corresponding hook is not defined and if initialised, will only every get written
+// to once.
+var exechookChannel, webhookChannel chan bool
 
 const (
 	metricKeySuccess = "success"
@@ -362,8 +364,7 @@ func main() {
 		if *flExechookBackoff < time.Second {
 			handleError(log, true, "ERROR: --exechook-backoff must be at least 1s")
 		}
-
-		execHookChannel = make(chan bool)
+		exechookChannel = make(chan bool, 1)
 	}
 
 	if *flWebhookURL != "" {
@@ -376,8 +377,7 @@ func main() {
 		if *flWebhookBackoff < time.Second {
 			handleError(log, true, "ERROR: --webhook-backoff must be at least 1s")
 		}
-
-		webHookChannel = make(chan bool)
+		webhookChannel = make(chan bool, 1)
 	}
 
 	if *flPassword != "" && *flPasswordFile != "" {
@@ -563,7 +563,7 @@ func main() {
 			*flWebhookBackoff,
 			hook.NewHookData(),
 			log,
-			webHookChannel,
+			webhookChannel,
 		)
 		go webhookRunner.Run(context.Background())
 	}
@@ -584,7 +584,7 @@ func main() {
 			*flExechookBackoff,
 			hook.NewHookData(),
 			log,
-			execHookChannel,
+			exechookChannel,
 		)
 		go exechookRunner.Run(context.Background())
 	}
@@ -630,17 +630,17 @@ func main() {
 		}
 
 		if initialSync {
-			// Wait for hooks to complete at least once before checking whether to stop
+			// Wait for hooks to complete at least once, if not nil, before checking whether to stop program
 			// Assumes that if hook channels are not nil, they will have at least one value before getting closed
-			if execHookChannel != nil {
-				execHookChannelFinishedSuccessfully:= <-execHookChannel
-				if !execHookChannelFinishedSuccessfully {
+			if exechookChannel != nil {
+				exechookChannelFinishedSuccessfully := <-exechookChannel
+				if !exechookChannelFinishedSuccessfully {
 					log.Error(nil, "exec hook completed with error")
 				}
 			}
-			if webHookChannel != nil {
-				webHookChannelFinishedSuccessfully:= <-webHookChannel
-				if !webHookChannelFinishedSuccessfully {
+			if webhookChannel != nil {
+				webhookChannelFinishedSuccessfully := <-webhookChannel
+				if !webhookChannelFinishedSuccessfully {
 					log.Error(nil, "web hook completed with error")
 				}
 			}
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -638,13 +638,17 @@ func main() {
 				// Assumes that if hook channels are not nil, they will have at
 				// least one value before getting closed
 				exitCode := 0 // is 0 if all hooks succeed, else is 1
-				if err = exechookRunner.WaitForCompletion(); err != nil {
-					log.Error(err, "exechook completed with error")
-					exitCode = 1
+				if exechookRunner != nil {
+					if err = exechookRunner.WaitForCompletion(); err != nil {
+						log.Error(err, "exechook completed with error")
+						exitCode = 1
+					}
 				}
-				if err = webhookRunner.WaitForCompletion(); err != nil {
-					log.Error(err, "webhook completed with error")
-					exitCode = 1
+				if webhookRunner != nil {
+					if err = webhookRunner.WaitForCompletion(); err != nil {
+						log.Error(err, "webhook completed with error")
+						exitCode = 1
+					}
 				}
 				log.DeleteErrorFile()
 				os.Exit(exitCode)
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -174,6 +174,9 @@ var (
 	}, []string{"status"})
 )
 
+// Channels for ensuring hooks execute at least once before terminating due to GIT_SYNC_ONCE
+var execHookChannel, webHookChannel chan bool  // initialising as null to promptly catch logical errors and to avoid unneeded object creation
+
 const (
 	metricKeySuccess = "success"
 	metricKeyError   = "error"
@@ -359,6 +362,8 @@ func main() {
 		if *flExechookBackoff < time.Second {
 			handleError(log, true, "ERROR: --exechook-backoff must be at least 1s")
 		}
+
+		execHookChannel = make(chan bool)
 	}
 
 	if *flWebhookURL != "" {
@@ -371,6 +376,8 @@ func main() {
 		if *flWebhookBackoff < time.Second {
 			handleError(log, true, "ERROR: --webhook-backoff must be at least 1s")
 		}
+
+		webHookChannel = make(chan bool)
 	}
 
 	if *flPassword != "" && *flPasswordFile != "" {
@@ -556,6 +563,7 @@ func main() {
 			*flWebhookBackoff,
 			hook.NewHookData(),
 			log,
+			webHookChannel,
 		)
 		go webhookRunner.Run(context.Background())
 	}
@@ -576,6 +584,7 @@ func main() {
 			*flExechookBackoff,
 			hook.NewHookData(),
 			log,
+			execHookChannel,
 		)
 		go exechookRunner.Run(context.Background())
 	}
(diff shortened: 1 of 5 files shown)


thockin:
Thanks!  I am on mobile now, so just some thoughts rather than full review.

If we do it this way, we don't guarantee regular hooks (not run-once)
complete.  That's a little gross but probably ok, and it does make this
simpler. We need to document the semantics can well.

We need to make it a buffered channel (size 1) and we need to make sure we
only ever write to the channel one time per hook impl (again, document it
well).  In the normal case nobody will be receiving, and we don't want
senders to block.

We will need test cases added to test_e2e.sh

Nit: "webhook" and "exechook" are treated like words, rather than "webHook"
(capital H).

cmd/git-sync/main.go
```diff
@@ -174,8 +174,10 @@ var (
 	}, []string{"status"})
 )
 
-// Channels for ensuring hooks execute at least once before terminating due to GIT_SYNC_ONCE
-var execHookChannel, webHookChannel chan bool  // initialising as null to promptly catch logical errors and to avoid unneeded object creation
+// Channels for ensuring hooks execute at least once before terminating.
+// Should be nil if and only if corresponding hook is not defined and if initialised, will only every get written
```
thockin, line 178: s/every/ever

also, please linewrap comments at 80

cmd/git-sync/main.go
```diff
@@ -621,6 +630,22 @@ func main() {
 		}
 
 		if initialSync {
+			// Wait for hooks to complete at least once, if not nil, before checking whether to stop program
```
thockin, line 633: please linewrap comments at 80

cmd/git-sync/main.go
```diff
@@ -621,6 +630,22 @@ func main() {
...
+			// Assumes that if hook channels are not nil, they will have at least one value before getting closed
+			if exechookChannel != nil {
+				exechookChannelFinishedSuccessfully := <-exechookChannel
+				if !exechookChannelFinishedSuccessfully {
```
thockin, line 637: nit: "webhook" "exechook"

cmd/git-sync/main.go
```diff
@@ -621,6 +630,22 @@ func main() {
 		}
 
 		if initialSync {
+			// Wait for hooks to complete at least once, if not nil, before checking whether to stop program
+			// Assumes that if hook channels are not nil, they will have at least one value before getting closed
+			if exechookChannel != nil {
```
thockin, line 635: I think we should only do this in case of `--one-time` - is there a good reason to check it otherwise?

This also would let us exit with a non-zero code if one or more hooks failed.  Accumulate a bool "hooksOK" - set it to false in case of error and `os.Exit(1)` at the end?

pkg/hook/hook.go
```diff
@@ -96,6 +96,9 @@ type HookRunner struct {
 	data *hookData
 	// Logger
 	logger *logging.Logger
+	// hasCompletedOnce is used to send true if and only if first run executed successfully and false otherwise to some receiver.
```
thockin, line 99: linewrap comments

test_e2e.sh
```diff
@@ -1072,6 +1074,81 @@ function e2e::exechook_fail_retry() {
     fi
 }
 
+##############################################
+# Test exechook-success with GIT_SYNC_ONE_TIME
+##############################################
+function exechook_success_once_helper() {
+    # First sync
+    echo "$FUNCNAME 2" > "$REPO"/file
```
thockin, line 1082: FUNCNAME 1 the first time thru :)

test_e2e.sh
```diff
@@ -1072,6 +1074,81 @@ function e2e::exechook_fail_retry() {
     fi
 }
 
+##############################################
+# Test exechook-success with GIT_SYNC_ONE_TIME
```
thockin, line 1078: document args

test_e2e.sh
```diff
@@ -1072,6 +1074,81 @@ function e2e::exechook_fail_retry() {
...
+    GIT_SYNC \
+        --period=${2}ms \
+        --one-time \
+        --repo="file://$REPO" \
+        --branch="$MAIN_BRANCH" \
+        --root="$ROOT" \
+        --link="link" \
+        --exechook-command="$EXECHOOK_COMMAND_SLEEPY" \
+        >> "$1" 2>&1 &
+    sleep 7
```
thockin, line 1094: can we reduce sleeps to minimum?

test_e2e.sh
thockin, line 1078: comment should go on the real test fn, the helper can just describe args and purpose

test_e2e.sh
```diff
@@ -1072,6 +1074,81 @@ function e2e::exechook_fail_retry() {
...
+    assert_file_exists "$ROOT"/link/file
+    assert_file_exists "$ROOT"/link/exechook
+    assert_file_exists "$ROOT"/link/link-exechook
+    assert_file_eq "$ROOT"/link/file "$FUNCNAME 2"
+    assert_file_eq "$ROOT"/link/exechook "$FUNCNAME 2"
+    assert_file_eq "$ROOT"/link/link-exechook "$FUNCNAME 2"
+}
+
+function e2e::exechook_success_once() {
+  for i in $(seq 0 50 200); do
```
thockin, line 1105: Why do we need this in a loop?  It should be sufficient to demonstrate that a slow hook actually completed one time (that main waitied for it), right?

thockin:
Also your first commit has a pr ref in the comment - our bot doesn't want that :)

cmd/git-sync/main.go
ChrisERo: I do not follow what this comment means; I suspect you want exechookChannel != nil to be exechook != nil. May you please clarify?

cmd/git-sync/main.go
ChrisERo: For the first question, the statement I'm trying to guarantee is the following: "Assuming git-sync thread has not terminated with an error code, exechook and webhook, if set,  will always run (not necessarily successfully) at least once." If the user git-syncs a hash for whatever reason, the proposed change would break this invariant we wish to establish. However, since tags are not considered commits in the source code, this won't affect my use case and I would be fine with doing this change if you would like.

As for the boolean thing...let me think about it. [TBT]

cmd/git-sync/main.go
thockin: I meant in the log messages - "web hook" -> "webhook"

cmd/git-sync/main.go
thockin: Yeah, I think we can't really make aany guarantees if a sync doesn't succeed, and if it did succeed, `--one-time` is the only really synchronous thing we support.

ChrisERo: Sounds good, so in the case of termination due to pulling a hash/specific-commit, we will allow for the hook race condition described in this issue to persist. 

In that case, will move checks to '--one-time' block and add the status-code change.

cmd/git-sync/main.go
ChrisERo: Sounds good. I also need to double space sentences instead of single spacing.

test_e2e.sh
ChrisERo: Hmm, this makes sense. My reason for running the same test several times was to be more sure of the results. Since race conditions may or may not be detected due to the OS' scheduling of process/threads, I thought we needed multiple test runs to be more sure of the results.

However, we could never really be 100% sure using this approach anyway, so I think we may be good with running each test exactly once, and maybe making the sleep last for a bit :)

ChrisERo: In this case, I don't think we need the helper functions really, so I'll go ahead and delete them.

test_e2e.sh
ChrisERo: You mean reducing total time test components spend sleeping correct? I need to sleep in order to add extra time for the hook to complete, to ensure that main thread is waiting for hook to complete before terminating.

test_e2e.sh
ChrisERo: N/A anymore since removing helper function since only running test once.

ChrisERo: N/A see below

ChrisERo:
My apologies, misread the k8s docs on good commit messages. Updated my first and second commit, since the second one referenced @thockin.

> Also your first commit has a pr ref in the comment - our bot doesn't want that :)

cmd/git-sync/main.go
```diff
@@ -359,6 +364,10 @@ func main() {
 		if *flExechookBackoff < time.Second {
 			handleError(log, true, "ERROR: --exechook-backoff must be at least 1s")
 		}
+
+		if *flOneTime {
+			exechookChannel = make(chan bool, 1)
```
thockin, line 369: What if this was a field of Webhook and maaybe a method like "WaitForCompletion"  ?  Then this all gets localized even further.  Better or worse in your opinion?

pkg/hook/hook.go
```diff
@@ -123,16 +128,32 @@ func (r *HookRunner) Run(ctx context.Context) {
 			if err := r.hook.Do(ctx, hash); err != nil {
 				r.logger.Error(err, "hook failed")
 				updateHookRunCountMetric(r.hook.Name(), "error")
+				// don't want to sleep unnecessarily terminating anyways
+				r.sendCompletedOnceMessageIfApplicable(false)
 				time.Sleep(r.backoff)
```
thockin, line 133: If we do send a notification, it means the parent will exit soon, so we shouldn't even bother to retry, or we end up with anaother race, right?  If we pass `oneTime bool` into the hook creation, then we know not to bother here, and we can be smart enough to return an error if you like my idea of a method above.  E.g.

```
if !r.oneTime {
    return fmt.Errorf("hook.WaitForCompletion called on async runner")
}

File: test_e2e.sh [hunk 12]
```diff
@@ -1072,6 +1074,66 @@ function e2e::exechook_fail_retry() {
...
+
+    GIT_SYNC \
+        --period=100ms \
+        --one-time \
+        --repo="file://$REPO" \
+        --branch="$MAIN_BRANCH" \
+        --root="$ROOT" \
+        --link="link" \
+        --exechook-command="$EXECHOOK_COMMAND_SLEEPY" \
+        >> "$1" 2>&1 &
```
thockin, line 1093: Don't make this a background run.  Most of the `--one-time` runs are foreground, so we can make assertions about what happened aftger it exited.  This also means your `sleep` below can be the normal 2

test_e2e.sh
```diff
@@ -1072,6 +1074,66 @@ function e2e::exechook_fail_retry() {
...
+        GIT_SYNC \
+            --period=100ms \
+            --one-time \
+            --repo="file://$REPO" \
+            --branch="$MAIN_BRANCH" \
+            --root="$ROOT" \
+            --link="link" \
+            --exechook-command="$EXECHOOK_COMMAND_FAIL_SLEEPY" \
+            --exechook-backoff=1s \
+            >> "$1" 2>&1 &
```
thockin, line 1125: same as above - not a BG run, sleep 2

test_e2e.sh
```diff
@@ -1072,6 +1074,66 @@ function e2e::exechook_fail_retry() {
...
+
+        # Check that exechook was called
+        # Original test sleeps for 2 seconds, so sleep for 2 seconds more than
+        # hook's sleep time
+        sleep 5
+        RUNS=$(cat "$RUNLOG" | wc -l)
```
thockin, line 1131: why would this be 2?

test_e2e.sh
```diff
@@ -1168,6 +1230,79 @@ function e2e::webhook_fail_retry() {
...
+    GIT_SYNC \
+        --period=100ms \
+        --one-time \
+        --repo="file://$REPO" \
+        --branch="$MAIN_BRANCH" \
+        --root="$ROOT" \
+        --webhook-url="http://$IP" \
+        --webhook-success-status=200 \
+        --link="link" \
+        >> "$1" 2>&1 &
```
thockin, line 1258: same comments as above

test_e2e.sh
```diff
@@ -1168,6 +1230,79 @@ function e2e::webhook_fail_retry() {
...
+
+    # Check that webhook was called
+    sleep 5 # preserve original diff between hook sleep and test sleep time
+    HITS=$(cat "$HITLOG" | wc -l)
```
thockin, line 1299: What is this proving?

thockin (commented):
Looking good!!

cmd/git-sync/main.go
ChrisERo: I like this localization suggestion, although I think I will implement this in the Hook**Runner** instead of Hook to minimize code changes, and because HookRunner already has the channel passed in.

test_e2e.sh
ChrisERo: Sounds like a good idea!

test_e2e.sh
ChrisERo: You are right, bumped this down to 1. My apologies, made these tests before being able to test them (was having issues setting up Fedora VM). So just copied as much from similar tests as possible.

ChrisERo: Have it running now and ran the tests.

test_e2e.sh
ChrisERo: This is proving that my solution works for both webhooks in the case where the hook fails.

ChrisERo:
wait before reviewing, one of the tests failed

thockin:
Thanks.  Rather than nitpick further, I'll merege this and follow-up.  I have soem other cleanups, and I'll need to port to v4 branch, anyway.

thockin:
Thanks!

/lgtm
/approve

PR #471, opened 2022-01-22
Fix e2e, dropped 'finish()' func

thockin:
Small SNAFU in last PR


thockin:
self-merge to get this out of the way, PTAL but it is trivial.  I just dropped this block previously.

Issue #470, opened 2022-01-21
clone into existing dir fails after cleanup


sed-i:
## Intro
I am using the `-root` and `-dest` args to have git-sync overwrite the folder without manual intervention.

## Environment

| Tool | Version |
|------|---------|
| git-sync     |    3.3.4     |

## Steps to reproduce
Start git-sync with -root and -dest args to point to an existing, non-empty folder.

## Expected result
Target folder is cleaned up and repo is cloned successfully.

## Actual result
Target folder is cleaned up but clone fails.

## Proof
When git-sync needs to cleanup, cloning fails:

```
2022-01-21T19:44:10.021Z [pebble] Service "git-sync" starting: /git-sync -repo https://github.com/canonical/grafana-k8s-operator.git -branch main -depth 1 -wait 60 -root /git -dest repo
2022-01-21T19:44:10.029Z [git-sync] I0121 19:44:10.029445      99 main.go:507] "level"=0 "msg"="starting up" "pid"=99 "args"=["/git-sync","-repo","https://github.com/canonical/grafana-k8s-operator.git","-branch","main","-depth","1","-wait","60","-root","/git","-dest","repo"]
2022-01-21T19:44:10.029Z [git-sync] I0121 19:44:10.029585      99 main.go:860] "level"=0 "msg"="cloning repo" "origin"="https://github.com/canonical/grafana-k8s-operator.git" "path"="/git"
2022-01-21T19:44:10.032Z [git-sync] I0121 19:44:10.032523      99 main.go:866] "level"=0 "msg"="git root exists and is not empty (previous crash?), cleaning up" "path"="/git"
2022-01-21T19:44:10.037Z [git-sync] E0121 19:44:10.037638      99 main.go:172] "msg"="too many failures, aborting" "error"="unlinkat //git: device or resource busy" "failCount"=0
```

When I restart it (after the cleanup), it works:

```
2022-01-21T19:46:13.666Z [pebble] Service "git-sync" starting: /git-sync -repo https://github.com/canonical/grafana-k8s-operator.git -branch main -depth 1 -wait 60 -root /git -dest repo
2022-01-21T19:46:13.672Z [git-sync] I0121 19:46:13.672542     119 main.go:507] "level"=0 "msg"="starting up" "pid"=119 "args"=["/git-sync","-repo","https://github.com/canonical/grafana-k8s-operator.git","-branch","main","-depth","1","-wait","60","-root","/git","-dest","repo"]
[...]

PR #472, opened 2022-01-22
Don't try to remove the root if it appears corrupt (v4)

thockin:
The `--root` is often a volume and can't be removed.  Instead, remove
the contents of it.  Adjust tests to hit this.

xref #470


stp-ip (approved):
/lgtm

PR #477, opened 2022-01-23
Allow --dest to be an absolute path (v3)

### LINKED ISSUE #428: Allow `--dest` to not be nested under `--root`
I have a use-case where I need to store the main `--dest` symlink in a folder which is not under `--root`, as the application consuming the repo will be confused by multiple folders.

For example, the folder structure might look like:
```
# GIT_SYNC_ROOT=/tmp/git
/tmp/git/SHAXXXX
/tmp/git/SHAXXXX

# GIT_SYNC_DEST= /home/repos/repo1
/home/repos/repo1
```

thockin:
This means it might live outside of --root.

Fixes #428


stp-ip (approved):
/lgtm

PR #478, opened 2022-01-23
Allow --dest to be an absolute path (v4)

thockin:
This means it might live outside of --root.

xref #428


stp-ip (approved):
/lgtm

First time I noticed due to the addition of spaces that was made as well.
Why do we use double spaces after periods? Or just a habit? Nothing to block for so merging.

thockin:
It was inconsistent in the file.  It used to be the norm to use 2 spaces
after periods, which is how I learned.  I don't care either way, just
wanted consistency. :)

PR #480, opened 2022-01-24
Fix bug caused in previous PR (v4)

thockin:
The .git to check is not the root/.git but the link/.git - tests pass now.


thockin:
Self-merge to fix v4 branch

stp-ip:
Sorry I missed that.

thockin:
It was my fault

PR #487, opened 2022-02-03
Implement previous hash argument for webhooks

darkvertex:
First time contributor here. Hello! 👋

I am interested in having access to the previous hash in the webhook functionality so that the receiver can be notified from what rev to what rev things just changed instead of simply sending the latest rev hash.

This change adds a new header (`Gitsync-Hash-Previous`) to the webhook's HTTP request call.

This makes it much simpler to comprehend the changes in an update as one can, for example, `git diff --name-only <oldhash> <hash>` to understand what files changed. I feel this is a valuable functionality to expose and simplifies webhook receivers from tracking the old hash themselves if they don't want to.

I thought a bit about the exechook hook and it didn't really seem to do anything with the hash argument so I didn't spend time adding any special behaviour for the old hash value beyond updating the function signature to satisfy the Hook.Do() interface.

---

As for the CNCF _Contributor License Agreement_, I signed it. I also updated the tests as best as I understood them.

Let me know if I missed anything and feel free to correct my style. (I enabled "_Allow edits by maintainers_".)

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -589,6 +589,7 @@ func main() {
 
 	initialSync := true
 	failCount := 0
+	oldHash := git.rev
 	for {
 		start := time.Now()
 		ctx, cancel := context.WithTimeout(context.Background(), *flSyncTimeout)
@@ -616,11 +617,12 @@ func main() {
 			continue
 		} else if changed {
 			if webhookRunner != nil {
-				webhookRunner.Send(hash)
+				webhookRunner.Send(hash, oldHash)
 			}
 			if exechookRunner != nil {
-				exechookRunner.Send(hash)
+				exechookRunner.Send(hash, oldHash)
 			}
+			oldHash = hash
 
 			updateSyncMetrics(metricKeySuccess, start)
 		} else {
--- a/pkg/hook/exechook.go
+++ b/pkg/hook/exechook.go
@@ -60,7 +60,7 @@ func (h *Exechook) Name() string {
 }
 
 // Do runs exechook.command, implements Hook.Do
-func (h *Exechook) Do(ctx context.Context, hash string) error {
+func (h *Exechook) Do(ctx context.Context, hash string, oldHash string) error {
 	ctx, cancel := context.WithTimeout(ctx, h.timeout)
 	defer cancel()
 
--- a/pkg/hook/exechook_test.go
+++ b/pkg/hook/exechook_test.go
@@ -36,7 +36,7 @@ func TestNotZeroReturnExechookDo(t *testing.T) {
 			time.Second,
 			l,
 		)
-		err := ch.Do(context.Background(), "")
+		err := ch.Do(context.Background(), "", "")
 		if err == nil {
 			t.Fatalf("expected error but got none")
 		}
@@ -54,7 +54,7 @@ func TestZeroReturnExechookDo(t *testing.T) {
 			time.Second,
 			l,
 		)
-		err := ch.Do(context.Background(), "")
+		err := ch.Do(context.Background(), "", "")
 		if err != nil {
 			t.Fatalf("expected nil but got err")
 		}
@@ -72,7 +72,7 @@ func TestTimeoutExechookDo(t *testing.T) {
 			time.Second,
 			l,
 		)
-		err := ch.Do(context.Background(), "")
+		err := ch.Do(context.Background(), "", "")
 		if err == nil {
 			t.Fatalf("expected err but got nil")
 		}
--- a/pkg/hook/hook.go
+++ b/pkg/hook/hook.go
@@ -43,13 +43,14 @@ type Hook interface {
 	// Describes hook
 	Name() string
 	// Function that called by HookRunner
-	Do(ctx context.Context, hash string) error
+	Do(ctx context.Context, hash string, oldHash string) error
 }
 
 type hookData struct {
-	ch    chan struct{}
-	mutex sync.Mutex
-	hash  string
+	ch      chan struct{}
+	mutex   sync.Mutex
+	hash    string
+	oldHash string
 }
 
 // NewHookData returns a new HookData
@@ -63,20 +64,21 @@ func (d *hookData) events() chan struct{} {
 	return d.ch
 }
 
-func (d *hookData) get() string {
+func (d *hookData) get() (string, string) {
 	d.mutex.Lock()
 	defer d.mutex.Unlock()
-	return d.hash
+	return d.hash, d.oldHash
 }
 
-func (d *hookData) set(newHash string) {
+func (d *hookData) set(newHash string, oldHash string) {
 	d.mutex.Lock()
 	defer d.mutex.Unlock()
 	d.hash = newHash
+	d.oldHash = oldHash
 }
 
-func (d *hookData) send(newHash string) {
-	d.set(newHash)
+func (d *hookData) send(newHash string, oldHash string) {
+	d.set(newHash, oldHash)
 
 	// Non-blocking write.  If the channel is full, the consumer will see the
 	// newest value.  If the channel was not full, the consumer will get another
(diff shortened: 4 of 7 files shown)


darkvertex:
So I think the EasyCLA is yelling at me because my commit is using my obfuscated email because I have configured Github to prevent commits from revealing my private email address. If I git push with my real email I get:
```
remote: error: GH007: Your push would publish a private email address.
```

Do I absolutely have to disable this security feature in my Github profile to go on? Or is there an override I can do somehow to prove I'm me? I already signed the agreement and linked it to my Github user.

stp-ip:
> So I think the EasyCLA bot is yelling at me because my commit is using my obfuscated email address (`941331+darkvertex@users.noreply.github.com`) because I have configured Github to prevent commits from revealing my private email address. If I git push with my real email I get:
> 
> ```
> remote: error: GH007: Your push would publish a private email address.
> ```
> 
> Do I _absolutely_ have to disable this security feature in my Github profile to go on? Or is there an override I or somebody can do to prove I'm me? I already signed the agreement and linked it to my Github user.

Can you sign the CLA with the obfuscated email by chance? Might be the easiest solution I guess.

cmd/git-sync/main.go
```diff
@@ -600,7 +601,8 @@ func main() {
 			}
 		}
 
-		if changed, hash, err := git.SyncRepo(ctx); err != nil {
+		changed, hash, err := git.SyncRepo(ctx)
+		if err != nil {
```
thockin, line 605: Why this stylistic change?

pkg/hook/exechook.go
```diff
@@ -59,7 +59,7 @@ func (h *Exechook) Name() string {
 }
 
 // Do runs exechook.command, implements Hook.Do
-func (h *Exechook) Do(ctx context.Context, hash string) error {
```
thockin, line 62: Huh, I could swear we were setting the has in an env var.  I can add that later

thockin (commented):
Hi, I appreciate the PR!  But I think there's a fatal flaw in this.  Hooks are not guaranteed to happen on every sync.  Hooks are async to the pulls.  Consider this case:

Synced at SHA "abc123"
Remote advances to SHA "def456"
We sync to "def456"
Queue webhook old=abc123, new=def456
Update oldHash to "def456"
Webhook hits an error, retry in 2 seconds
Remote advances to "aaa777"
We sync to aaa777
Queue webhook old=sef456, new=aaa777
2 seconds sleep finishes
Webhook retries with old=sef456, new=aaa777

Your hook receiver is relying unreliable information.  There are other corner cases, like sync, crash, restart which I think can cause the same failure.  Only you know what SHA you succeed in receiving last, unless we want to start tracking that in git-sync.

darkvertex:
> Hi, I appreciate the PR!  But I think there's a fatal flaw in this.  Hooks are not guaranteed to happen on every sync.  Hooks are async to the pulls.  Consider this case:
> 
> Synced at SHA "abc123"
> Remote advances to SHA "def456"
> We sync to "def456"
> Queue webhook old=abc123, new=def456
> Update oldHash to "def456"
> Webhook hits an error, retry in 2 seconds
> Remote advances to "aaa777"
> We sync to aaa777
> Queue webhook old=sef456, new=aaa777
> 2 seconds sleep finishes
> Webhook retries with old=sef456, new=aaa777
> 
> Your hook receiver is relying unreliable information.

You raise a good point.

Do you think the info would be more helpful in such delayed cases if the webhook headers also had a header for when the operation occurred? Maybe then it would be easier to spot stale queue retries?

thockin:
I think it comes down to "what question are you trying to answer"?  If the
question is "what changed since my app last checked", you need to either
track that yourself or else make git-sync checkpoint successful hook
deliveries.  I don't think we can do that latter because git-sync only
knows the hook was delivered, not acted upon.

Is there a different question that something like a timestamp would answer?

darkvertex:
For myself, I wish to answer "_what changed since the last git-sync sync upgraded the repo on my server?_"

With the timestamp header I could answer "_was this webhook sync ping that just came in done in the last 30m or was my webhook receiver infra down for the whole weekend and this was from an old queue that just cleared?_"

thockin:
> I wish to answer "what changed since the last git-sync sync upgraded the repo on my server?"

I can answer that, but I don't think it's actually what you want to know :)  I think that what you really want is "what changed since the last time I successfully acknowledged the git state".  The subtle difference is that it's not since git-sync did something but since you did something.  Those are asynchronous things.

Now, I don't know anything about your repo - maybe the corner case I described above doesn't happen in your setup.  But git-sync can't make that assumption, so "since the last git-sync sync" is a pretty low-value promise to make.

> With the timestamp header I could answer "was this webhook sync ping that just came in done in the last 30m or was my webhook receiver infra down for the whole weekend and this was from an old queue that just cleared?"

I'm not against a timestsamp, per se, but we'd need to be REALLY clear on which timestamp it would be.  It's not "the first time the repo changed since <something>".  It's "the most recent time the repo changed".  And in fact, that could be "the most recent time git-sync restarted" because, if git-sync crashes and restarts, I don't exactly know when the last update came in, I only know what I have NOW.

Consider again a timeline:

```
t0:  Synced at SHA "abc123"
t1:  Remote advances to SHA "def456"
t2:  Poll loop sees "def456"
t3:  Sync to "def456"
t4:  Queue webhook hash=def456, timestamp=t3
t5:  Webhook hits an error, retry in 2 seconds
t6:  Remote advances to "aaa777"
t7:  Poll loop sees "aaa777"
t8:  Sync to "aaa777"
t9:  Queue webhook hash=aaa777 timestamp=t8
t10: 2 second sleep finishes
t11: Webhook retries with new=aaa777 timestamp=t8
```

Does that really tell you what you wanted to know?

How About this:

```
t0:  Synced at SHA "abc123"
t1:  Remote advances to SHA "def456"
t2:  Poll loop sees "def456"
t3:  Sync to "def456"
t4:  git-sync crashes
t5:  git-sync restarts
t6:  Observe current SHA is "def456"
t7:  Queue webhook hash=def456, timestamp=t6
```

2 things to note:  First, I think we don't actually send a webhook at all in
this case, but probably should.  Second, the delay between t4 and t5 is
arbitrary.  System errors, disk IO problems, out-of-memory, node failure, etc.
could delay that by minutes.  The timestamp t6 isn't really meaningful, as far
as I can see.

cmd/git-sync/main.go
darkvertex: It's a mistake, sorry. I'll change it back.

I thought "hash" wouldn't be available in the "else if" scope if it was declared inline in the "if" but I now see that "changed" was initialized/available just fine, so I think I may not have been _sufficiently caffeinated_ to see it properly. I'll change it back.

I'm in the habit that in Golang when you make a var inside an "if" or a "for" it is scoped to that block and I mistook the inline "if" var init as being scoped inside its "if " block when apparently it's scoped just outside of it. My bad.

pkg/hook/exechook.go
thockin: I'm still not sure we should do this, but if you rebase you will now find that we set $GITSYNC_HASH :)

pkg/hook/exechook_test.go
```diff
@@ -27,7 +27,7 @@ import (
 
 func TestNotZeroReturnExechookDo(t *testing.T) {
 	t.Run("test not zero return code", func(t *testing.T) {
-		l := logging.New("", "")
+		l := logging.New("", "", 0)
```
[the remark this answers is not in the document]
thockin: Patched into master, thanks for flagging.  If you rebase they should cancel-out.

pkg/hook/exechook.go
thockin: So *if* we were to proceed with this proposal and add a header in HTTP, wew should add the same in exec.

PR #489, opened 2022-02-07
Fix tests to set missing arg (v4 only)

thockin:
Pointed out by @darkvertex


thockin:
self-merge, broken test

PR #493, opened 2022-02-12
Set repoReady even when there wasn't a 1st clone (v3 branch)

### LINKED ISSUE #488: readinessProbe/livenessProbe (k8s) fails if the pod has restarted and there is no change in git repo
**Scenario:**
The pod is up and a successful clone to git repo has been performed. 
At some point the pod restarts and there is no change in the git repo, in this case the git-sync is not ready and the whole pod is considered not ready (especially in the situation that it runs as a sidecar and causes the whole pod to be not ready).

**The problem is in the next section of the code:**
The call to **setRepoReady** is performed in the following function -**addWorktreeAndSwap**. 
The call to the **addWorktreeAndSwap** function is performed in the **syncRepo** function. 
But we do not call **addWorktreeAndSwap** if the following condition is met, and immediately exit the function:

```
	default:
		// Not the first time. Figure out if the ref has changed.
		local, remote, err := getRevs(ctx, dest, branch, rev)
		if err != nil {
			return false, "", err
		}
		if local == remote {
			log.V(1).Info("no update required", "rev", rev, "local", local, "remote", remote)
			return false, "", nil
		}
		log.V(0).Info("update required", "rev", rev, "local", local, "remote", remote)
		hash = remote
```

The problem with this condition: **local == remote**. 
If this condition is met then it should be announced that the repo is also ready, because otherwise it will not be ready until the next clone...

thockin:
E.g. if the repo is already present (after a restart).

Fixes #488

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -513,16 +513,20 @@ func main() {
 			cancel()
 			time.Sleep(waitTime(*flWait))
 			continue
-		} else if changed {
-			if webhookRunner != nil {
-				webhookRunner.Send(hash)
-			}
-			if exechookRunner != nil {
-				exechookRunner.Send(hash)
-			}
-			updateSyncMetrics(metricKeySuccess, start)
 		} else {
-			updateSyncMetrics(metricKeyNoOp, start)
+			// this might have been called before, but also might not have
+			setRepoReady()
+			if changed {
+				if webhookRunner != nil {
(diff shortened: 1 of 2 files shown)


thockin:
Hold while I make sure tests are good and make a v4 equivalent.

thockin:
hold removed

stp-ip (approved):
/lgtm

Issue #439, opened 2021-08-09
Error on git-sync running as a CronJon


michal-jagiello-tmpl:
Hi,
I'm using git-sync `v3.3.4` in a CrobJob as an initContainer. It's a definition:
```
- name: clone-results-repo
  image: "{{ .Values.init.cloneRepo.image.repository }}:{{ .Values.init.cloneRepo.image.tag }}"
  imagePullPolicy: {{ .Values.init.cloneRepo.image.pullPolicy }}
  volumeMounts:
    - name: persistent-storage
      mountPath: /git
  env:
  - name: GIT_SYNC_REPO
    value: {{ .Values.git.url }}
  - name: GIT_SYNC_ONE_TIME
    value: "true"
  - name: GIT_SYNC_BRANCH
    value: my_awesome_branch
  - name: GIT_SYNC_DEPTH
    value: "1"
  - name: GIT_SYNC_USERNAME
    valueFrom:
      secretKeyRef:
        name: {{ include "my-awesome-app.fullname" . }}-git-credentials
        key: GIT_PULL_USERNAME
  - name: GIT_SYNC_PASSWORD
    valueFrom:
      secretKeyRef:
        name: {{ include "my-awesome-app.fullname" . }}-git-credentials
        key: GIT_PULL_PASSWORD
  - name: GIT_SYNC_ROOT
    value: /git
  - name: GIT_SYNC_TIMEOUT
    value: "99999"
```
There is another container which also mounts `persistent-storage` volume and consumes the data from the cloned repo. The issue is that after few succeeded executions I have always the same error:
```
INFO: detected pid 1, running init handler
I0809 06:29:23.648815      11 main.go:507] "level"=0 "msg"="starting up" "pid"=11 "args"=["/git-sync"]
I0809 06:29:24.077876      11 main.go:1003] "level"=0 "msg"="update required" "rev"="HEAD" "local"="1172cc4eeed3a3dd6d5e8fb65f3c15134adf9f32" "remote"="bfa07ea5354c25fa7e267dbcb6bbb305f2bd315f"
I0809 06:29:24.077969      11 main.go:690] "level"=0 "msg"="syncing git" "rev"="HEAD" "hash"="bfa07ea5354c25fa7e267dbcb6bbb305f2bd315f"
[...]

PR #495, opened 2022-02-21
Add GC controls, e2e regexes (v3 branch)

thockin:
Looking for feedback, re: #439.  This does not solve the timeout problem, and it does not do ideal repo-packing, but that is bigger and will have to come in v4 only.

Set `gc.autoDetach` to "false"

Add `--git-gc` flag to control GC on each sync
Values:
* "auto" - run `git gc --auto` (default, respects git gc.* configs)
* "always" - run `git gc`
* "aggressive" - run `git gc --aggressive` (may require a longer timeout)
* "off" - do not run `git gc` on each sync (good for --one-time use)

Also makes e2e test treat args as regexes.

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -408,6 +408,12 @@ func main() {
 		askpassCount.WithLabelValues(metricKeySuccess).Inc()
 	}
 
+	// Set additional configs we want, but users might override.
+	if err := setupDefaultGitConfigs(ctx); err != nil {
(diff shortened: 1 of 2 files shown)


thockin:
OK, I think this is ready for review.

thockin:
ping on this if anyone cares

stp-ip (approved):
/lgtm

stp-ip:
Might wanna update the Readme with the latest gc arg.

PR #502, opened 2022-03-11
update the base image to fix vulnerabilities

Liujingfang1:
It fixes the critical vulnerabilities: CVE-2021-45960

PR #503, opened 2022-03-11
update the base image to fix vulnerabilities (v4)

thockin:
It fixes the critical vulnerabilities: CVE-2021-45960

From Liujingfang1

Patch of #502


thockin:
Self-merging to prep a v3 release (don't want v4 to be out of sync)

PR #504, opened 2022-03-11
Clean up recently added test case

thockin:
Just a rename and reorder.


thockin:
Slef-merge to cut a release

PR #514, opened 2022-03-17
Use logger.WithName() for web/exechook (v4)

thockin:
This makes logs easier to comprehend, since hooks are async.

This required an intermediate interface.


stp-ip (approved):
/lgtm

PR #515, opened 2022-03-17
Pass the whole environment to exechooks (v3 branch)

### LINKED ISSUE #508: Slack incoming webhooks payload issue
Slack disabled the url of an incoming webhook because it was publicly exposed in github. The project is private but anyways they detected it and disabled the webhook and send us an email.

The issue is related with the already solved [issue here](https://github.com/kubernetes/git-sync/issues/492).

As `GIT_SYNC_EXECHOOK_COMMAND` doesn't allow parameters I can't setup the incoming webhook url as a parameter, so I have to harcode it in the sh file that I created.

Something like:

`#!/bin/sh
SHA=$(git rev-parse HEAD)
curl -X POST -H 'Content-type: application/json' --data "{\"text\":\"Airflow code was updated to git commit $SHA\"}" $1`

where `$1` has to be changed to the url.

I tried setting up: `GIT_SYNC_EXECHOOK_COMMAND=/tmp/git/airflow/git-sync-webhook.sh https://hooks.slack.com/services/XX/XX/XX`.
But the log returns:

`I0317 12:12:10.307832      12 exechook.go:68] "level"=0 "msg"="running exechook" "command"="/tmp/git/airflow/git-sync-webhook.sh" "timeout"=30000000000`
`E0317 12:12:10.326114      12 hook.go:136] "msg"="hook failed" "error"="Run(/tmp/git/airflow/git-sync-webhook.sh ): exit status 2: { stdout: "", stderr: "curl: no URL specified!\ncurl: try 'curl --help' or 'curl --manual' for more information\n" }"`

So the url is not received in the git-sync code and then we are forced to harcode it, running the risk that in sometime the webhook gets disabled again by slack.

Any way to solve that that I didn't think about it?

thockin:
Rather than wiping it all when installing GITSYNC_HASH, retain it and add GITSYNC_HASH.

Fixes #508

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -75,6 +75,9 @@ var flMaxSyncFailures = flag.Int("max-sync-failures", envInt("GIT_SYNC_MAX_SYNC_
 	"the number of consecutive failures allowed before aborting (the first sync must succeed, -1 will retry forever after the initial sync)")
 var flChmod = flag.Int("change-permissions", envInt("GIT_SYNC_PERMISSIONS", 0),
 	"the file permissions to apply to the checked-out files (0 will not change permissions at all)")
+var flSparseCheckoutFile = flag.String("sparse-checkout-file", envString("GIT_SYNC_SPARSE_CHECKOUT_FILE", ""),
+	"the path to a sparse-checkout file.")
+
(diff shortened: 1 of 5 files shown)


stp-ip (approved):
/lgtm

PR #521, opened 2022-05-25
Pass the whole environment to exechooks (v4 branch)

thockin:
Rather than wiping it all when installing GITSYNC_HASH, retain it and add GITSYNC_HASH.


stp-ip (approved):
/lgtm

PR #525, opened 2022-06-05
Makefile .licenses target should depend on correct architecture

### LINKED ISSUE #524: Make Error with Cross Arch Build
Trying to building an arm64 container on amd64 platform. Specifically I am building on Windows 11 WSL2 running `Ubunbtu 22.04 LTS`.

```
> go version
go version go1.18.1 linux/amd64
```

Running command:
```
make container REGISTRY=gazzyt VERSION=3.5.0 GOOS=linux GOARCH=arm64
```

Produces error:
```
making bin/linux_arm64/git-sync
make: *** No rule to make target 'bin/linux_amd64/git-sync', needed by '.licenses'.  Stop.
```

The arm64 binary is built correctly at `bin/linux_arm64/git-sync` but the `.licenses` target has dependency on the local arch variant rather than the built one.

gazzyt:
Fix for #524 ensuring that the dependency of target `.licenses` is to the correct architecture version of the build `git-sync` executable rather than the local platform architecture

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -122,7 +122,7 @@ DOTFILE_IMAGE = $(subst /,_,$(IMAGE))-$(TAG)
 
 LICENSES = .licenses
 
-$(LICENSES): bin/$(shell go env GOOS)_$(shell go env GOARCH)/$(BIN)
+$(LICENSES): bin/$(OS)_$(ARCH)/$(BIN)
 	@go build -o ./bin/tools github.com/google/go-licenses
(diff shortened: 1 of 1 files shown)


gazzyt:
I fixed the commit message but the `do-not-merge/invalid-commit-message` label remains

thockin:
#527 is a more correct fix, I think.  No need for this dep at all.

PR #527, opened 2022-06-06
Fix Makefile to run tools on host OS/ARCH

### LINKED ISSUE #524: Make Error with Cross Arch Build

thockin:
Fixes #524

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,23 +94,23 @@ $(OUTBIN): .go/$(OUTBIN).stamp
 .PHONY: .go/$(OUTBIN).stamp
 .go/$(OUTBIN).stamp: $(BUILD_DIRS)
 	@echo "making $(OUTBIN)"
-	@docker run                                                                  \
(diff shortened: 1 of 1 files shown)


stp-ip (approved):
/lgtm

PR #534, opened 2022-06-17
upgrade base image to address vulnerabilities

sdowell:


thockin (commented):
Thanks!

Can you also remove the "buster" backports lines from Dockerfile.in - they should not be needed any more.

thockin:
Thanks!

/lgtm
/approve

PR #538, opened 2022-06-24
v4: e2e: Don't manually "docker kill"

thockin:
It happens automatically on test cleanup

### DIFF
--- a/test_e2e.sh
+++ b/test_e2e.sh
@@ -1066,12 +1066,12 @@ function e2e::askpass_url() {
         --root="$ROOT" \
         --link="link" \
         >> "$1" 2>&1 || true
-    docker_kill "$CTR"
 
     # check for failure
     assert_file_absent "$ROOT"/link/file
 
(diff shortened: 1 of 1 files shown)


test_e2e.sh
```diff
@@ -1066,12 +1066,12 @@ function e2e::askpass_url() {
         --root="$ROOT" \
         --link="link" \
         >> "$1" 2>&1 || true
-    docker_kill "$CTR"
 
     # check for failure
     assert_file_absent "$ROOT"/link/file
 
     # run with askpass_url service with correct password
+    docker_kill "$CTR"
```
stp-ip, line 1074: Why the manual kill here? To force a kill before CTR gets rewritten?

thockin: correct - CTR is a helped and I need to kill and restart it here

thockin:
Rebased

stp-ip (approved):
/lgtm

PR #577, opened 2022-07-03
v3: Clean up credential init 

thockin:
Dep #571


thockin:
Gentle ping - this is next in the v3 series

PR #608, opened 2022-08-13
Move tools into a tools/ subdir

### LINKED ISSUE #593: Move tools to a sub-module

thockin:
This avoids vendoring all the deps in the root.

Fixes #593


stp-ip (approved):
I assume not vendoring tools was deliberate. Agreed.
/lgtm

PR #610, opened 2022-08-14
Better passwd redacting - including URLs

### LINKED ISSUE #602: Using HTTPS with an embedded username and PAT causes it to print in logs
Using a repository URL without a PAT and username does not work.
`E0809 16:38:35.073853      12 main.go:539] "msg"="error syncing repo, will retry" "error"="Run(git clone -v --no-checkout -b main --depth 1 https://github.com/MyRepo/manifests.git /var/git-sync): exit status 128: { stdout: "", stderr: "Cloning into '/var/git-sync'...\nremote: Support for password authentication was removed on August 13, 2021.\nremote: Please see https://docs.github.com/en/get-started/getting-started-with-git/about-remote-repositories#cloning-with-https-urls for information on currently recommended modes of authentication.\nfatal: Authentication failed for 'https://github.com/MyRepo/manifests.git/'" }" "failCount"=1 "waitTime"=60000000000`

Alternatively, using a repository URL with a PAT and username does work, but it causes the PAT to be printed in the kubectl logs (as part of the url). This seems like a security issue... Is there a way to fix this?

Which one is intended? Am I missing something?

Thanks in advance.

thockin:
Fixes #602

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -27,6 +27,7 @@ import (
 	"net"
 	"net/http"
 	"net/http/pprof"
+	"net/url"
 	"os"
 	"os/exec"
 	"os/signal"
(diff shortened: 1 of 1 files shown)


stp-ip (approved):
/lgtm

PR #612, opened 2022-08-18
Log before running command

thockin:
log before running the command - this seems like a mistake made in cut-paste.


stp-ip (approved):
/lgtm

PR #613, opened 2022-08-18
e2e: Allow N tests to fail and catch errexit

### LINKED ISSUE #611: e2e: `errexit` does not work in a function if called in `if`
All of the e2e functions are called under `if` and bash says that ignores errexit.  Need to find a way to convert that to `return 1` or something

thockin:
Prior to this, `errexit` was not active inside test functions (see
comments in this commit).

Now, multiple tests can fail - it does not abort on the first one.

Fixes #611

### DIFF
--- a/test_e2e.sh
+++ b/test_e2e.sh
@@ -19,14 +19,12 @@ set -o nounset
 set -o pipefail
 
 function fail() {
-    echo "FAIL: " "$@"
-    remove_containers || true
-    exit 1
+    echo "FAIL:" "$@"
+    return 42
 }
 
 function pass() {
     echo "PASS"
-    remove_containers || true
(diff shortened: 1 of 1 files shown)


stp-ip:
Got a few tests not passing with this. Haven't been able to pinpoint and worked through the other PRs for now.

thockin:
Interesting.  I'll re-run locally.  It passed for me before.  I have another few diffs which will make it a bit more deterministic

thockin:
I found a few issues that this exposed - runs that we expect to fail were causing failures, and some non-compliant fake-HTTP responses.

PTAL

stp-ip (approved):
/lgtm

PR #615, opened 2022-08-20
Support shallow sync when the rev is not in-range

### LINKED ISSUE #601: Download /  Pull a specific tag with git-sync
Hi:
I am trying to set up git-sync to pull a specific tag from a git repository, but it is not working. git-sync keeps pulling the latest commit instead of the files from the specific tag.

I am using the following env variables setup for the git-sync:

```
GIT_SYNC_REV=test
GIT_SYNC_REPO=<repository link>
GIT_SYNC_DEPTH=1
GIT_SYNC_ROOT=/git
GIT_SYNC_DEST=dags
GIT_SYNC_GIT_CONFIG=http.sslVerify:false
GIT_SYNC_WAIT=600
```

Can someone help me to figure out what is wrong with my setup? Does someone have a sample about how I should set up git-sync to pull a specific tag from a git repository?

Thank you

thockin:
If I ask for `--depth 1` and a branch, it's fine.

If I ask for `--depth 1` and a tag, and that tag is not within 1 commit
of the branch that was cloned, it will give an error.  Oddly, if the
initial `clone` was OK, and subsequent syncs drift, it is OK, because of
how we `fetch`.  But if it is too far away at the beginning, kaboom.

This betrays that the current model of `--branch` and `--rev` is really
broken, and should be revamped.  For now, I did something simple - if
the rev can't be found, try a fetch.  A "real" fix is more involved.

Also add tests.

Fixes #601

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -1345,7 +1345,7 @@ func (git *repoSync) SyncRepo(ctx context.Context, refreshCreds func(context.Con
 		if err != nil {
 			return false, "", err
 		}
-		hash, err = git.LocalHashForRev(ctx, git.rev)
+		hash, err = git.ensureLocalHashForRev(ctx, git.rev)
 		if err != nil {
 			return false, "", err
 		}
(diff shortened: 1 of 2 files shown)


stp-ip (approved):
/lgtm

PR #616, opened 2022-08-20
Retain user-provided GIT_SSH_COMMAND

### LINKED ISSUE #107: GIT_SSH_COMMAND is overwritten

thockin:
Fixes #107


stp-ip (approved):
/lgtm

PR #620, opened 2022-08-21
--max-sync-failures -> --max-failures

### LINKED ISSUE #617: Simplify flag: max-sync-failures -> max-failures ?

thockin:
Deprecate but retain the old flag and env.  This is in v4 only.

Fixes #617


stp-ip (approved):
/lgtm

PR #621, opened 2022-08-21
Remove now-obsolete flag from struct

thockin:
This was changed to the more general `exechook`, but must have been
missed.


stp-ip (approved):
/lgtm

PR #631, opened 2022-09-05
Change the sync loop to do a full sync

### LINKED ISSUE #622: Simplify initial-sync vs the rest
The main loop currently distinguishes the "initial" sync from all subsequent.  It makes the code more complex and doesn't seem super useful.  I'd like to try to make it just "while true { sync-repo }".  This would change some semantics like max-failures, but I think it would be for the better.

thockin:
No longer distinguish the first sync from the rest. This will make later changes easier.

Fixes #622

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -667,18 +667,15 @@ func main() {
 		return nil
 	}
 
-	initialSync := true
 	failCount := 0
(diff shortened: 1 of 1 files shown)


thockin:
ping?

stp-ip (approved):
/lgtm

PR #632, opened 2022-09-05
Make --error-file allow abs paths

### LINKED ISSUE #624: Some file args are relative, some are abs
They should probably all work the same way

thockin:
This also enforces th previously unenforced "must not start with a period" rule.

Fixes #624


thockin:
ping?

stp-ip (approved):
/lgtm

PR #639, opened 2022-10-31
Add --touch-file flag and use it in tests

thockin:
This is mostly designed to make testing easier instead of sleep.


stp-ip (approved):
/lgtm

Issue #641, opened 2022-11-15
Error during the pass GIT_SYNC_GIT_CONFIG with colon(:) inside key


Rattnik:
Hi, I'm trying to configure git to replace ssh to https by
`git config --global  'url.https://github.com/.insteadOf' 'git@github.com:'`

I pass next value to ENV git-sync container via Helm:
```
env:
    - name: GIT_SYNC_GIT_CONFIG
      value: "'url.https://github.com/.insteadOf':'git@github.com:'"
```

but I got next error inside container

`ERROR: can't set additional git configs: error configuring additional git configs "'url.https" "//github.com/.insteadOf':'git@github.com:'": Run(git config --global 'url.https //github.com/.insteadOf':'git@github.com:'): exit status 1: { stdout: "", stderr: "error: invalid key: 'url.https" `

My ENV from container
```
$ env
GIT_SYNC_GIT_CONFIG='url.https://github.com/.insteadOf':'git@github.com:'
```

Your parser view first colon and think that is value.

thockin:
You are right.  That's sloppy of me.  I'll have to think about how to fix
this in a compatible way.  I don't think I can literally accept `"url.
***@***.***"` because `:` is defined as
the separator already.

Options off the top of my head:
1) ***@***.***"`  # escape the ':'
- lowest impact but least obvious
2) ***@***.***"  # respect ':'
in quoted interior strings - breaks the meaning of interior quotes
3) use a different flag which has cleaner semantics for parsing - most
complete, most impact on users

I will have to think about this one...

thockin:
Fixed in v3.6.2

PR #643, opened 2022-11-20
V4: Allow quoted keys for --git-config

thockin:
This allows keys to contain literal ':' which would previously confuse the parser.

xref #641


stp-ip (approved):
/lgtm

Issue #646, opened 2022-11-24
git broke submodules with file://


thockin:
https://bugs.launchpad.net/ubuntu/+source/git/+bug/1993586

Our e2e needs this

PR #648, opened 2022-11-24
V3 e2e: fix git submodules for file://

thockin:
I guess upstream git intentionally broke this because of a CVE.

Also fix tag tests to pass -m

xref #646


stp-ip (approved):
/lgtm

PR #653, opened 2022-12-22
Update Dockerfile docs and default for volumes

thockin:
Mostly this updates docs around volumem mounting and permissions, and changes the default root back to /git (which was created for this).

e2e passes


stp-ip (approved):
/lgtm

PR #652, opened 2022-12-22
Fix bug when using GIT_SYNC_REV, and GIT_SYNC_ONE_TIME

yoyehan:
- When using git-sync to sync specific commits, git-sync fails after syncing the first time.
- Error occurs when different commit hashes (not tags) are inputed to GIT_SYNC_REV, and GIT_SYNC_ONE_TIME = true is used.
- Using a commit hash in rev config fails since GetRevs function doesn't account for the possibility that git.rev can be a commit hash. 
- I added a simple function to check if the given rev is a hash (and not a tag) that exists in the remote repository, and returns a HEAD hash for local, and a full given commit for remote to ensure that sync does happen.

PR #656, opened 2022-12-25
Fix cases of syncing different SHAs back to back

thockin:
Prior to this, it would fail if the 2nd SHA wasn't in the local repo. Now it doesn't care what the local SHA for rev is, it only cares what is checked out at HEAD.

Based on #652 from @yoyehan

I really need to finish my fetch overhaul


stp-ip (approved):
/lgtm

Issue #662, opened 2023-01-25
git-sync fails when `--branch` is not specified and repo has no 'master' branch


ForeverWintr:
Hello, 

The manual entry for the `--branch` argument says:

> --branch <string>, $GIT_SYNC_BRANCH
            The git branch to check out.  If not specified, this defaults to
            the default branch of --repo.

However, when I omit that argument I get a failure when my repo does not use 'master' as its main branch. This occurs even when I include `--rev`, which I imagine should remove the need to specify a branch too. Is this a bug, a documentation error, or am I misunderstanding how it should work?

Here's an example of my usage, using the python API:

```python
from kubernetes import client, config

# Setup:
config.load_kube_config()
repo_holder = client.V1Volume(
    name="repo-holder", empty_dir=client.V1EmptyDirVolumeSource()
)
repo_holder_mount = client.V1VolumeMount(
    mount_path="/repo_holder", name=repo_holder.name
)
git_secret = client.V1Volume(
    name="git-secret",
    secret=client.V1SecretVolumeSource(default_mode=256, secret_name="git-creds"),
)
git_secret_mount = client.V1VolumeMount(
    mount_path="/etc/git-secret", name=git_secret.name
)

# Container definition
gitsync = client.V1Container(
    name="gitsync",
    image="registry.k8s.io/git-sync/git-sync:v3.6.3",
    args=[
        "--one-time",
        "--repo=https://github.com/ForeverWintr/docker-example.git",
        "--root=/repo_holder",
    ],
    volume_mounts=[repo_holder_mount, git_secret_mount],
    security_context=client.V1SecurityContext(run_as_user=65533),
)

# Kubernetes Job
template = client.V1PodTemplateSpec(
    metadata=client.V1ObjectMeta(name="container-creator-pod"),
    spec=client.V1PodSpec(
        restart_policy="Never",
        containers=[gitsync],
        volumes=[repo_holder, git_secret, ecr_push_secret],
        security_context=client.V1PodSecurityContext(fs_group=65533),
    ),
)
spec = client.V1JobSpec(
    template=template,
    backoff_limit=1,
    active_deadline_seconds=3000,
    ttl_seconds_after_finished=300,
)
job = client.V1Job(
    api_version="batch/v1",
    kind="Job",
    metadata=client.V1ObjectMeta(name="image-maker"),
    spec=spec,
)

api = client.BatchV1Api()
api.create_namespaced_job(body=job, namespace="my-namespace")
```

[...]

PR #663, opened 2023-01-25
Update README on master to point to v3

thockin:
Fixes https://github.com/kubernetes/git-sync/issues/662


thockin:
self-merge, docs only

Issue #660, opened 2023-01-22
Allow sync wait time to be truncated so that a sync occurs immediately.


trulede:
We would like to trigger a sync immediately, either from our CI systems pushing a change out, or interactively while debugging. Currently when we use git-sync as a sidecar container we can do this by deleting the Pod, after which the deployment recreates the Pod (and the sync occurs) - but it is a bit messy.

The most recent use case I've had for this has been to push Hugo based documentation to an NGINX container. In this case, git-sync is used to pull the Hugo publish branch from the doc repo to a volume on a Pod, from which the NGINX container serves static content. The wait duration we use is 3600s, but could easily be set higher since normally the content of the published branch does not change too often.

I've made a simple prototype which truncates the wait time using a signal (os.Interrupt). If you like I could open a PR with that change and develop it further.

Related issues: #226 #482

thockin:
I'm not opposed to either signal or HTTP, it just needs some design work.  Those obviously solve for different things - signals require a local agent, HTTP can be over network but requres some amount of trust.

trulede:
Right. Using a signal will require a sidecar no matter what (technique described here: [Share Process Namespace](https://kubernetes.io/docs/tasks/configure-pod-container/share-process-namespace/#configure-a-pod)). Otherwise it's fairly straight forward.

With HTTP I guess it's a bit tricky. Use the existing endpoint (and trust ... ) or create a new one specifically for the purpose (of control)? I'm not really sure what to do there, but interested- Perhaps in my use case (CI system) I don't want to deal with that, so a signal is easier to deal with, however I suspect my use case is rather narrow in scope.

In any case, to support the discussion, I opened a PR (#664 ) with my signal based implementation.

trulede:
Solution using signals is available via #664

PR #664, opened 2023-01-25
Allow sync wait period to be truncated by an HUP signal.

trulede:
For discussion related to issue #660 

The HUP signal is used so that normal use of SIGINT (i.e. CTRL-C) is not affected.

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -35,11 +35,13 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/spf13/pflag"
+	"golang.org/x/sys/unix"
 	"k8s.io/git-sync/pkg/cmd"
 	"k8s.io/git-sync/pkg/hook"
 	"k8s.io/git-sync/pkg/logging"
@@ -77,6 +79,8 @@ var flSyncTimeout = pflag.Duration("sync-timeout", envDuration("GIT_SYNC_SYNC_TI
 	"the total time allowed for one complete sync, must be >= 10ms; --timeout overrides this")
 var flOneTime = pflag.Bool("one-time", envBool("GIT_SYNC_ONE_TIME", false),
 	"exit after the first sync")
+var flSyncOnSignal = pflag.String("sync-on-signal", envString("GIT_SYNC_SYNC_ON_SIGNAL", ""),
+	"sync on receipt of the specified signal (e.g. SIGHUP)")
 var flMaxFailures = pflag.Int("max-failures", envInt("GIT_SYNC_MAX_FAILURES", 0),
 	"the number of consecutive failures allowed before aborting (the first sync must succeed, -1 will retry forever")
 var flChmod = pflag.Int("change-permissions", envInt("GIT_SYNC_PERMISSIONS", 0),
@@ -397,6 +401,24 @@ func main() {
 		handleConfigError(log, true, "ERROR: --period must be at least 10ms")
 	}
 
+	var syncSig syscall.Signal
+	if *flSyncOnSignal != "" {
+		if num, err := strconv.ParseInt(*flSyncOnSignal, 0, 0); err == nil {
+			// sync-on-signal value is a number
+			syncSig = syscall.Signal(num)
+		} else {
+			// sync-on-signal value is a name
+			syncSig = unix.SignalNum(*flSyncOnSignal)
+			if syncSig == 0 {
+				// last resort - maybe they said "HUP", meaning "SIGHUP"
+				syncSig = unix.SignalNum("SIG" + *flSyncOnSignal)
+			}
+		}
+		if syncSig == 0 {
+			handleConfigError(log, true, "ERROR: --sync-on-signal must be a valid signal name or number")
+		}
+	}
+
 	if *flTimeout != 0 {
 		// Back-compat
 		*flSyncTimeout = time.Duration(*flTimeout) * time.Second
@@ -661,6 +683,13 @@ func main() {
 		go exechookRunner.Run(context.Background())
 	}
 
+	// Setup signal notify channel
+	sigChan := make(chan os.Signal, 1)
+	if syncSig != 0 {
+		log.V(2).Info("installing signal handler", "signal", unix.SignalName(syncSig))
+		signal.Notify(sigChan, syncSig)
+	}
+
 	// Craft a function that can be called to refresh credentials when needed.
 	refreshCreds := func(ctx context.Context) error {
 		// These should all be mutually-exclusive configs.
@@ -763,7 +792,16 @@ func main() {
 
 		log.V(1).Info("next sync", "waitTime", flPeriod.String())
 		cancel()
-		time.Sleep(*flPeriod)
+
+		// Sleep until the next sync. If syncSig is set then the sleep may
+		// be interrupted by that signal.
+		t := time.NewTimer(*flPeriod)
+		select {
+		case <-t.C:
+		case <-sigChan:
+			log.V(2).Info("caught signal", "signal", unix.SignalName(syncSig))
+			t.Stop()
+		}
 	}
 }
 
@@ -2077,6 +2115,13 @@ OPTIONS
     --submodules <string>, $GIT_SYNC_SUBMODULES
             The git submodule behavior: one of "recursive", "shallow", or
             "off".  If not specified, this defaults to "recursive".
+
+    --sync-on-signal <string>, $GIT_SYNC_SYNC_ON_SIGNAL
+            Indicates that a sync attempt should occur upon receipt of the 
+            specified signal name (e.g. SIGHUP) or number (e.g. 1). If a sync
+            is already in progress, another sync will be triggered as soon as
+            the current one completes. If not specified, signals will not 
+            trigger syncs.
 
     --sync-timeout <duration>, $GIT_SYNC_SYNC_TIMEOUT
             The total time allowed for one complete sync.  This must be at least
--- a/test_e2e.sh
+++ b/test_e2e.sh
@@ -115,6 +115,14 @@ function docker_kill() {
     docker kill "$1" >/dev/null
 }
 
+function docker_signal() {
+    if [[ -z "$1" || -z "$2" ]]; then
+        echo "usage: $0 <id> <signal>"
+        return 1
+    fi
+    docker kill "--signal=$2" "$1" >/dev/null
+}
+
 # E2E_TAG is the tag used for docker builds.  This is needed because docker
 # tags are system-global, but one might have multiple repos checked out.
 E2E_TAG=$(git rev-parse --show-toplevel | sed 's|/|_|g')
@@ -974,6 +982,108 @@ function e2e::sync_slow_git_long_timeout() {
     echo "$FUNCNAME 2" > "$REPO"/file
     git -C "$REPO" commit -qam "$FUNCNAME 2"
     wait_for_sync 10
+    assert_link_exists "$ROOT"/link
+    assert_file_exists "$ROOT"/link/file
+    assert_file_eq "$ROOT"/link/file "$FUNCNAME 2"
+}
+
+##############################################
+# Test sync-on-signal with SIGHUP
+##############################################
+function e2e::sync_on_signal_sighup() {
+     # First sync
+    echo "$FUNCNAME 1" > "$REPO"/file
+    git -C "$REPO" commit -qam "$FUNCNAME 1"
+
+    GIT_SYNC \
+        --period=100s \
+        --sync-on-signal="SIGHUP" \
+        --repo="file://$REPO" \
+        --branch="$MAIN_BRANCH" \
+        --rev=HEAD \
+        --root="$ROOT" \
+        --link="link" \
+        >> "$1" 2>&1 &
+    wait_for_sync 3
+    assert_link_exists "$ROOT"/link
+    assert_file_exists "$ROOT"/link/file
+    assert_file_eq "$ROOT"/link/file "$FUNCNAME 1"
+
+    # Move HEAD forward
+    echo "$FUNCNAME 2" > "$REPO"/file
+    git -C "$REPO" commit -qam "$FUNCNAME 2"
+    # Send signal (note --period is 100s, signal should trigger sync)
+    CTR=$(docker ps --filter label="git-sync-e2e=$RUNID" --format="{{.ID}}")
+    docker_signal "$CTR" SIGHUP
+    wait_for_sync 3
+    assert_link_exists "$ROOT"/link
+    assert_file_exists "$ROOT"/link/file
+    assert_file_eq "$ROOT"/link/file "$FUNCNAME 2"
+}
+
+##############################################
+# Test sync-on-signal with HUP
+##############################################
+function e2e::sync_on_signal_hup() {
+     # First sync
+    echo "$FUNCNAME 1" > "$REPO"/file
+    git -C "$REPO" commit -qam "$FUNCNAME 1"
+
+    GIT_SYNC \
+        --period=100s \
+        --sync-on-signal="HUP" \
+        --repo="file://$REPO" \
+        --branch="$MAIN_BRANCH" \
+        --rev=HEAD \
+        --root="$ROOT" \
+        --link="link" \
+        >> "$1" 2>&1 &
+    wait_for_sync 3
+    assert_link_exists "$ROOT"/link
+    assert_file_exists "$ROOT"/link/file
+    assert_file_eq "$ROOT"/link/file "$FUNCNAME 1"
+
+    # Move HEAD forward
+    echo "$FUNCNAME 2" > "$REPO"/file
+    git -C "$REPO" commit -qam "$FUNCNAME 2"
+    # Send signal (note --period is 100s, signal should trigger sync)
+    CTR=$(docker ps --filter label="git-sync-e2e=$RUNID" --format="{{.ID}}")
+    docker_signal "$CTR" SIGHUP
+    wait_for_sync 3
+    assert_link_exists "$ROOT"/link
+    assert_file_exists "$ROOT"/link/file
+    assert_file_eq "$ROOT"/link/file "$FUNCNAME 2"
+}
+
+##############################################
+# Test sync-on-signal with 1 (SIGHUP)
+##############################################
+function e2e::sync_on_signal_1() {
+     # First sync
+    echo "$FUNCNAME 1" > "$REPO"/file
+    git -C "$REPO" commit -qam "$FUNCNAME 1"
+
+    GIT_SYNC \
+        --period=100s \
+        --sync-on-signal=1 \
+        --repo="file://$REPO" \
+        --branch="$MAIN_BRANCH" \
+        --rev=HEAD \
+        --root="$ROOT" \
+        --link="link" \
+        >> "$1" 2>&1 &
+    wait_for_sync 3
+    assert_link_exists "$ROOT"/link
+    assert_file_exists "$ROOT"/link/file
+    assert_file_eq "$ROOT"/link/file "$FUNCNAME 1"
+
+    # Move HEAD forward
+    echo "$FUNCNAME 2" > "$REPO"/file
+    git -C "$REPO" commit -qam "$FUNCNAME 2"
+    # Send signal (note --period is 100s, signal should trigger sync)
+    CTR=$(docker ps --filter label="git-sync-e2e=$RUNID" --format="{{.ID}}")
+    docker_signal "$CTR" SIGHUP
+    wait_for_sync 3
     assert_link_exists "$ROOT"/link
     assert_file_exists "$ROOT"/link/file
     assert_file_eq "$ROOT"/link/file "$FUNCNAME 2"


cmd/git-sync/main.go
```diff
@@ -77,6 +78,8 @@ var flSyncTimeout = pflag.Duration("sync-timeout", envDuration("GIT_SYNC_SYNC_TI
 	"the total time allowed for one complete sync, must be >= 10ms; --timeout overrides this")
 var flOneTime = pflag.Bool("one-time", envBool("GIT_SYNC_ONE_TIME", false),
 	"exit after the first sync")
+var flSleepLightly = pflag.Bool("sleep-lightly", envBool("GIT_SYNC_SLEEP_LIGHTLY", false),
```
thockin, line 81: LOL at the name.

How about `--sync-on-signal=<signal-spec>`, where the spec is either:
* an integer for which `unix.SignalName()` returns a non-empty string
* a string for which `unix.SignalNum()` returns non-zero

We can document that as "either a valid signal name (e.g. SIGINT) or signal number (e.g. 15)".

See https://pkg.go.dev/golang.org/x/sys/unix

The default would be "" meaning, disabled.

thockin: Also, down near the bottom of main.go is the manual-text.  Please update that with a nice entry.

cmd/git-sync/main.go
```diff
@@ -980,6 +991,19 @@ func sleepForever() {
...
+// Sleep for specified period but allow interruption from a signal. 	
+func sleepLightly(period time.Duration) {
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, syscall.SIGHUP)
+	t := time.NewTimer(period)
+	select {
+	case <-t.C:
+	case <-c:
+		t.Stop()
+	}
```
thockin, line 1003: leaks the channel, right?  Don't you need to `close()` it?

thockin (commented):
Surprisingly simple.

This also needs an e2e test (test_e2e.sh).

cmd/git-sync/main.go
trulede: I will make an update based on SignalNum().
https://pkg.go.dev/golang.org/x/sys/unix#SignalNum

cmd/git-sync/main.go
```diff
@@ -763,7 +778,13 @@ func main() {
 
 		log.V(1).Info("next sync", "waitTime", flPeriod.String())
 		cancel()
-		time.Sleep(*flPeriod)
+
+		t := time.NewTimer(*flPeriod)
+		select {
+		case <-t.C:
+		case <-signal_ch:
+			t.Stop()
```
thockin, line 786: docs for Timer.Stop say:

```
   To ensure the channel is empty after a call to Stop, check the return value
    and drain the channel. For example, assuming the program has not received
    from t.C already:

        if !t.Stop() {
        	<-t.C
        }
```

thockin:
still needs e2e, also

cmd/git-sync/main.go
trulede: Yes, that documentation mentions the necessity of draining the channel before calling Reset(). Do we need to drain the channel otherwise? Or does GC take care of that?

thockin: I just realized that this is scoped to the loop, so we shouldn't need to worry about it.

trulede:
@thockin Unless, by some miracle, the e2e tests work ... I will need a bit more time to debug them. In the meantime, please let me know if you have any comments regarding test coverage.

cmd/git-sync/main.go
```diff
@@ -397,6 +401,20 @@ func main() {
...
+	if *flSyncOnSignal != "" {
+		if num, err := strconv.ParseInt(*flSyncOnSignal, 0, 0); err == nil {
+			// sync-on-signal value is a number
+			syncSig = syscall.Signal(num)
+		} else {
+			// sync-on-signal value is a name
+			syncSig = unix.SignalNum(*flSyncOnSignal)
+		}
+		if unix.SignalName(syncSig) == "" {
+			handleConfigError(log, true, "ERROR: --sync-on-signal unknown signal")
```
thockin, line 414: nit: ...must be a valid signal name or number

cmd/git-sync/main.go
```diff
@@ -661,6 +679,10 @@ func main() {
 		go exechookRunner.Run(context.Background())
 	}
 
+	// Setup signal notify channel
+	sigChan := make(chan os.Signal, 1)
+	signal.Ignore(syncSig)
```
thockin, line 684: I'm not sure I understand the need to ignore the signal?

What I don't like is that a signal which arrives JUST BEFORE we go to sleep will be totally ignored, but a signal 0.0001 seconds later will be received.

Why not just leave the signal active?

cmd/git-sync/main.go
```diff
@@ -2111,6 +2144,10 @@ OPTIONS
             The git submodule behavior: one of "recursive", "shallow", or
             "off".  If not specified, this defaults to "recursive".
 
+    --sync-on-signal <string>, $GIT_SYNC_SYNC_ON_SIGNAL
+            Indicate that a sync should occur on receipt of the specifed signal.
```
thockin, line 2148: """
Indicates that an immediate sync should occur upon receipt of the specified signal name (e.g. SIGHUP) or number (e.g. 1).  If not specified, signals will not trigger syncs.
"""

test_e2e.sh
```diff
@@ -115,6 +115,14 @@ function docker_kill() {
     docker kill "$1" >/dev/null
 }
 
+function docker_signal_process() {
+    if [[ -z "$3" ]]; then
+        echo "usage: $0 <id> <signal> <name>"
+        return 1
+    fi
+    docker exec -t $1 "killall -s $2 $3"
```
thockin, line 123: `docker kill -s <signal>` exists for this (a rather hidden feature)

`docker kill -s "$2" "$1" >/dev/null`

test_e2e.sh
```diff
@@ -1063,6 +1071,74 @@ function e2e::sync_slow_git_long_timeout() {
     assert_file_eq "$ROOT"/link/file "$FUNCNAME 2"
 }
 
+##############################################
+# Test sync-on-signal with SIGHUP
+##############################################
+function e2e::sync_on_signal() {
```
thockin, line 1077: sync_on_signal_name

test_e2e.sh
```diff
@@ -1063,6 +1071,74 @@ function e2e::sync_slow_git_long_timeout() {
...
+    wait_for_sync 3
+    aassert_link_exists "$ROOT"/link
+    assert_file_exists "$ROOT"/link/file
+    assert_file_eq "$ROOT"/link/file "$FUNCNAME 2"
+}
+
+##############################################
+# Test sync-on-signal with 1 (SIGHUP)
+##############################################
+function e2e::sync_on_signal() {
```
thockin, line 1111: sync_on_signal_num

test_e2e.sh
```diff
@@ -1063,6 +1071,74 @@ function e2e::sync_slow_git_long_timeout() {
...
+    assert_link_exists "$ROOT"/link
+    assert_file_exists "$ROOT"/link/file
+    assert_file_eq "$ROOT"/link/file "$FUNCNAME 1"
+
+    # Move HEAD forward
+    echo "$FUNCNAME 2" > "$REPO"/file
+    git -C "$REPO" commit -qam "$FUNCNAME 2"
+    # Send signal (note --period is 100s)
+    CTR=$(docker ps --filter label="git-sync-e2e=$RUNID" --format="{{.ID}}")
+    docker_signal_process "$CTR", SIGHUP, git-sync
```
thockin, line 1101: no commas needed on shell args :)

test_e2e.sh
```diff
@@ -1063,6 +1071,74 @@ function e2e::sync_slow_git_long_timeout() {
...
+    assert_link_exists "$ROOT"/link
+    assert_file_exists "$ROOT"/link/file
+    assert_file_eq "$ROOT"/link/file "$FUNCNAME 1"
+
+    # Move HEAD forward
+    echo "$FUNCNAME 2" > "$REPO"/file
+    git -C "$REPO" commit -qam "$FUNCNAME 2"
+    # Send signal (note --period is 100s)
+    CTR=$(docker ps --filter label="git-sync-e2e=$RUNID" --format="{{.ID}}")
+    docker_signal_process "$CTR", 1, git-sync
```
thockin, line 1135: no commas needed on shell args :)

test_e2e.sh
```diff
@@ -1063,6 +1071,74 @@ function e2e::sync_slow_git_long_timeout() {
...
+    wait_for_sync 3
+    aassert_link_exists "$ROOT"/link
```
thockin, line 1137: s/aa/a

test_e2e.sh
```diff
@@ -1063,6 +1071,74 @@ function e2e::sync_slow_git_long_timeout() {
...
+    assert_file_eq "$ROOT"/link/file "$FUNCNAME 1"
+
+    # Move HEAD forward
+    echo "$FUNCNAME 2" > "$REPO"/file
+    git -C "$REPO" commit -qam "$FUNCNAME 2"
+    # Send signal (note --period is 100s)
+    CTR=$(docker ps --filter label="git-sync-e2e=$RUNID" --format="{{.ID}}")
+    docker_signal_process "$CTR", SIGHUP, git-sync
+    wait_for_sync 3
+    aassert_link_exists "$ROOT"/link
```
thockin, line 1103: s/aa/a

test_e2e.sh
```diff
@@ -115,6 +115,14 @@ function docker_kill() {
     docker kill "$1" >/dev/null
 }
 
+function docker_signal_process() {
+    if [[ -z "$3" ]]; then
```
thockin, line 119: if [[ -z "$1" || -z "$2" ]]; then

thockin (commented):
test looks good, but they can't be the same name :)

Here's my accumulated diff, after which tests pass!

```diff
diff --git a/test_e2e.sh b/test_e2e.sh
index bc8184e..352e573 100755
--- a/test_e2e.sh
+++ b/test_e2e.sh
@@ -115,12 +115,12 @@ function docker_kill() {
     docker kill "$1" >/dev/null
 }
 
-function docker_signal_process() {
-    if [[ -z "$3" ]]; then
-        echo "usage: $0 <id> <signal> <name>"
+function docker_signal() {
+    if [[ -z "$1" || -z "$2" ]]; then
+        echo "usage: $0 <id> <signal>"
         return 1
     fi
-    docker exec -t $1 "killall -s $2 $3"
+    docker kill -s "$2" "$1" >/dev/null
 }
 
 # E2E_TAG is the tag used for docker builds.  This is needed because docker
@@ -1074,7 +1074,7 @@ function e2e::sync_slow_git_long_timeout() {
 ##############################################
 # Test sync-on-signal with SIGHUP
 ##############################################
-function e2e::sync_on_signal() {
+function e2e::sync_on_signal_name() {
      # First sync
     echo "$FUNCNAME 1" > "$REPO"/file
     git -C "$REPO" commit -qam "$FUNCNAME 1"
@@ -1098,9 +1098,9 @@ function e2e::sync_on_signal() {
     git -C "$REPO" commit -qam "$FUNCNAME 2"
     # Send signal (note --period is 100s)
     CTR=$(docker ps --filter label="git-sync-e2e=$RUNID" --format="{{.ID}}")
-    docker_signal_process "$CTR", SIGHUP, git-sync
+    docker_signal "$CTR" SIGHUP
     wait_for_sync 3
-    aassert_link_exists "$ROOT"/link
+    assert_link_exists "$ROOT"/link
     assert_file_exists "$ROOT"/link/file
     assert_file_eq "$ROOT"/link/file "$FUNCNAME 2"
 }
@@ -1108,7 +1108,7 @@ function e2e::sync_on_signal() {
 ##############################################
 # Test sync-on-signal with 1 (SIGHUP)
 ##############################################
-function e2e::sync_on_signal() {
+function e2e::sync_on_signal_num() {
      # First sync
     echo "$FUNCNAME 1" > "$REPO"/file
     git -C "$REPO" commit -qam "$FUNCNAME 1"
@@ -1132,9 +1132,9 @@ function e2e::sync_on_signal() {
     git -C "$REPO" commit -qam "$FUNCNAME 2"
     # Send signal (note --period is 100s)
     CTR=$(docker ps --filter label="git-sync-e2e=$RUNID" --format="{{.ID}}")
-    docker_signal_process "$CTR", 1, git-sync
+    docker_signal "$CTR" 1
     wait_for_sync 3
-    aassert_link_exists "$ROOT"/link
+    assert_link_exists "$ROOT"/link
     assert_file_exists "$ROOT"/link/file
     assert_file_eq "$ROOT"/link/file "$FUNCNAME 2"
 }
```

cmd/git-sync/main.go
trulede: What I found here was that if the signal was not disabled at this point, then the git process would receive the signal (and terminate). I would like to find a better way of dealing with that ... possibly having a signal handler function (goroutine) catch the signal.

test_e2e.sh
trulede: Thanks! Killall was not working anyway (not included in the image).

cmd/git-sync/main.go
trulede: No, that's fine, now that you point it out ... it's obviously wrong.

cmd/git-sync/main.go
trulede: This might touch on what is happening (https://stackoverflow.com/questions/33165530/prevent-ctrlc-from-interrupting-exec-command-in-golang).

_Perhaps_ a change here: https://github.com/trulede/git-sync/blob/3bf26ac2be804b7337a4951191fd636f2a0a5325/pkg/cmd/cmd.go#L56 , possibly triggered by putting something in the context (ctx) so that its only active when `sigChan` is set

thockin: The answer at thar StackOverflow seems promising:

```diff
diff --git a/pkg/cmd/cmd.go b/pkg/cmd/cmd.go
index bdadf7d..c869228 100644
--- a/pkg/cmd/cmd.go
+++ b/pkg/cmd/cmd.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"os/exec"
 	"strings"
+	"syscall"
 
 	"github.com/go-logr/logr"
 )
@@ -72,6 +73,9 @@ func runWithStdin(ctx context.Context, log logintf, cwd string, env []string, st
 	cmd.Stdout = outbuf
 	cmd.Stderr = errbuf
 	cmd.Stdin = bytes.NewBufferString(stdin)
+	cmd.SysProcAttr = &syscall.SysProcAttr{
+		Setpgid: true,
+	}
 
 	err := cmd.Run()
 	stdout := strings.TrimSpace(outbuf.String())
```

Can we craft an e2e case that forces this condition?

trulede: It might be possible to use the slow fetch script (see e2e::sync_fetch_skip_depth_1) and do something like this:

1. Start the sync. Shell script for a "slow fetch" runs in own process, sleeps for 5, then runs the git command.
2. Test sleeps 1 second, to be sure that the shell script has started.
3. Test sends the signal. This signal should _not_ be propagated to the shell script.
4. Wait for the git-sync to complete, wait 4+3 seconds (remaining wait time + normal 3 seconds).

I will try it tomorrow.

thockin: I'm trying to craft an e2e that forces this but I can't, so far.

thockin: I'm not buying it.  Here's what I tried.

```
diff --git a/slow_git_clone.sh b/slow_git_clone.sh
index 0724cdb..449e5cd 100755
--- a/slow_git_clone.sh
+++ b/slow_git_clone.sh
@@ -20,5 +20,9 @@ if [ "$1" != "clone" ]; then
   exit $?
 fi
 
-sleep 1.1
+trap "exit 1" HUP
+# shell needs to be foreground for signals to be delivered
+sleep 1.1 &
+wait
+
 git "$@"
index 0671e2b..bc1fcb7 100755
--- a/test_e2e.sh
+++ b/test_e2e.sh
@@ -1139,6 +1139,33 @@ function e2e::sync_on_signal_number() {
     assert_file_eq "$ROOT"/link/file "$FUNCNAME 2"
 }
 
+# Test sync-on-signal doesn't interrupt git
+function e2e::sync_on_signal_not_interrupt() {
+     # First sync
+    echo "$FUNCNAME 1" > "$REPO"/file
+    git -C "$REPO" commit -qam "$FUNCNAME 1"
+
+    GIT_SYNC \
+        --git="$SLOW_GIT_CLONE" `# will exit on SIGHUP` \
+        --period=100s \
+        --sync-on-signal="SIGHUP" \
+        --branch="$MAIN_BRANCH" \
+        --rev=HEAD \
+        --root="$ROOT" \
+        --link="link" \
+        >> "$1" 2>&1 &
+    sleep 1 # long enough to get into the slow clone
+    CTR=$(docker ps --filter label="git-sync-e2e=$RUNID" --format="{{.ID}}")
+    docker_signal "$CTR" SIGHUP # should kill it
+    wait_for_sync 3
+    assert_link_exists "$ROOT"/link
+    assert_file_exists "$ROOT"/link/file
+    assert_file_eq "$ROOT"/link/file "$FUNCNAME 1"
+}
```

but it never dies - you can futz with the timing so that is GURANTEED and it still does not pass thru git-sync to the git.

thockin:
While playing to try to repro the signal thing I accumulated this diff - please apply if it makes sense

```diff
diff --git a/cmd/git-sync/main.go b/cmd/git-sync/main.go
index c94e0d0..7e6fa21 100644
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -409,8 +409,12 @@ func main() {
                } else {
                        // sync-on-signal value is a name
                        syncSig = unix.SignalNum(*flSyncOnSignal)
+                       if syncSig == 0 {
+                               // last resort - maybe they said "HUP", meaning "SIGHUP"
+                               syncSig = unix.SignalNum("SIG" + *flSyncOnSignal)
+                       }
                }
-               if unix.SignalName(syncSig) == "" {
+               if syncSig == 0 {
                        handleConfigError(log, true, "ERROR: --sync-on-signal must be a valid signal name or number")
                }
        }
@@ -681,7 +685,10 @@ func main() {
 
        // Setup signal notify channel
        sigChan := make(chan os.Signal, 1)
-       signal.Ignore(syncSig)
+       if syncSig != 0 {
+               log.V(2).Info("installing signal handler", "signal", unix.SignalName(syncSig))
+               signal.Notify(sigChan, syncSig)
+       }
 
        // Craft a function that can be called to refresh credentials when needed.
        refreshCreds := func(ctx context.Context) error {
@@ -787,16 +794,14 @@ func main() {
                cancel()
 
                // Sleep until the next sync. If syncSig is set then the sleep may
-               // be interrupted by that signal. After sleeping, ignore the signal
-               // again (so that it does not interrupt the git sync).
-               signal.Notify(sigChan, syncSig)
+               // be interrupted by that signal.
                t := time.NewTimer(*flPeriod)
                select {
                case <-t.C:
                case <-sigChan:
+                       log.V(2).Info("caught signal", "signal", unix.SignalName(syncSig))
                        t.Stop()
                }
-               signal.Ignore(syncSig)
        }
 }
 
```

cmd/git-sync/main.go
thockin:
```keyboard ^C signals the PGRP.  Also SIGTERM seems like a poor choice ;)

Try:
```
rm -rf /tmp/gs; bin/linux_amd64/git-sync -v5 --repo https://github.com/kubernetes/kubernetes --period=6s --root /tmp/gs --sync-on-signal SIGTERM & sleep 2; echo kill; kill -TERM $!; sleep 1; echo jobs; jobs
```

Comment (trulede) [reply cmd/git-sync/main.go]: Alright. I was not installing the signal handler until after the initial clone/sync. So any attempts at testing on the first clone would not have the desired behaviour ... which confused matters.

Need to check if the behaviour that I previously mentioned is still occuring (manually CTRL-C during the sync). It should, since I would do that after the initial clone. And then ... try again with the E2E test.

File: cmd/git-sync/main.go [hunk 15]
```diff
@@ -787,16 +787,14 @@ func main() {
 		cancel()
 
 		// Sleep until the next sync. If syncSig is set then the sleep may
-		// be interrupted by that signal. After sleeping, ignore the signal
-		// again (so that it does not interrupt the git sync).
+		// be interrupted by that signal.
 		signal.Notify(sigChan, syncSig)
```
thockin, line 791: you should not need to run this again

cmd/git-sync/main.go
trulede: I had no luck in getting a signal to propagate from git-sync to the slow_git_clone.sh script. My last attempt looked like:

```
if [ "$1" != "clone" ]; then
  git "$@"
  exit $?
fi

set +m
signal_handler() {
  echo XXXXX SIGNAL YYYYY
#  exit
}

trap "signal_handler" HUP INT
sleep 5.1
git "$@"
```

Works perfectly fine from the shell.

thockin: signals in shell are tricky - during an exec() the shell blocks signals.  sleep is an exec().  Try this:

```
sleep 5.1 &
wait
```

`wait` is a shell builtin.

thockin: signals in shell are tricky - during an exec() the shell blocks signals.  sleep is an exec().  Try this:

```
sleep 5.1 &
wait
```

`wait` is a shell builtin.

trulede: > ```diff
> cmd.SysProcAttr = &syscall.SysProcAttr{
> +		Setpgid: true,
> +	}
> ```

With `Setpgid: false`, and running this command:
```
$ bin/linux_amd64/git-sync --repo https://github.com/trulede/raspberry.docker.git --period 6s -v 5 --root /tmp/foo --sync-on-signal="SIGINT"
....
{"logger":"","ts":"2023-02-15 19:00:48.679291","caller":{"file":"main.go","line":794},"level":1,"msg":"next sync","waitTime":"6s"}
{"logger":"","ts":"2023-02-15 19:00:54.684689","caller":{"file":"main.go","line":715},"level":1,"msg":"syncing repo"}
{"logger":"","ts":"2023-02-15 19:00:54.684872","caller":{"file":"main.go","line":949},"level":0,"msg":"sanity-checking git repo","repo":"/tmp/foo"}
{"logger":"","ts":"2023-02-15 19:00:54.685222","caller":{"file":"cmd.go","line":55},"level":5,"msg":"running command","cwd":"/tmp/foo","cmd":"git rev-parse --show-toplevel"}
{"logger":"","ts":"2023-02-15 19:00:54.688410","caller":{"file":"cmd.go","line":55},"level":5,"msg":"running command","cwd":"/tmp/foo","cmd":"git fsck --no-progress --connectivity-only"}
{"logger":"","ts":"2023-02-15 19:00:54.704816","caller":{"file":"main.go","line":930},"level":0,"msg":"root directory is valid","path":"/tmp/foo"}
{"logger":"","ts":"2023-02-15 19:00:54.705000","caller":{"file":"cmd.go","line":55},"level":5,"msg":"running command","cwd":"/tmp/foo","cmd":"git rev-parse HEAD"}
{"logger":"","ts":"2023-02-15 19:00:54.708792","caller":{"file":"cmd.go","line":55},"level":5,"msg":"running command","cwd":"/tmp/foo","cmd":"git ls-remote -q https://github.com/trulede/raspberry.docker.git HEAD"}
^C{"logger":"","ts":"2023-02-15 19:00:54.988749","caller":{"file":"main.go","line":730},"msg":"too many failures, aborting","error":"Run(git ls-remote -q https://github.com/trulede/raspberry.docker.git HEAD): signal: interrupt: { stdout: \"\", stderr: \"\" }","failCount":1}
```
... it is possible to "CTRL-C" just as the next sync begins, killing the underlying command (git ls-remote in this case). As a result git-sync exits.

Setting `Setpgid: true` prevents that behaviour.

thockin: Again, ^C seems to confuse things.  Try this instead

```
rm -rf /tmp/gs; bin/linux_amd64/git-sync --one-time --repo https://github.com/kubernetes/kubernetes -v 5 --root /tmp/gs --sync-on-signal="SIGINT"
```

I chose this repo because it is big and slow to clone, making timing easier.

Now hit ^C and see that git does get killed:

```
{"logger":"","ts":"2023-02-15 10:43:33.392142","caller":{"file":"cmd.go","line":54},"level":5,"msg":"running command","cwd":"","cmd":"git clone --no-checkout https://github.com/kubernetes/kubernetes /tmp/gs"}
^C{"logger":"","ts":"2023-02-15 10:43:34.768060","caller":{"file":"main.go","line":725},"msg":"too many failures, aborting","error":"Run(git clone --no-checkout https://github.com/kubernetes/kubernetes /tmp/gs): signal: interrupt: { stdout: \"\", stderr: \"Cloning into '/tmp/gs'...\\nfetch-pack: unexpected disconnect while reading sideband packet\" }","failCount":1}
```

But this is not really the representative case.  Now try:

```
{"logger":"","ts":"2023-02-15 10:47:48.278868","caller":{"file":"cmd.go","line":54},"level":5,"msg":"running command","cwd":"","cmd":"git clone --no-checkout https://github.com/kubernetes/kubernetes /tmp/gs"}


SENDING SIGKILL


```

It will eventually complete the sync (it takes well over a minute for me :)

So, I think "hit ^C" is actually doing the right thing - you are telling your shell to "stop this process" which REALLY does mean the whole PGroup.  ^C is more than "send a signal", and doesn't represent how people will use this.

pkg/cmd/cmd.go
```diff
@@ -65,6 +66,14 @@ func (r *Runner) RunWithStdin(ctx context.Context, cwd string, env []string, std
 	cmd.Stdout = outbuf
 	cmd.Stderr = errbuf
 	cmd.Stdin = bytes.NewBufferString(stdin)
+	if v := ctx.Value("syncSig"); v != nil {
```
thockin, line 69: I initially said I was OK with this, but when I apply this and do manual testing, I *want* ^C to kill git.  So I don't think we should do this part.

cmd/git-sync/main.go
```diff
@@ -2111,6 +2150,11 @@ OPTIONS
             The git submodule behavior: one of "recursive", "shallow", or
             "off".  If not specified, this defaults to "recursive".
 
+    --sync-on-signal <string>, $GIT_SYNC_SYNC_ON_SIGNAL
+            Indicates that an immediate sync should occur upon receipt of the 
+            specified signal name (e.g. SIGHUP) or number (e.g. 1). If not
```
thockin, line 2155: for clarity, something like:

"""
Indicates that a sync attempt should occur upon receipt of the specified signal name (e.g. SIGHUP) or number (e.g. 1).  If a sync is already in progress, another sync will be triggered as soon as the current one completes.  If not specified...
"""

The key delta being "immediate" semantics.

thockin (commented):
Other than these two points it looks good to me!

thockin:
Also, you probably want to squash down to 1 commit with a good message and un "draft" this PR :)

pkg/cmd/cmd.go
trulede: ok.

thockin:
It looks like your rebase accidentally pulled another commit in - not sure why?  Try a full `get fetch upstream; git rebase -i upstream/master` ?

thockin (commented):
Thanks!  

/lgtm
/approve

thockin:
Should we close #660 now?  Or leave it open in hopes that someone wants to do an HTTP version (which needs its own discussion :)

trulede:
> Should we close #660 now? Or leave it open in hopes that someone wants to do an HTTP version (which needs its own discussion :)

I just closed it. Its easy enough to implement a sidecar to terminate HTTP and send a signal towards git-sync 💡 

Thanks for all the help with the PR!

PR #666, opened 2023-02-04
Update bullseye image to address vulnerability

victorpras:
Update bullseye image to address vulnerability


thockin (commented):
I don't think I can target this project at a GKE base image, publicly readable or not.

Part of the build for git-sync is apt-get upgrade - does that not solve the vuln?

@tallclair is the k8s debian-base stale?

victorpras:
Looks like this is not needed to fix the vulnerabilities

PR #670, opened 2023-02-07
Fix e2e on v4

thockin:
Looking to cut a CVE release I found 2 things:

1) git broke itself (disallowing file:// which we need in e2e)
2) through a typo I have been relying on long-ago built image names to pass e2e


stp-ip (approved):
/lgtm

PR #673, opened 2023-02-11
v4: e2e improvements

thockin:
* Set RUNS to run tests cases multiple times
* Print path to logfiles upon failure
* Set MAXWAIT to fine-tune how long we are willing to wait for a sync
* Assert metrics in tests


stp-ip (approved):
/lgtm

PR #674, opened 2023-02-11
Require --http-bind when using other http flags

thockin:
Also use the word "set" for bool flags, rather than "specified" (easier to read.


stp-ip (approved):
/lgtm

PR #676, opened 2023-02-11
v4: Deref tags on ls-remote

thockin:
A previous commit (2f7335868eec846f12ea90342d67f1ba870caa07) introduced a quiet bug which results in the "update needed" condition triggering every loop.  e2e passed for me by sheer luck of winning races.  Until it didn't.  Walking thru with the debugger to figure it out, I realized this issue.

The short story: `ls-remote` for a tag gets us the SHA of the tag, but `rev-parse HEAD` gets us the SHA of the commit to which that tag is attached.  Those are never equal, so we detect "update needed" every loop.

Now we ask `ls-remote` for the rev and the dereferenced rev.  If that rev is a branch, the deref does nothing.  If that rev is a tag it produces both results.  ls-remote does its own sort, so the deref (if found) comes after the non-deref.  This means that, in both cases, the last line is the one we want.


stp-ip (approved):
/lgtm

PR #677, opened 2023-02-11
v4: Default http-metrics to false

thockin:
After the change to require http-bind, this default doesn't make sense. My local branch had other commits, which hid this error.


thockin:
Self-merge to fix e2e

PR #682, opened 2023-02-16
e2e: rename init tests, remove default-val args

thockin:
Also don't pass --rev=HEAD when that's not part of the test.  It's the default value.


stp-ip (approved):
/lgtm

PR #686, opened 2023-02-25
e2e: compare numbers numerically, not lexically

thockin:
Thanks, bash, for supporting `<` and `>` and making them NOT mean the obvious thing.

This shows up when you set RUNS=10:
`[[ 2 < 10 ]]` is false
`(( 2 < 10 ))` is true


thockin:
Gentle ping on this - I have more in the queue :)

PR #689, opened 2023-02-25
Use a private gitconfig file

thockin:
This was changing my own ~/.gitconfig whenever run by hand outside of a container.  Oops.


thockin:
Gentle ping on this - I have more in the queue :)

PR #691, opened 2023-02-25
Total overhaul of the fetch loop for v4                                                                                                                     

thockin:
This one is a doozy but can't easily be broken up further.  I have a dozen more on top of this, building to a v4 release candidate, I hope.

The previous (v3) sync loop betrays my lack of understanding about git.  It tried to codify my archaic mental model (e.g. --branch and --rev being disting things) and was ultimately a patchwork of corner-cases evolved over a few years.

This commit is less of a "diff" and more of a "rewrite".

The new logic is simpler and more efficient.  It does not `git clone` ever.  It does not differentiate the first sync from subsequent syncs.  It uses `git fetch` to get the exact SHA and then makes a worktree from that.

The new `--ref` flag replaces both `--rev` and `--branch`, though it will use those if specified.  In fact, almost all of the e2e tests passed without change - using --ref and --branch!

I will follow this commit up with more cleanups and e2es.


thockin:
I am so sorry for this one :)

thockin:
self-merge to make progress

stp-ip:
Damn thought you self merged all. Will review after the fact once I have some capacity. So don't mind you merging your progress.

thockin:
No sweat, this was a LOT of load to dump, and I have more queued up!  :)

PR #700, opened 2023-03-17
V3: prevent git's 'dubious ownership' error

### LINKED ISSUE #696: update to 3.6.4 breaks git sync
git-sync v3.6.4 breaks the cloning process in our k8s pods with the issue mentioned. However, it works fine until version v3.6.3 . 

Note: 
1. We use **submodules** in our repos.
2. This is **git-sync** deployed as sidecar container in airflow via helm

````
12 main.go:535] "msg"="too many failures, aborting" "error"="Run(git rev-parse HEAD): exit status 128: { stdout: "", stderr: "fatal: detected dubious ownership in repository at '/dags'
To add an exception for this directory, call:

\tgit config --global --add safe.directory /dags" }" "failCount"=1
````

thockin:
To prove this, we run the e2e test as a different user.  To do that, we need git-sync to make sure that everything is group accessible.  To clean up after the test, we need everything to be group writable.  To do that, we add a new flag: `--group-write`.

Fixes #696

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -36,6 +36,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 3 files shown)


thockin:
I have a V4 equivalent, but it's on top of the stack of PRs already out for review :)

stp-ip:
/lgtm

Have the v4 PRs on the radar, but a bit underwater.

thockin:
I can appreciate being under-water.  These PRs are the only fresh air I have gotten in a while :)

PR #710, opened 2023-04-22
Tidy up log levels - logs read better now

thockin:


thockin:
Self-merge to move forward

PR #711, opened 2023-04-22
Add test for repo size and change default GC

thockin:
--git-gc=always seems to be the right tradeoff.


thockin:
Self-merge to move forward

PR #712, opened 2023-04-22
Change the default depth to 1

thockin:
For most users this is functionally equivalent and more efficient.  For users who really need the full history, set this to 0.


thockin:
Self-merge to move forward

PR #713, opened 2023-04-22
Fine tune git command verbosity flags

thockin:


thockin:
Self-merge to move forward

Issue #692, opened 2023-02-27
Keep [N] Old Worktrees


doddsiedodds:
Hey,

We're using git-sync in a variety of application deployments (for example, apache airflow for DAG code) and find that when a workdir is deleted from under the application's feet, it tends to not handle it gracefully.

I was thinking of forking and adding support for either 1) keeping N previous worktrees or 2) keep all previous worktrees. 2 is significantly easier to implement I expect, I would then write application code to clean-up workdirs over a certain age.

Is this something you've considered before or, if I was to do the work, something you'd consider as a feature?

Cheers,
Matt

thockin:
I am in the middle of a major overhaul for V4, but it doesn't change this
particular behavior (it does change a lot of the code, though!).

I have never considered this, no. Not that I am, per se, against it but I
don't think anyone has asked for it.

Are you using the worktree as the working directory for you app?  That
could be a problem, for sure.  Or are you just not happy with doing
file-accesses across two different SHAs, even if the switchover is atomic
and otherwise undetectable?

Part of the new v4 work is a focus on efficiency - for example making
--depth default to 1.  Keeping _all_ worktrees would change that a lot, but
not so bad if you are opting in to it.  Impl should be easy.

A FIFO of worktrees seems cute, but requires an index to track, which makes
it more complicated.  So unless that is ACTUALLY the needed behavior, I'd
avoid it.

Help me understand what the app is doing?

doddsiedodds:
Exactly as described, the worktree is a working directory for the app.  As a concrete example, apache airflow loads a large set of python scripts containing DAG definitions from a git-synced directory, if the current tree symlink is flipped and the old folder is cleaned up while its going through this process it can cause errors.

Agreed re: FIFO being significantly more complicated to implement, it's not _needed_ per se, just that without it, we'd end up probably writing a garbage collection/housekeeping job to clean up old worktrees after some period of time. The reason I considered git-sync itself taking care of it via a FIFO queue was that I feel the expected pattern is that git-sync is the *only* thing managing the folders in the root checkout folder.

thockin:
Does it need to be the working directory?  Can the app just refer to files by a fuller path?  That would trade this problem for a different one, in particular I could read file A from hash abc123 and file B from hash def456.

The problem with FIFO is that it just extends the problem from "delete now" to "delete later" without any indication of whether/when it is actually OK to do so.

Keeping all worktrees is disk-inefficient, for sure.

[...]

PR #715, opened 2023-04-28
Add --stale-worktree-timeout option

sviscaino:
Adds an option to wait for a certain duration before cleaning up the old worktrees.

Test:

https://user-images.githubusercontent.com/23248585/235175534-240fc875-5931-4f6f-886b-4624ec69989d.mp4

Resolves https://github.com/kubernetes/git-sync/issues/692

### DIFF
--- a/cmd/git-sync/main.go
+++ b/cmd/git-sync/main.go
@@ -21,8 +21,10 @@ package main // import "k8s.io/git-sync/cmd/git-sync"
 import (
 	"context"
 	"crypto/md5"
+	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"net"
 	"net/http"
 	"net/http/pprof"
@@ -165,6 +167,9 @@ var flCookieFile = pflag.Bool("cookie-file",
 var flAskPassURL = pflag.String("askpass-url",
 	envString("", "GITSYNC_ASKPASS_URL", "GIT_SYNC_ASKPASS_URL", "GIT_ASKPASS_URL"),
 	"a URL to query for git credentials (username=<value> and password=<value>)")
+
+var flStaleWorktreeTimeout = pflag.Duration("stale-worktree-timeout", envDuration(0, "GITSYNC_STALE_WORKTREE_TIMEOUT"),
+	"how long to retain non-current worktrees")
 
 var flGitCmd = pflag.String("git",
 	envString("git", "GITSYNC_GIT", "GIT_SYNC_GIT"),
@@ -475,19 +480,20 @@ func (abs absPath) Base() string {
 
 // repoSync represents the remote repo and the local sync of it.
 type repoSync struct {
-	cmd        string         // the git command to run
-	root       absPath        // absolute path to the root directory
-	repo       string         // remote repo to sync
-	ref        string         // the ref to sync
-	depth      int            // for shallow sync
-	submodules submodulesMode // how to handle submodules
-	gc         gcMode         // garbage collection
-	link       absPath        // absolute path to the symlink to publish
-	authURL    string         // a URL to re-fetch credentials, or ""
-	sparseFile string         // path to a sparse-checkout file
-	syncCount  int            // how many times have we synced?
-	log        *logging.Logger
-	run        cmd.Runner
+	cmd          string         // the git command to run
+	root         absPath        // absolute path to the root directory
+	repo         string         // remote repo to sync
+	ref          string         // the ref to sync
+	depth        int            // for shallow sync
+	submodules   submodulesMode // how to handle submodules
+	gc           gcMode         // garbage collection
+	link         absPath        // absolute path to the symlink to publish
+	authURL      string         // a URL to re-fetch credentials, or ""
+	sparseFile   string         // path to a sparse-checkout file
+	syncCount    int            // how many times have we synced?
+	log          *logging.Logger
+	run          cmd.Runner
+	staleTimeout time.Duration // time for worktrees to be cleaned up
 }
 
 func main() {
@@ -791,18 +797,19 @@ func main() {
 
 	// Capture the various git parameters.
 	git := &repoSync{
-		cmd:        *flGitCmd,
-		root:       absRoot,
-		repo:       *flRepo,
-		ref:        *flRef,
-		depth:      *flDepth,
-		submodules: submodulesMode(*flSubmodules),
-		gc:         gcMode(*flGitGC),
-		link:       absLink,
-		authURL:    *flAskPassURL,
-		sparseFile: *flSparseCheckoutFile,
-		log:        log,
-		run:        cmdRunner,
+		cmd:          *flGitCmd,
+		root:         absRoot,
+		repo:         *flRepo,
+		ref:          *flRef,
+		depth:        *flDepth,
+		submodules:   submodulesMode(*flSubmodules),
+		gc:           gcMode(*flGitGC),
+		link:         absLink,
+		authURL:      *flAskPassURL,
+		sparseFile:   *flSparseCheckoutFile,
+		log:          log,
+		run:          cmdRunner,
+		staleTimeout: *flStaleWorktreeTimeout,
 	}
 
 	// This context is used only for git credentials initialization. There are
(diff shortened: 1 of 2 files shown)


thockin (commented):
This looks pretty nice!  Unfortunately I have a bunch of PRs queued up, so one of us will have to resolve conflicts.  I'll get this in ASAP.

One thing that is missing is what happens if git-sync crashes and restarts.  We still want to clean up.  To do that, we need the timestamp to survive app restarts.  What if we use the worktree's mtime?  When we flip the link, if we don't delete the old wortree, instead touch it.  Then when we do cleanup (my PRs change this area), we read the mtime and only clean up if the mtime is > timeout.

Then you don't need the map at all.

The tests need to go into test_e2e.sh.  Specifically:
* Prove it works for 3 stale worktrees with a 3 second timeout
* Prove it works across restarts (using --run-once)

cmd/git-sync/main.go
```diff
@@ -2074,6 +2096,10 @@ OPTIONS
             Enable the pprof debug endpoints on git-sync's HTTP endpoint.
             Requires --http-bind to be specified.
 
+    --keep-worktrees-duration, $GIT_SYNC_KEEP_WORKTREES_DURATION
+            Keep the past worktrees for the given number of seconds before
```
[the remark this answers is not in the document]
sviscaino: Yeah that's a better description, thanks - made the change

thockin (commented):
Ping me when rebased (sorry) and updated.  Looking good.

cmd/git-sync/main.go
```diff
@@ -1272,6 +1299,15 @@ func removeDirContents(dir string, log *logging.Logger, except ...string) error
 			continue
 		}
 		p := filepath.Join(dir, name)
+		if exceptMoreRecentThan > 0 {
+			stat, err := os.Stat(p)
+			if err != nil {
+				log.Error(err, "Unable to stat path", "path", p)
+			}
+			if stat.ModTime().Add(exceptMoreRecentThan).After(time.Now()) {
```
thockin, line 1307: Can this be rephrased as `time.Since(stat.ModTime()) > exceptMoreRecentThan`

cmd/git-sync/main.go
```diff
@@ -1254,6 +1277,10 @@ func dirIsEmpty(dir string) (bool, error) {
 // removeDirContents iterated the specified dir and removes all contents,
 // except entries which are specifically excepted.
 func removeDirContents(dir string, log *logging.Logger, except ...string) error {
+	return removeDirContentsExceptMostRecent(dir, log, 0, except...)
+}
+
+func removeDirContentsExceptMostRecent(dir string, log *logging.Logger, exceptMoreRecentThan time.Duration, except ...string) error {
```
thockin, line 1283: How about a slightly different factoring.  I'm thinking something like:

```
func removeDirContentsIf(dir string, log *logging.Logger, fn func(fi os.FileInfo,) (bool, error) ) error {
```

That would read each dirent, call `Stat`, and call the `fn`.  If `fn` returns error, it's an error.  If it returns true, remove the path.

That way, `removeDirContents` just passes a nil func, and the cleanup path can return a local func which compares the timestamps and filenames.  

I *think* this gets is better locality of logic, what do you think?

cmd/git-sync/main.go
```diff
@@ -1692,6 +1728,14 @@ func (git *repoSync) SyncRepo(ctx context.Context, refreshCreds func(context.Con
 		if err != nil {
 			return false, "", err
 		}
+		if currentWorktree != "" {
+			// Touch the old worktree -- which will make cleanupStaleWorktrees delete it in the future,
+			// if the stale timeout option is enabled
+			err = os.Chtimes(currentWorktree.Path(), time.Now(), time.Now())
```
thockin, line 1734: We could encapsulate this into `touch()` so it's like:

```
err = chtimes
if err is NotFound error
    os.Create()
```

test_e2e.sh
```diff
@@ -576,6 +576,238 @@ function e2e::worktree_cleanup() {
...
+    assert_file_absent "$ROOT"/.worktrees/$WT1/file2
+    assert_file_absent "$ROOT"/.worktrees/$WT1/file3
+    assert_file_exists "$ROOT"/.worktrees/$WT2/file
+    assert_file_exists "$ROOT"/.worktrees/$WT2/file2
+    assert_file_absent "$ROOT"/.worktrees/$WT2/file3
+    assert_file_exists "$ROOT"/.worktrees/$WT3/file
+    assert_file_exists "$ROOT"/.worktrees/$WT3/file2
+    assert_file_exists "$ROOT"/.worktrees/$WT3/file3
+
+    # wait for 4 seconds
```
thockin, line 642: A better comment would be "why" - Wait until the oldest WT expires

test_e2e.sh
```diff
@@ -576,6 +576,238 @@ function e2e::worktree_cleanup() {
...
+    assert_file_absent "$ROOT"/.worktrees/$WT1/file2
+    assert_file_absent "$ROOT"/.worktrees/$WT1/file3
+    assert_file_exists "$ROOT"/.worktrees/$WT2/file
+    assert_file_exists "$ROOT"/.worktrees/$WT2/file2
+    assert_file_absent "$ROOT"/.worktrees/$WT2/file3
+    assert_file_exists "$ROOT"/.worktrees/$WT3/file
+    assert_file_exists "$ROOT"/.worktrees/$WT3/file2
+    assert_file_exists "$ROOT"/.worktrees/$WT3/file3
+
+    # wait for another 2 seconds
```
thockin, line 661: wait for another WT to expire

test_e2e.sh
```diff
@@ -576,6 +576,238 @@ function e2e::worktree_cleanup() {
...
+function e2e::stale_worktree_timeout_restart() {
+    echo "$FUNCNAME 1" > "$REPO"/file
+    git -C "$REPO" commit -qam "$FUNCNAME"
+    WT1=$(git -C "$REPO" rev-list -n1 HEAD)
+    GIT_SYNC \
+        --period=100ms \
+        --repo="file://$REPO" \
+        --root="$ROOT" \
+        --link="link" \
+        --stale-worktree-timeout="5s" \
```
thockin, line 693: don't need this arg on the first run

test_e2e.sh
```diff
@@ -576,6 +576,238 @@ function e2e::worktree_cleanup() {
...
+        --one-time
+
+    # wait for first sync
+    wait_for_sync "${MAXWAIT}"
```
thockin, line 697: no need to wait for sync - the command above won't exit until it is synced

thockin (commented):
I'm afraid the timing on the sleeps will be flaky, espcially in the second test - did you see it flaking?  E.g. if docker takes just a little to long to start the next run?

cmd/git-sync/main.go
sviscaino: I agree it seems cleaner, I applied your suggestion

sviscaino:
Rebased and comments addressed :)

sviscaino:
> I'm afraid the timing on the sleeps will be flaky, espcially in the second test - did you see it flaking?  E.g. if docker takes just a little to long to start the next run?

I hadn't observed any flakiness locally, but I doubled the delays just in case

cmd/git-sync/main.go
```diff
@@ -475,19 +480,20 @@ func (abs absPath) Base() string {
...
 // repoSync represents the remote repo and the local sync of it.
 type repoSync struct {
-	cmd        string         // the git command to run
-	root       absPath        // absolute path to the root directory
-	repo       string         // remote repo to sync
-	ref        string         // the ref to sync
-	depth      int            // for shallow sync
-	submodules submodulesMode // how to handle submodules
-	gc         gcMode         // garbage collection
-	link       absPath        // absolute path to the symlink to publish
```
thockin, line 485: was this a merge mistake?

thockin: I think this is the only remaining - looks great otherwise.  Thanks for the PR!

sviscaino: I just added `staleTimeout time.Duration` at the end, and re-aligned the spacing of the types.

sviscaino: Oops, fixed!

thockin (commented):
Thanks!

/lgtm
/approve

PR #716, opened 2023-04-28
Move worktrees to .worktrees/*

thockin:
This will make it easier to enumerate old worktrees and do better cleanup.


thockin:
self-merge to make progress.

PR #717, opened 2023-04-28
e2e: handle log capture automatically

thockin:
Same as test_git.sh


thockin:
self-merge to make progress.

PR #718, opened 2023-04-28
Back-compat for --v usage

thockin:


thockin:
self-merge because it's simple, and to make progress.

PR #719, opened 2023-04-28
Add a doc about changing from v3 to v4

thockin:


thockin:
self-merge because it's just docs, and to make progress.

PR #720, opened 2023-04-28
Make --webhook-success-status=0 the same as -1

thockin:


thockin:
self-merge because it's simple, and to make progress.

PR #721, opened 2023-04-28
Makefile: set GOOS,GOARCH defaults to kill warning

thockin:


thockin:
self-merge because it is trivial

PR #722, opened 2023-04-29
Make all env var names GITSYNC_FOO

thockin:
Still support GIT_SYNC_FOO for compat.


thockin:
Self-merge for progress.

PR #725, opened 2023-05-04
Capture and simplify a git.Run() method

thockin:


thockin:
self-merge for progress

PR #729, opened 2023-05-05
Fetch just once per run and when hash changes

thockin:
Also add metrics and tests to prove we are not in a fetch loop!


thockin:
self-merge to make progress

PR #730, opened 2023-05-05
Update to go 1.20

thockin:


nan-yu (commented):
The go version in https://github.com/kubernetes/git-sync/blob/master/go.mod#L21 also needs to be updated.

thockin:
oops, done

nan-yu (approved):
/lgtm

PR #733, opened 2023-05-05
Prevent git's 'dubious ownership' error (on v4 branch)

thockin:


thockin:
self-merge to make progress

PR #735, opened 2023-05-07
Support in-place updates from v3

thockin:
A few small accomodations and we can clean up properly.


thockin:
Self-merge for progress

PR #736, opened 2023-05-07
Reorganize and rename deprecated flags vars

thockin:


thockin:
Self-merge for progress

PR #741, opened 2023-05-11
Get rid of "must not start with ." logic

thockin:
It serves very little purpose and isn't even correct as-is.


thockin:
self-merge

PR #753, opened 2023-06-09
Make the container image smaller

thockin:
For each package and binary we need, this pulls in all the files and deps (shared libs, mostly).  The build is slower but the final image is 85 MB (versus 157 MB before).  e2e passes.

This is based on scripts used in kubernetes and KinD.

### DIFF
--- a/stage_binaries.sh
+++ b/stage_binaries.sh
@@ -1,0 +1,278 @@
+#!/bin/bash
+
+# Copyright 2022 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# USAGE: stage-binaries.sh -o <staging-dir> ( -p <package> | -b binary )..."
+#
+# Stages all the packages or files and their dependencies (+ libraries and
+# copyrights) to the staging dir.
+#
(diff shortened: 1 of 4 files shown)


stage_binaries.sh
```diff
@@ -0,0 +1,278 @@
...
+        cp -a --parents "${copyright}" "${staging}"
+    fi
+
+    # stage the package status mimicking bazel
+    # https://github.com/bazelbuild/rules_docker/commit/f5432b813e0a11491cf2bf83ff1a923706b36420
+    # instead of parsing the control file, we can just get the actual package status with dpkg
+    dpkg -s "${package}" > "${staging}/var/lib/dpkg/status.d/${package}"
+}
+
+# helper because bash sometimes barfs on embedded comments
```
BenTheElder, line 95: nit: I can't really tell from this comment and the usage context why this is needed. At the call site it's also not obvious what's happening without going back and reading this.

BenTheElder (commented):
one nit :-)

stage_binaries.sh
thockin: grep treats "no match" as an error, which trips up `pipefail` and `errexit`

```
$ echo foo | grep bar; echo $?
1
```

So we want to do:

```
$ echo foo | { grep bar || true; }; echo $?
0
```

Adding back the comments in the style of `binary_to_libraries()`:

```
$ echo foo \
    `# comment1` \
    | { grep "bar" || true; } \
    `# comment2`; echo $?
bash: syntax error near unexpected token ``# comment2`'
```
Parens fails the same way.

It works without the braces and `|| true`, but returns error status

```
$ echo foo     `# comment1`     | grep "bar"     `# comment2`; echo $?
1
```

Once bash parses `{}` or `()` it can't handle subsequent `#comment` blocks.  `happy()` hides that.  I could rename it :)

thockin: cute, I guess I should actually read the man page :)

thockin: PTAL.  I want to run a scanner against it to be sure it works, when I get a moment

BenTheElder (commented):
/lgtm

PR #764, opened 2023-07-06
Update to use bookworm base image

thockin:
This include symlinks from /bin -> /usr/bin (and lib, and ...), which broke the build script.  See comments in there for details.


thockin:
Hmm, unexpected e2e fails.

thockin:
Harumph, this expects /bin to be a symlink, but distroless has a real dir there, resulting in a collision.  Not sure how to fix it.  Will have to think.

thockin:
Passes e2e now, final image size is 63.7 MB.  Far too manual, but `<shrug>`

PR #778, opened 2023-07-25
Handle errors from credential refresh

thockin:

Now the error indicates the credential refresh as the problem, and either does not try to sync (if no creds) or tries to use previous creds (if they were fetched at some point).

Replaces #774

### DIFF
--- a/main.go
+++ b/main.go
@@ -944,8 +944,8 @@ func main() {
 			}
 		}
 		if *flAskPassURL != "" {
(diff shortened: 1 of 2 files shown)


thockin:
Need to adapt this to v3 (https://github.com/kubernetes/git-sync/pull/775)

karlkfi (approved):
/lgtm

PR #782, opened 2023-07-28
Handle errors from credential refresh (v3)

thockin:

Now the error indicates the credential refresh as the problem, and either does not try to sync (if no creds) or tries to use previous creds (if they were fetched at some point).

Replaces #775


karlkfi (approved):
/lgtm

PR #784, opened 2023-07-29
Remove trailing newlines from errors

justinsb:
I believe we print a new line anyway from e.g. envInt, which calls fmt.Fprintln.

Found by staticcheck:

main.go:127:17: error strings should not end with punctuation or newlines (ST1005)
main.go:157:13: error strings should not end with punctuation or newlines (ST1005)
main.go:187:13: error strings should not end with punctuation or newlines (ST1005)
main.go:217:13: error strings should not end with punctuation or newlines (ST1005)

### DIFF
--- a/main.go
+++ b/main.go
@@ -124,7 +124,7 @@ func envBoolOrError(def bool, key string, alts ...string) (bool, error) {
 		if err == nil {
 			return parsed, nil
 		}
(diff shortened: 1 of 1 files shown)


thockin (commented):
Thanks!

/lgtm
/approve

PR #785, opened 2023-07-29
lint: break out of loop, not just out of select statement

justinsb:
Caught by staticcheck:

main.go:2083:5: ineffective break statement. Did you mean to break out of the outer loop? (SA4011)

### DIFF
--- a/main.go
+++ b/main.go
@@ -2077,10 +2077,11 @@ func parseGitConfigs(configsFlag string) ([]keyVal, error) {
 	ch := make(chan rune)
 	stop := make(chan bool)
 	go func() {
+	loopOverConfigFlags:
 		for _, r := range configsFlag {
 			select {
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -2077,10 +2077,11 @@ func parseGitConfigs(configsFlag string) ([]keyVal, error) {
 	ch := make(chan rune)
 	stop := make(chan bool)
 	go func() {
+	loopOverConfigFlags:
```
thockin, line 2080: yes, BUT....this whole function is borked.  We never send stop, so we leave a dangling goroutine.  I only vaguely recall this code, and frankly, I am not sure what I was thinking.  I think this can be much simpler.

Adding leak detection to the test proves it.

I'm going to close this and send a different one which doesn't leak.

PR #786, opened 2023-07-29
Enable github actions for CI

justinsb:
Makes it easy for anyone to run tests on github, doesn't preclude us using prow also.

Second commit fixes an issue where we were relying on global git configuration, which didn't work in the isolated github CI environment.

Issue #424

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,48 @@
+---
+name: CI
+
+'on':
+  - push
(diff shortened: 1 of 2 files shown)


justinsb:
Can be previewed at https://github.com/justinsb/git-sync/actions

thockin (commented):
We should also turn on dependabot :)

/lgtm
/approve

Issue #777, opened 2023-07-24
Should we enable more linting?


justinsb:
Inspired by #775 I was wondering why the linter didn't catch the error (or warn us about it).  Turns out checking for unhandled errors [does not have consensus](https://github.com/golang/go/issues/4045), and is thus not implemented in `go vet`.

But we could turn on `golangci-lint` - other k8s subprojects do use it.

Here's the current output with the default configuration:

```
git-sync  (master) > golangci-lint run
main.go:480:34: Error return value of `pflag.CommandLine.MarkDeprecated` is not checked (errcheck)
        pflag.CommandLine.MarkDeprecated("branch", "use --ref instead")
                                        ^
main.go:483:34: Error return value of `pflag.CommandLine.MarkDeprecated` is not checked (errcheck)
        pflag.CommandLine.MarkDeprecated("change-permissions", "use --group-write instead")
                                        ^
main.go:486:34: Error return value of `pflag.CommandLine.MarkDeprecated` is not checked (errcheck)
        pflag.CommandLine.MarkDeprecated("dest", "use --link instead")
                                        ^
main.go:1726:14: Error return value is not checked (errcheck)
        refreshCreds(ctx)
                    ^
main.go:1895:16: Error return value of `io.WriteString` is not checked (errcheck)
        io.WriteString(h, s)
                      ^
main_test.go:750:7: Error return value is not checked (errcheck)
        touch(dirPath)
             ^
main_test.go:759:7: Error return value is not checked (errcheck)
        touch(filePath)
             ^
main.go:1933:13: S1039: unnecessary use of fmt.Sprintf (gosimple)
                sshCmd += fmt.Sprintf(" -o StrictHostKeyChecking=no")
                          ^
main.go:2087:3: S1023: redundant `return` statement (gosimple)
                return
                ^
main.go:1152:33: SA1016: os.Kill cannot be trapped (did you mean syscall.SIGTERM?) (staticcheck)
        signal.Notify(c, os.Interrupt, os.Kill)
```

justinsb:
Looking at the list the only one I think is important is the one @karlkfi identified (refreshCreds).  The `touch` errcheck warning is test code.

This is just the default set of linters, there could well be others in the not-enabled-by-default linters.  If there are any recent bugs (that a linter could have caught) I can see if they would have been caught.

thockin:
In short, yes, I'd be happy with more lint-free code.

The `signal` one seems like a mistake, and `touch` is suspect to me.

PR #787, opened 2023-07-29
Add linters to makefile/github actions

justinsb:
Builds on #786, adds a linter task to the makefile and a task to the github actions CI.

I think if we merge #786, then this will run in github actions CI.

Issue #777

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -46,3 +46,30 @@ jobs:
         working-directory: git-sync
         run: |
           make test
+
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '1.20.6'
+
+      - uses: actions/checkout@v3
+        with:
+          path: git-sync
+
+      # We run the github action, even though this is duplicated in `make lint` below.
+      # This is because the action gives easier-to-read output than the linter.
+      # There is a risk of drift between the two, but this is only linting,
+      # not runtime correctness!
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v3
+        with:
+          working-directory: git-sync
+          version: v1.53.3
+
+      - name: make lint
+        working-directory: git-sync
+        run: |
+          make lint
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,0 +1,26 @@
+# This file configures checks that all new code for Kubernetes is meant to
+# pass, in contrast to .golangci.yaml which defines checks that also the
+# existing code passes.
+
+run:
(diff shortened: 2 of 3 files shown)


justinsb:
Note this won't pass yet, because we need to fix the issues (hence the stream of small PRs).

/hold

.github/workflows/main.yml
```diff
@@ -46,3 +46,26 @@ jobs:
...
+        uses: actions/setup-go@v4
+        with:
+          go-version: '1.20.6'
+
+      - uses: actions/checkout@v3
+        with:
+          path: git-sync
+
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v3
```
thockin, line 63: Should we copy the k/k .golangci-lint config?

.github/workflows/main.yml
```diff
@@ -46,3 +46,26 @@ jobs:
...
+        with:
+          working-directory: git-sync
+          version: v1.53.3
+
+      - name: make lint
```
thockin, line 68: why make lint AND golangci-lint-action ?

thockin:
#786 merged

.github/workflows/main.yml
justinsb: Ah, I'll add a comment.  It's because the github action is supposed to format the output such that github will then put comments onto the right lines of the PR.  So I added it first.  But generally I think we want to be able to run things outside of github actions, hence I consider it a more friendly first-pass on the "canonical test", which is `make lint`.

Of course we do now (potentially) have two sources of truth, and some linter warnings (the ones from gostaticcheck) will be presented in the less friendly way.  So not sure whether this is a win or not.  I figured we could try it, but I'm flexible.  WDYT?

.github/workflows/main.yml
justinsb: Sure, I can pull it in :-)  Might as well have a consistent style.

.github/workflows/main.yml
thockin: I am fine with `make lint`, but iwhy would they produce different results?

thockin (commented):
Thanks!

/lgtm
/approve

.github/workflows/main.yml
justinsb: Well `make lint` runs two linters, and the github action form for the second linter didn't look as useful.

But my bigger concern is that while I think they do the same thing today, as we add more options and evolve, drift is possible.  If we weren't "wearing two watches", drift wouldn't be possible.

It's probably a balancing act between how helpful the "pretty" warnings are vs the drift risk over time.  And I suspect we can just evaluate that from experience.  Thanks for lgtm :-)

PR #788, opened 2023-07-29
Fix goroutine leak in git option parsing

thockin:
Alternative to #785

k/k uses goleak, too, so it seemed a reasonable dep.


justinsb:
LGTM and avoids the "goto" statement that I proposed - thanks.

Not sure why we're using channels here at all TBH, but .. lgtm

/approve
/lgtm

thockin:
The channel is just a simple way of consuming an input string.  We could have saved an index, just as easily.

PR #791, opened 2023-07-29
Rename github build workflow

thockin:
It says amd64, but calls all-build

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,7 +13,7 @@ permissions:
   contents: read
 
 jobs:
-  build-linux-amd64:
+  build:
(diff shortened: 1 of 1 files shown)


justinsb:
Good catch & I really like this.  There's a choice in how we name the github actions "jobs", and I have previously proposed mapping them 1:1 to scripts under (for example) `dev/ci` - with the intention that then it becomes easy to run locally when needed.  But if we're using `make` it makes a lot more sense to align them with the make targets.

(We might want to consider naming the thing which calls `make all-build` -> `all-build`, therefore, but now I'm just nit-picking)

/approve
/lgtm

PR #792, opened 2023-07-29
lint: fix remaining lint issues.

justinsb:
These fix the issues identified by the k/k inspired linter
configuration that we will be adding:

```
pkg/pid1/pid1.go:72:14: ST1005: error strings should not end with punctuation or newlines (stylecheck)
                        return 0, fmt.Errorf("unhandled exit status: 0x%x\n", status)
                                  ^
pkg/pid1/pid1.go:86:21: ST1005: error strings should not end with punctuation or newlines (stylecheck)
                        return false, 0, fmt.Errorf("wait4(): %w\n", err)
                                         ^
main.go:480:34: Error return value of `pflag.CommandLine.MarkDeprecated` is not checked (errcheck)
        pflag.CommandLine.MarkDeprecated("branch", "use --ref instead")
                                        ^
main.go:483:34: Error return value of `pflag.CommandLine.MarkDeprecated` is not checked (errcheck)
        pflag.CommandLine.MarkDeprecated("change-permissions", "use --group-write instead")
                                        ^
main.go:486:34: Error return value of `pflag.CommandLine.MarkDeprecated` is not checked (errcheck)
        pflag.CommandLine.MarkDeprecated("dest", "use --link instead")
                                        ^
main.go:1897:16: Error return value of `io.WriteString` is not checked (errcheck)
        io.WriteString(h, s)
                      ^
main.go:555:2: ifElseChain: rewrite if-else to switch statement (gocritic)
        if *flDeprecatedBranch != "" && (*flDeprecatedRev == "" || *flDeprecatedRev == "HEAD") {
        ^
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -477,31 +477,31 @@ func main() {
 	// Obsolete flags, kept for compat.
 	flDeprecatedBranch := pflag.String("branch", envString("", "GIT_SYNC_BRANCH"),
 		"DEPRECATED: use --ref instead")
-	pflag.CommandLine.MarkDeprecated("branch", "use --ref instead")
+	mustMarkDeprecated("branch", "use --ref instead")
 	flDeprecatedChmod := pflag.Int("change-permissions", envInt(0, "GIT_SYNC_PERMISSIONS"),
 		"DEPRECATED: use --group-write instead")
-	pflag.CommandLine.MarkDeprecated("change-permissions", "use --group-write instead")
+	mustMarkDeprecated("change-permissions", "use --group-write instead")
 	flDeprecatedDest := pflag.String("dest", envString("", "GIT_SYNC_DEST"),
 		"DEPRECATED: use --link instead")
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -552,17 +552,19 @@ func main() {
 		handleConfigError(log, true, "ERROR: --repo must be specified")
 	}
 
-	if *flDeprecatedBranch != "" && (*flDeprecatedRev == "" || *flDeprecatedRev == "HEAD") {
+	switch {
```
justinsb, line 555: I'm not sure I personally agree with this linter, but it is based on https://go.dev/doc/effective_go#switch (according to https://github.com/go-critic/go-critic/issues/453)

thockin: Yeah, I don't agree with this guidance as dogmatic.  I prefer the else-if but it's fine.  not a big deal if it shuts up the tool

thockin (commented):
Thanks!

/lgtm
/approve

PR #798, opened 2023-08-30
fix: Fix backward compatibility of environment variables with bool type

### LINKED ISSUE #797: Fix backward compatibility of environment variables with bool type
Boolean type environment variable arguments such as GITSYNC_ONE_TIME and GITSYNC_ADD_USER are not backward compatible.

According to the documentation(https://github.com/kubernetes/git-sync/blob/master/v3-to-v4.md), backward compatibility of Version 3's GIT_SYNC_ -> GITSYNC_ should also be supported for boolean type.

phil-park:
Fixed a bug in the code that selects keys from environment variables.

Which issue(s) this PR fixes:

Fixes #797

### DIFF
--- a/main.go
+++ b/main.go
@@ -131,7 +131,7 @@ func envBoolOrError(def bool, key string, alts ...string) (bool, error) {
 		return parse(val)
 	}
 	for _, alt := range alts {
-		if val := os.Getenv(key); val != "" {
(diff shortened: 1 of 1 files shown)


thockin (commented):
Thanks!

/lgtm
/approve

PR #821, opened 2023-09-22
Skip removing the error file when removing the former gitRoot dir

nan-yu:
If `git clone` fails, it removes all the content in the gitRoot director, including the error file, and then re-run `git clone`.

The presence of an error file indicates a prior failure. If `git clone` is slow, the fact that the error file is deleted may incorrectly suggest the previous clone was successful.

This commit skips deleting the error file when `git clone` fails. If it fails again, the error file will be overwriten.


thockin (commented):
If the error file is in the `--root`, we can never call `git clone` again - it won't run unless the directory is empty.

Am I missing something?

If your e2e had a second git-sync which is expected to pass, I think it will fail.

nan-yu:
> If the error file is in the `--root`, we can never call `git clone` again - it won't run unless the directory is empty.
> 
> Am I missing something?
> 
> If your e2e had a second git-sync which is expected to pass, I think it will fail.

Oh, right, `git clone` requires the destination to be empty. 😞  I think the new test didn't fully cover the scenario.

I updated the PR to add a subdirectory under --root, so we can delete the subdirectory and keep the error file if `git clone` fails.
I also refined the e2e test to check the error file changes with continuous git-sync loops.

thockin (commented):
This is now a substantial change to the way git-sync works, done on a "maintenance" branch.  I'm afraid of what side-effects it might have (knowing that e2e never covers every use case).

The simpler answer would be to not write the error into --root, wouldn't it?

Before considering making such a change to v3, can we evaluate what v4 (master branch) would look like?  It doesn't use git-clone anymore - it uses git-fetch, which I think will not have this problem.

nan-yu:
> The simpler answer would be to not write the error into --root, wouldn't it?

I considered that solution. If --error-file is not under --root, then it needs to be an absolute path, instead of relative to --root. Given that the full path is provided by users, it is likely the absolute path is still under --root. We'll need to add validation and error handling for --error-file.

Issue #841, opened 2023-11-07
git-sync v4 fetches the HEAD of the remote default branch instead of `master` if --ref is not specified


nan-yu:
This is a breaking change in git-sync v4.

In git-sync v3, if neither `--branch` or `--rev` is specified, it fetches the HEAD commit on the default `master` branch.

In git-sync v4, if `--ref` is not specified, if fetches the HEAD commit on the remote default branch, which may not be `master`. 
Upgrading from v3 to v4 will cause different sets of contents being selected, which will lead to cascading issues (e.g. unexpected resource deletion in Config Sync).

thockin:
Fun. The rationale for changing this was that most new fit repos are switching from "master" to "main" as the default branch.

The reality is that branch and rev are redundant - it came from a lack of understanding of how git works.

I will have to think hard about this.  If you specify neither flag, is the default branch and unreasonable default?

nan-yu:
I think using the remote default branch makes sense if neither flag is specified, but it behaves differently from v3.
v3 forces a default which may not be the same as remote default.

Unlike #840, this change won't return any error, but cause a different set of contents being selected silently.

thockin:
So, on one hand this is simpler and it is a major version update, so people SHOULD beware.

On the other hand, the failure mode could be nasty.

Is it the case that people will blindly update from v3 to v4 in prod, without qualifying it? Is there no way to force them to pay attention?

In hindsight, maybe having a default for --ref is a mistake, and we should force people to say "master" or "HEAD", but that would be ANOTHER breaking change, meaning v5.

Issue #840, opened 2023-11-07
git-sync v4 no longer supports abbreviated commit SHA


nan-yu:
In git-sync v3, `--rev` can point to an abbreviated commit SHA. `git rev-parse <SHORT SHA>` works because it clones the repo first. Then `git rev-parse` is able to parse the full commit hash from the short version.

However, in git-sync v4, `git rev-parse` is not able to resolve the short commit SHA.

thockin:
This is true.  I should document that.  That said, how important is this?  To use short SHAs, I have to first get all SHAs and use that to search locally (I could not find a way to do remote listing of short SHA), which makes everything slower and more costly.

For example, if I added a fallback to fetch all SHAs if the --rev flag isn't found, is that really worth the effort vs. telling people to use full SHAs?

nan-yu:
> That said, how important is this?

It's more about avoiding breaking changes to customers. Config Sync uses git-sync. With this change, the upgrade process won't be seamless. They have to update to full SHAs.

I think this seems to be okay if we document it as a breaking change. If they use short SHAs, they only get a git-sync error, but their resources won't be accidentally deleted.

There is another breaking change seems to be more serious. I just filed https://github.com/kubernetes/git-sync/issues/841 to track it separately.

thockin:
I found a similar issue: https://github.com/actions/checkout/issues/265

I have not yet found a way to support short SHAs

PR #843, opened 2023-11-11
Update the v3-to-v4 doc to cover some changes that I missed

thockin:
xref #840 and #841

I will still try to find better answers, but document them for now.


nan-yu (commented):
/lgtm

PR #849, opened 2023-12-09
WIP: Make the local repo a subdir of the root, rather than the root itself

### LINKED ISSUE #846: Consider putting the local repo in a subdir of the --root
This makes it easier to use a filesystem as `--root`.

I did a simple hack and the e2e's "just work".  It does change an unstated contract, specifically that the --root is the repo, and git operations can be run in there.  I'm not sure if that matters or not

thockin:
Fixes #846


thockin:
FTR I am holding this until/unless I have time to spend on a v5 with some.other features I would like to see.

PR #862, opened 2024-02-08
chore: fix missing unit duration in the k8s example

jmichalek132:
Using the example the git-sync sidecar exits and logs:
```
INFO: detected pid 1, running init handler
invalid argument "60" for "--period" flag: time: missing unit in duration "60"
Usage of /git-sync:
```


thockin:
Thanks!

/lgtm
/approve

PR #868, opened 2024-04-04
fix(auth): Fix ssh authentication

yyvess:
## Ssh authentification 
Fix ssh authentification break by https://github.com/kubernetes/git-sync/commit/d5c61f7ad19dea6ae6938dcb5dea7daa0b54676b

User must not be extracted from ssh as ssh://git@github.com/kubernetes/git-sync.git

PR #869, opened 2024-04-04
Add e2e for ssh:// URLs

thockin:
Followup to #868 from @yyvess 

The test failed before #868 and passes now

### DIFF
--- a/main.go
+++ b/main.go
@@ -572,7 +572,9 @@ func main() {
 	if *flUsername == "" {
 		// username and user@host URLs are validated as mutually exclusive
 		if u, err := url.Parse(*flRepo); err == nil { // it may not even parse as a URL, that's OK
-			if u.User != nil && u.Scheme != "ssh" {
+			// Note that `ssh://user@host/path` URLs need to retain the user
+			// field. Out of caution, we only handle HTTP(S) URLs here.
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -572,7 +572,9 @@ func main() {
 	if *flUsername == "" {
 		// username and user@host URLs are validated as mutually exclusive
 		if u, err := url.Parse(*flRepo); err == nil { // it may not even parse as a URL, that's OK
-			if u.User != nil && u.Scheme != "ssh" {
+			// Note that `ssh://user@host/path` URLs need to retain the user
+			// field. Out of caution, we only handle HTTP(S) URLs here.
+			if u.User != nil && (u.Scheme == "http" || u.Scheme == "https") {
```
yyvess, line 577: Why not keep test on schema ssh here ?
Note that for url  format `git@github.com:kubernetes/git-sync` method `url.Parse` return an err.

thockin: I don't know what OTHER schemes might have the problem that `ssh://...` has, but I *do* know that this logic was only intended for HTTP and HTTPS.  I'd rather fall back on not changing the URL than changing it wrong (as I did with SSH)

yyvess (approved):

PR #871, opened 2024-04-10
Bump base bookworm image to 1.0.2

dims:
Updated golang to 1.22 as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,13 +45,13 @@ ALL_PLATFORMS := linux/amd64 linux/arm linux/arm64 linux/ppc64le linux/s390x
 OS := $(if $(GOOS),$(GOOS),$(shell go env GOOS))
 ARCH := $(if $(GOARCH),$(GOARCH),$(shell go env GOARCH))
 
-BASEIMAGE ?= registry.k8s.io/build-image/debian-base:bookworm-v1.0.0
+BASEIMAGE ?= registry.k8s.io/build-image/debian-base:bookworm-v1.0.2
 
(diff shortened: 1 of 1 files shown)


rikatz:
/lgtm

on my lunch time and saw on twitter :P

norrs (approved):
This looks like a very sane thing to do :-) 
Hopefully latest debian-base contains the security fixes.

thockin:
Thanks!

/lgtm
/approve

norrs:
@thockin :  My idea was that this would fix https://security-tracker.debian.org/tracker/CVE-2023-45853 , but reading more in details in https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1054290 it seems like even tho zlib package itself does not build the minizip binary, it contains the sources for it that other source packages could vendor this again from zlib. In the bug report mentioned below, it seems like other packages have actually done this. 😅  

I believe maybe https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1054290#64 has released a fix for debian unstable and trixie (current what is defined as `debian next` upcoming stable release). But current stable debian bookworm is not currently patched with a backport of the fix. 

Should we maybe request a backport upstream in debian for bookworm? Because currently the security-tracker has the following note: `[bookworm] - zlib <ignored> (contrib/minizip not built and producing binary packages)`, so I'm a bit unsure if this will get a follow up if we don't ask about it?

thockin:
We can ask - I am not an expert in this sort of assessment, so I usually defer to Debian baseimage owners

PR #874, opened 2024-05-04
Fix: Use hugo mod get

mattn:
Recently hugo does not support hugoThemes and some themes are already missing. In the latest hugo, themes are treated as modules. So this change fixes that themes are retrieved in realtime with hugo mod get.

### DIFF
--- a/demo/blog/go.mod
+++ b/demo/blog/go.mod
@@ -1,0 +1,3 @@
+module blog
+
+go 1.22.2
--- a/demo/hugo/Dockerfile
+++ b/demo/hugo/Dockerfile
@@ -1,4 +1,4 @@
-# Copyright 2014 Google Inc. All rights reserved.
+# Copyright 2024 Google Inc. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


demo/hugo/Dockerfile
```diff
@@ -1,4 +1,4 @@
-# Copyright 2014 Google Inc. All rights reserved.
```
thockin, line 1: Please don't change copyright dates - the original copyright is correct.

demo/blog/go.mod
```diff
@@ -0,0 +1,3 @@
+module blog
```
thockin, line 1: Why do we need this file?

demo/hugo/run-hugo
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Copyright 2014 Google Inc. All rights reserved.
```
thockin, line 3: Don't change copyright dates

thockin (commented):
The README for this directory doesn't make a lot of sense to me.  It tells you to build images but then it loads a config which doesn't use those images.  Do you want to make this PR actually make the demo work, or just fix the hugo part and we can fix the instructions afterwards?

mattn:
Since hugo nowadays uses hugo mod get, it is no longer possible to bundle all themes as before. The theme list is already not complete and deprecated.

thockin:
I was staring at this and I realized I really have no desire to maintain a hugo blog (hence why this has been broken so long).

I am putting up #889 instead.  I appreciate the work on this, but I'm going to decline for now.

Issue #877, opened 2024-05-14
Support GitHub App authentication


risset:
Older issues on this topic:
https://github.com/kubernetes/git-sync/issues/450
https://github.com/kubernetes/git-sync/issues/769

Hey, I wanted to revive this discussion again since I have a use case that would benefit from using a GitHub app for auth, and using an operator like that github-app-operator isn't ideal for me right now.

Would the following design potentially work?
- allow github app private key, app ID and installation ID as parameters to git-sync
- if github app auth is enabled, in `refreshCreds`, request a new token (can check if it's expired first)
    - I think something similar to this should work: https://github.com/samirtahir91/github-app-operator/blob/0aca1da117933cad3075a55565eb422c0784f30c/internal/controller/githubapp_controller.go#L727
-  should then be able to authenticate over HTTPS with that token 

`CallAskPassURL` is doing something a bit like this already, but doesnt give you much control over the request it makes.

I'm not too familar with git-sync internals yet but I have plenty of time to implement and try this out if maintainers think it might work

thockin:
In this area, git-sync is not very complex.  Something like you describe sounds fine.  Ideally we would start with an e2e case, but it's not clear to me if there's an easy way to e2e reproduce the experience locally.  Is there a trivial github app that we can use to at least manually run tests?  Hard to support something I can't test at all :)

risset:
Yeah, testing this doesn't seem straightforward :(
I'm not sure of being able to use an existing public app for testing.
When creating a new app, GitHub gives you a private key which running applications can use along with the app ID and installation ID (tied to a repo or organization) to authenticate as that particular app installation. So whoever is running the test would probably have to create a dummy app that's installed to the repo being synced and provide those three values to the test, which isn't great.

thockin:
Can we automate or at least document clearly how I (an arbitrary person) could create my own "just for testing" app?  Or that we could set up thru GH actions to run, but maybe an arbitrary person cant?

risset:
I could create documentation for creating a testing app. Automation might be tricky since there's no GitHub API endpoints or the like for creating apps but it's worth looking into further

risset:
Update from me on this, I've got the change working now and there's a couple things I'm unsure about before opening a PR (should have that done by the weekend)

- Where should documentation for creating a github app for testing go? The docs directory seemed more user-facing
[...]

PR #878, opened 2024-05-19
Add support for GitHub app authentication

risset:
Related issue: https://github.com/kubernetes/git-sync/issues/877

Currently the `e2e::github_app_auth` test should work locally, if the correct env vars are set. I've tested the feature that way before creating the PR. Ideally it would be possible to run it in the GHA workflow. One way I thought that might be feasible is if an app was created and installed to a particular (private) repo, and then its private key was stored as a git-sync repository secret. 

Some information about GitHub app authentication: https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app


main.go
```diff
@@ -1851,6 +1878,72 @@ func (git *repoSync) CallAskPassURL(ctx context.Context) error {
...
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		_ = resp.Body.Close()
+	}()
+	if resp.StatusCode != 201 {
+		errMessage, err := io.ReadAll(resp.Body)
```
thockin, line 1919: Shouldn't we bound the read here to 1K or 4K or 32K or something?

main.go
```diff
@@ -1851,6 +1878,72 @@ func (git *repoSync) CallAskPassURL(ctx context.Context) error {
 	return nil
 }
 
+func (git *repoSync) RefreshAppToken(ctx context.Context, githubBaseURL, privateKey string, appID, installationID int) error {
```
thockin, line 1881: I'd like the entire set of functions to reflect "GitHubApp" rather than just app, so it's easier to read.  Could also do with a nice comment, please

main.go
```diff
@@ -254,6 +257,19 @@ func main() {
 		envString("", "GITSYNC_ASKPASS_URL", "GIT_SYNC_ASKPASS_URL", "GIT_ASKPASS_URL"),
 		"a URL to query for git credentials (username=<value> and password=<value>)")
 
+	flGithubAPIURL := pflag.String("github-api-url",
```
thockin, line 260: it's not really a URL is it?  Usually a URL includes "scheme" or "protocol", but later you prepend "https://".  Also - is this allowed to have paths?  E.g. could I say "notgithub.com/mytenant/path" ?

I propose to either make this a URL ("https://host:port/path") or rename it to "host" or "domain".

What do you think?

main.go
```diff
@@ -254,6 +257,19 @@ func main() {
...
+		envString("api.github.com", "GITHUB_API_URL"),
```
thockin, line 261: Is this env name special (e.g. used somewhere deep in a library)?  I don't think so, right?

All the othe env vars are named "GITSYNC_xyz", so these should probably be "GITSYNC_GITHUB_APP_xyz"

main.go
```diff
@@ -254,6 +257,19 @@ func main() {
...
+		"the API URL to use when making requests to GitHub when using app auth")
+	flGithubAppPrivateKey := pflag.String("github-app-private-key",
+		envString("", "GITHUB_APP_PRIVATE_KEY"),
+		"the GitHub app private key to use for GitHub app auth")
+	flGithubAppApplicationID := pflag.Int("github-app-application-id",
+		envInt(-1, "GITHUB_APP_APPLICATION_ID"),
```
thockin, line 267: why not 0 for default?  Is that a valid value?

main.go
```diff
@@ -254,6 +257,19 @@ func main() {
 		envString("", "GITSYNC_ASKPASS_URL", "GIT_SYNC_ASKPASS_URL", "GIT_ASKPASS_URL"),
 		"a URL to query for git credentials (username=<value> and password=<value>)")
 
+	flGithubAPIURL := pflag.String("github-api-url",
+		envString("api.github.com", "GITHUB_API_URL"),
+		"the API URL to use when making requests to GitHub when using app auth")
+	flGithubAppPrivateKey := pflag.String("github-app-private-key",
```
thockin, line 263: We need to validate these (see ~ L493).  Specifically:

* You can't use these *and* some other form of auth at the same time
* If user sets this flag, they also need to set the app ID and installation ID

main.go
```diff
@@ -789,6 +806,16 @@ func main() {
 			}
 			metricAskpassCount.WithLabelValues(metricKeySuccess).Inc()
 		}
+
+		if *flGithubAppPrivateKey != "" && *flGithubAppInstallationID != -1 && *flGithubAppApplicationID != -1 {
+			if git.appTokenExpiry.Before(time.Now()) {
```
thockin, line 811: Shouldn't there be some buffer here, so we don't fail TOCTOU ?  E.g. require max(20% of the token's lifespan, 30s) or something? 

I don't want tokens to be valid here but fail to sync 1 second later, and some of the ops we do can take O(seconds).

main.go
```diff
@@ -789,6 +806,16 @@ func main() {
...
+				if err := git.RefreshAppToken(ctx, *flGithubAPIURL, *flGithubAppPrivateKey, *flGithubAppApplicationID, *flGithubAppInstallationID); err != nil {
+					metricAskpassCount.WithLabelValues(metricKeyError).Inc()
```
thockin, line 813: I think we should have a new metric for this

main.go
thockin: Actually, we should not put private keys as flags, they are listable via /proc.  The --password flag is bad for the same reason. 

We should have a way to declare a non-flag which can be set by env but not via CLI.  I will think about that.  It's going to be a busy few weeks, so please hang tight.

 Also, we should be able to pass this in via a file, like `--password-file`.  For all of these, I'd like it to be possible to use a kubernetes secret and mount a volume, rather than flags/env vars.  I'll have to refresh myself on formatting, i think we can only demand one value per file.  So maybe we want an alternative `--github-app-config=<dir>` with documented files within it, e.g. `api-url` (optional), `private-key` (required), `application-id` (required), and `installation-id` (required).   That way a user could make a single k8s Secret object with those key, and use a Secret-type volume.

test_e2e.sh
```diff
@@ -3250,6 +3250,21 @@ function e2e::github_https() {
     assert_file_exists "$ROOT/link/LICENSE"
 }
 
+##############################################
+# Test github app auth
+##############################################
+function e2e::github_app_auth() {
```
thockin, line 3256: Put this with the other `e2e::auth_*` tests and name it `e2e::auth_github_app`.

thockin (commented):
I'll have to think more about this when I get a second.  It looks pretty clean overall.  

Also need to add flags to the manual (at the bottom of the file).

I'll think about how to make e2e tests optional.

main.go
risset: Yeah you're right that 

I think having it as an actual base URL is better, I'll make this change

main.go
risset: I wasn't 100% sure that it couldn't be a valid value, but now you mention it, it doesn't seem likely that the IDs would start at 0 anyway. I'll have them default to 0

main.go
risset: That makes sense

main.go
risset: > It's going to be a busy few weeks, so please hang tight.
No worries! thanks for taking the time for the detailed review, I will address the simple things you brought up up in the meantime

main.go
risset: That would make sense, should the same apply to the `CallAskPassURL` usage of `io.ReadAll`?

test_e2e.sh
risset: Done

main.go
risset: Done

main.go
```diff
@@ -495,6 +512,31 @@ func main() {
...
+			handleConfigError(log, true, "ERROR: --password-file may not be specified when --github-app-private-key-file is specified")
+		}
+	} else {
+		if *flGithubAppApplicationID != 0 {
+			handleConfigError(log, true, "ERROR: --github-app-application-id may only be specified when --github-app-private-key-file is specified")
+		}
+		if *flGithubAppInstallationID != 0 {
+			handleConfigError(log, true, "ERROR: --github-app-installation-id may only be specified when --github-app-private-key-file is specified")
+		}
+	}
```
nan-yu, line 538: This makes the Github app auth mutually exclusive with username/password. What about other auth types, like ssh,  cookiefile, or etc?

main.go
```diff
@@ -1851,6 +1904,83 @@ func (git *repoSync) CallAskPassURL(ctx context.Context) error {
...
+		IssuedAt:  jwt.NewNumericDate(now),
+		ExpiresAt: jwt.NewNumericDate(now.Add(10 * time.Minute)),
+	}
+
+	jwt, err := jwt.NewWithClaims(jwt.SigningMethodRS256, claims).SignedString(pkey)
+	if err != nil {
+		return err
+	}
+
+	url, err := url.JoinPath(githubBaseURL, fmt.Sprintf("installations/%d/access_tokens", installationID))
```
nan-yu, line 1935: The path should be `app/installations/%d/access_tokens`.

main.go
```diff
@@ -254,6 +257,19 @@ func main() {
 		envString("", "GITSYNC_ASKPASS_URL", "GIT_SYNC_ASKPASS_URL", "GIT_ASKPASS_URL"),
 		"a URL to query for git credentials (username=<value> and password=<value>)")
 
+	flGithubBaseURL := pflag.String("github-base-url",
+		envString("https://api.github.com/", "GITSYNC_GITHUB_BASE_URL", "GIT_SYNC_GITHUB_BASE_URL"),
```
nan-yu, line 261: These flags are new, so you don't need the alternative keys with the `GIT_SYNC` prefix for backward compatibility.

docs/dev/testing_github_app_auth.md
```diff
@@ -0,0 +1,34 @@
...
+4. Click on `Repository permissions` under `Permissions`, and set `Contents` to `Read-only`
+5. Click on `Create GitHub App` at the bottom of the page.
+6. You should be navigated to a new page with a `Registration successful. You must generate a private key in order to install your GitHub App.` message. Click on the `generate a private key` link, and then the `Generate a private key` button, and save it somewhere; it will be used to test the app authentication.
+7. Click on the `Install App` tab on the left, and then click on `Install` on the right.
+8. Select `Only select repositories`, and pick any private repository that contains a "LICENSE" file (may need to be created beforehand).
+
+## Step 2: Export the necessary environment variables
+
+The following environment variables are *required* to run the git-sync github app auth test:
+- `GITHUB_APP_PRIVATE_KEY`
```
nan-yu, line 19: These env variables need to have the `GITSYNC_` prefix.

docs/dev/testing_github_app_auth.md
```diff
@@ -0,0 +1,34 @@
...
+- `GITHUB_APP_APPLICATION_ID`
+- `GITHUB_APP_INSTALLATION_ID`
+- `GITHUB_APP_AUTH_TEST_REPO`
+
+### GITHUB_APP_PRIVATE_KEY
+Should have been saved when creating the app
+
+### GITHUB_APP_APPLICATION_ID
+The value after "App ID" in the app's settings page
```
nan-yu, line 28: Suggest adding the page URL: https://github.com/settings/apps/<unique_app_name>

main.go
nan-yu: +1 for using a single k8s Secret object with these keys and add validation for them. It would be much easier for the downstream service to provide these values.

nan-yu:
While setting up a Github app, I noticed instead of using App ID to get installation tokens, it also supports [using the Client ID](https://docs.github.com/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-json-web-token-jwt-for-a-github-app).

I'm wondering have you experimented with the client ID?

risset:
> While setting up a Github app, I noticed instead of using App ID to get installation tokens, it also supports [using the Client ID](https://docs.github.com/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-json-web-token-jwt-for-a-github-app).
> 
> I'm wondering have you experimented with the client ID?

I didn't know about this, thanks. It appears to be a recent feature: https://github.blog/changelog/2024-05-01-github-apps-can-now-use-the-client-id-to-fetch-installation-tokens/

So the App ID and Client ID can be used interchangeably in this case.

> The application ID is not being deprecated at this time, nor are their plans to remove it. However, compatibility with future features will rely on use of the client ID, so updating is recommended.

Should we have both the app ID and client ID as parameters and accept either of them, or just replace app ID with client ID?

main.go
risset: Thanks, addressed this

docs/dev/testing_github_app_auth.md
risset: These ones don't correspond to the actual application env vars though, they're the values that get passed in via flags to the `GIT_SYNC` function in the test

nan-yu:
> Should we have both the app ID and client ID as parameters and accept either of them, or just replace app ID with client ID?

I would vote for supporting both scenarios.

main.go
```diff
@@ -254,6 +257,19 @@ func main() {
 		envString("", "GITSYNC_ASKPASS_URL", "GIT_SYNC_ASKPASS_URL", "GIT_ASKPASS_URL"),
 		"a URL to query for git credentials (username=<value> and password=<value>)")
 
+	flGithubBaseURL := pflag.String("github-base-url",
+		envString("https://api.github.com/", "GITSYNC_GITHUB_BASE_URL"),
+		"the GitHub base URL to use when making requests to GitHub when using app auth")
+	flGithubAppPrivateKeyFile := pflag.String("github-app-private-key-file",
```
thockin, line 263: See #886 for how this can be offered both as an env ($GITSYNC_GITHUB_APP_PRIVATE_KEY) and as a file as you have here

main.go
risset: I think I omitted those because I didn't see them being checked in the existing validations, but I can add them sure

main.go
risset: Thanks, I'll take a look at that now :)

nan-yu:
Hi @risset , any recent update on addressing the comments in this PR?

I'm working on [Config Sync](https://cloud.google.com/kubernetes-engine/enterprise/config-sync/docs/overview). If possible, we would like to leverage this feature in our next minor release in July.

thockin:
Hi Nan,

Before we can merge this we need to make sure we can test it properly,
which involves a private key.  I have not had the time to play and see how
private it really needs to be, or if maybe we can get it to test via GH
actions.

If you have time, that's where we need to spend it.

risset:
> Hi @risset , any recent update on addressing the comments in this PR?

Hi, I will be working on this some more over the next few days! I've been having some issues with running `test_e2e.sh` on my current (macOS) machine, but I'm able to test my changes on another Linux machine at least.

nan-yu:
> > Hi Liam Wyllie , any recent update on addressing the comments in this PR?
> 
> Hi, I will be working on this some more over the next few days! I've been having some issues with running `test_e2e.sh` on my current (macOS) machine, but I'm able to test my changes on another Linux machine at least.

Thanks for the update. Did you run the e2e test with a public repo or a private repo?
I followed the instruction with a private repo, and git-sync returned a repo not found error. Not sure if that works for you.

risset:
> > > Hi Liam Wyllie , any recent update on addressing the comments in this PR?
> > 
> > 
> > Hi, I will be working on this some more over the next few days! I've been having some issues with running `test_e2e.sh` on my current (macOS) machine, but I'm able to test my changes on another Linux machine at least.
> 
> Thanks for the update. Did you run the e2e test with a public repo or a private repo? I followed the instruction with a private repo, and git-sync returned a repo not found error. Not sure if that works for you.

I've been testing with a private repo. I'll double check the instructions and how I've set the app up etc. in case I missed something

risset:
> > > > Hi Liam Wyllie , any recent update on addressing the comments in this PR?
> > > 
> > > 
> > > Hi, I will be working on this some more over the next few days! I've been having some issues with running `test_e2e.sh` on my current (macOS) machine, but I'm able to test my changes on another Linux machine at least.
> > 
> > 
> > Thanks for the update. Did you run the e2e test with a public repo or a private repo? I followed the instruction with a private repo, and git-sync returned a repo not found error. Not sure if that works for you.
> 
> I've been testing with a private repo. I'll double check the instructions and how I've set the app up etc. in case I missed something

Does the app you created definitely have read-only or read-write access for `Contents` under **Repository permissions**, and is installed to the private repository itself (*only select repositories*) under **Install App**?

When I re-run the tests with either of those not true, I also get the repo not found error.

nan-yu:
> > > > > Hi Liam Wyllie , any recent update on addressing the comments in this PR?
> > > > 
> > > > 
> > > > Hi, I will be working on this some more over the next few days! I've been having some issues with running `test_e2e.sh` on my current (macOS) machine, but I'm able to test my changes on another Linux machine at least.
> > > 
> > > 
> > > Thanks for the update. Did you run the e2e test with a public repo or a private repo? I followed the instruction with a private repo, and git-sync returned a repo not found error. Not sure if that works for you.
> > 
> > 
> > I've been testing with a private repo. I'll double check the instructions and how I've set the app up etc. in case I missed something
> 
> Does the app you created definitely have read-only or read-write access for `Contents` under **Repository permissions**, and is installed to the private repository itself (_only select repositories_) under **Install App**?
> 
> When I re-run the tests with either of those not true, I also get the repo not found error.

No worries. I think it was the issue with my testing account. I created a robot Github account. The account is flagged and requires reinstatement.

It works with my personal account.

risset:
> No worries. I think it was the issue with my testing account. I created a robot Github account. The account is flagged and requires reinstatement.
> 
> It works with my personal account.

Nice! Hopefully we can get this over the line. I'm still not quite sure how getting the tests to run properly via actions will work.

nan-yu:
Regarding the e2e test for the github app integration, I created a robot GitHub account, a GitHub app, along with a private repository under the account. After that, I set up the required environment variables in GitHub Secret in my personal fork. All e2e test passed with changes in https://github.com/risset/git-sync/pull/1: https://github.com/nan-yu/git-sync/actions/runs/9947634472.

Can we use the [robot account ](https://github.com/git-sync-test-robot)for testing, @thockin ?

main.go
```diff
@@ -256,6 +259,22 @@ func main() {
...
+		envString("", "GITSYNC_GITHUB_APP_PRIVATE_KEY_FILE"),
+		"the file from which the private key for GitHub app auth will be sourced")
+	flGithubAppClientID := pflag.String("github-app-client-id",
+		envString("", "GTSYNC_GITHUB_APP_CLIENT_ID"),
```
thockin, line 269: Typo: GTSYNC should be GITSYNC.

Same on the next one, too

main.go
```diff
@@ -256,6 +259,22 @@ func main() {
 		envString("", "GITSYNC_ASKPASS_URL", "GIT_SYNC_ASKPASS_URL", "GIT_ASKPASS_URL"),
 		"a URL to query for git credentials (username=<value> and password=<value>)")
 
+	flGithubBaseURL := pflag.String("github-base-url",
+		envString("https://api.github.com/", "GITSYNC_GITHUB_BASE_URL"),
+		"the GitHub base URL to use when making requests to GitHub when using app auth")
+	flGithubAppPrivateKeyFile := pflag.String("github-app-private-key-file",
```
thockin, line 265: Originally this was exposing the key as a flag.  Now with envy bars it can be done safely (see flPassword above)

main.go
thockin: Yeah, this is a mess.  Was OK when it was 3 mutually exclusive modes, but it's getting out of hand.  OK for now, but we should plan a followup which is more declarative feeling and handles the mutual exclusion more automatically.

main.go
thockin: This should be a different metric - it's not askpass

main.go
thockin: Missed this before, yeah, it should :)

main.go
```diff
@@ -2276,6 +2416,26 @@ OPTIONS
...
+            The file from which the private key to use for GitHub app
+            authentication will be read.
+
+    --github-app-installation-id <int>, $GITSYNC_GITHUB_APP_INSTALLATION_ID
+            The installation ID of the GitHub app used for GitHub app
+            authentication.
+
+    --github-app-application-id <int>, $GITSYNC_GITHUB_APP_APPLICATION_ID
+            The app ID of the GitHub app used for GitHub app authentication.
+            One of --github-app-application-id or --github-app-client-id is required.
```
thockin, line 2434: Can you put this same note on the client-id section, too?

thockin (commented):
Can you also update the README?  Basically just delete the last section, build the binary and run it with --man >> README.md, then add the last triple-ticks.

Other than my own automated and manual testing, it looks good.

risset:
Have updated to address recent feedback ! 

Pardon my ignorance here but I'm still a bit unsure on how best to address your comment on the usage of `io.ReadAll` to read the HTTP response body. Is reading with `io.Copy` to a `bytes.Buffer` preferable here, or is there a more efficient approach to that?

Also I've added one, but not sure if you feel a metric is actually needed for this feature, it was just sloppiness from me having the askpass code fragment left in there before ...

main.go
```diff
@@ -2492,6 +2665,22 @@ AUTHENTICATION
             When --cookie-file ($GITSYNC_COOKIE_FILE) is specified, the
             associated cookies can contain authentication information.
 
+    github app
+           When --github-app-private-key-file ($GITSYNC_GITHUB_APP_PRIVATE_KEY_FILE),
+           --github-app-application-id ($GITSYNC_GITHUB_APP_APPLICATION_ID)
```
nan-yu, line 2670: or --github-app-client-id

test_e2e.sh
```diff
@@ -295,6 +295,7 @@ function GIT_SYNC() {
         -v "$DOT_SSH/1/id_test":"/ssh/secret.1":ro \
         -v "$DOT_SSH/2/id_test":"/ssh/secret.2":ro \
         -v "$DOT_SSH/3/id_test":"/ssh/secret.3":ro \
+	-v "$(pwd)/$GITHUB_APP_PRIVATE_KEY_FILE":"/github_app_private_key.pem":ro \
```
nan-yu, line 298: This requires the github app private key file to be located at the same directory. It won't work with the e2e test in the GitHub Actions Workflow. 

I'm not sure if you get a chance to take a look at https://github.com/risset/git-sync/pull/1/commits/f02d659803e20a00f988d54c91e529dbd7daf9ca. Basically, it loads the private key from GitHub Secrets and stored it in a temp file. It would be good to incorporate it in this PR and enables the e2e test.

nan-yu (commented):
It looks mostly good to me. I would like to enable the e2e tests for this new feature.
Feel free to port over changes in https://github.com/risset/git-sync/pull/1/files.

main.go
```diff
@@ -71,13 +73,19 @@ var (
 		Name: "git_sync_askpass_calls",
 		Help: "How many git askpass calls completed, partitioned by state (success, error)",
 	}, []string{"status"})
+
+	metricRefreshGitHubAppTokenCount = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "git_sync_refresh_github_app_token_count",
+		Help: "How many times github app token was refreshed, partitioned by state (success, error)",
```
thockin, line 79: Super nit: "the GitHub app token"

main.go
```diff
@@ -838,6 +896,17 @@ func main() {
 			}
 			metricAskpassCount.WithLabelValues(metricKeySuccess).Inc()
 		}
+
+		if *flGithubAppPrivateKeyFile != "" && *flGithubAppInstallationID != 0 && (*flGithubAppApplicationID != 0 || *flGithubAppClientID != "") {
```
thockin, line 900: Should this be considering the newly added environment also?

main.go
```diff
@@ -1916,6 +1985,89 @@ func (git *repoSync) CallAskPassURL(ctx context.Context) error {
 	return nil
 }
 
+// RefreshGitHubAppToken generates a new installation token for a GitHub app and stores it as a credential
+func (git *repoSync) RefreshGitHubAppToken(ctx context.Context, githubBaseURL, privateKeyFile, clientID string, appID, installationID int) error {
+	git.log.V(3).Info("refreshing GitHub app token")
+
+	privateKeyBytes, err := os.ReadFile(privateKeyFile)
```
thockin, line 1992: Should this consider the environment var key?

main.go
thockin: You know what, this is fine.  I'm just being paranoid.  Doing a bounded copy is nice, but I won't require it here.

main.go
```diff
@@ -2276,6 +2428,27 @@ OPTIONS
...
+            The file from which the private key to use for GitHub app
+            authentication will be read.
+
+    --github-app-installation-id <int>, $GITSYNC_GITHUB_APP_INSTALLATION_ID
+            The installation ID of the GitHub app used for GitHub app
+            authentication.
+
+    --github-app-application-id <int>, $GITSYNC_GITHUB_APP_APPLICATION_ID
+            The app ID of the GitHub app used for GitHub app authentication.
+            One of --github-app-application-id or --github-app-client-id is required.
```
thockin, line 2446: ...when GitHub app authentication is used.

Same below

thockin (commented):
I am AFK for a couple weeks, and don't want to merge until I can run some.manual tests, but it looks great

test_e2e.sh
risset: I will take a look at that

nan-yu: Thanks. I updated https://github.com/risset/git-sync/pull/1 to make the github app auth test skippable.

main.go
risset: Forgot about this, added it now

test_e2e.sh
risset: Merged this in now

test_e2e.sh
```diff
@@ -194,6 +194,48 @@ fi
...
+  if [[ "${CLEANUP:-}" == 0 ]]; then
+      echo "leaving logs in $DIR"
+  else
+      rm -rf "$DIR"
+  fi
+}
+# Set the trap to call the final_cleanup function on exit.
+trap final_cleanup EXIT
+
+skip_github_app_test="${SKIP_GITHUB_APP_TEST:-false}"
```
sdowell, line 207: nit: I think skipping may be a more user friendly default, since in most cases a developer won't have github app credentials ready to use

test_e2e.sh
```diff
@@ -194,6 +194,48 @@ fi
...
+required_env_vars=()
+LOCAL_GITHUB_APP_PRIVATE_KEY_FILE="github_app_private_key.pem"
+GITHUB_APP_PRIVATE_KEY_MOUNT=""
```
sdowell, line 210: This should be declared as an empty array so that it is properly expanded below when the volume mount is not added.

e.g.
```bash
GITHUB_APP_PRIVATE_KEY_MOUNT=()
...
  # Mount the GitHub App private key file to the git-sync container
  GITHUB_APP_PRIVATE_KEY_MOUNT+=(-v "${TEST_GITHUB_APP_PRIVATE_KEY_FILE}:/${LOCAL_GITHUB_APP_PRIVATE_KEY_FILE}:ro")
...
        "${GITHUB_APP_PRIVATE_KEY_MOUNT[@]}" \
        "${GIT_SYNC_E2E_IMAGE}" \
```

sdowell:
@risset @thockin Is this PR still being worked on?

risset:
> @risset @thockin Is this PR still being worked on?

Yeah I was just waiting for @thockin as he said he'd be away for a while. Also I think your review comments make sense but since @nan-yu wrote those parts of `test_e2e.sh`, I thought they might have some thoughts on it before I change those things. @nan-yu what do you think?

thockin:
Yes, the ball is in my court to evaluate testing.  I apologize, it has been a chaotic few weeks since I got back from holiday.  I have not forgotten it.

test_e2e.sh
nan-yu: Changing it to an empty array looks good to me. Thanks for pointing it out.

test_e2e.sh
nan-yu: That makes sense as long as the github action workflow doesn't skip the github app test.

test_e2e.sh
```diff
@@ -194,6 +194,48 @@ fi
...
+    "TEST_GITHUB_APP_CLIENT_ID"
+    "TEST_GITHUB_APP_PRIVATE_KEY_FILE"
+  )
+
+  # TEST_GITHUB_APP_PRIVATE_KEY, if set, overrides TEST_GITHUB_APP_PRIVATE_KEY_FILE
+  if [[ -v TEST_GITHUB_APP_PRIVATE_KEY && -n "${TEST_GITHUB_APP_PRIVATE_KEY}" ]]; then
+    if [[ ! -v TEST_GITHUB_APP_PRIVATE_KEY_FILE || -z "${TEST_GITHUB_APP_PRIVATE_KEY_FILE}" ]]; then
+      TEST_GITHUB_APP_PRIVATE_KEY_FILE="${DIR}/${LOCAL_GITHUB_APP_PRIVATE_KEY_FILE}"
+    fi
+    echo "${TEST_GITHUB_APP_PRIVATE_KEY}" > "${TEST_GITHUB_APP_PRIVATE_KEY_FILE}"
```
thockin, line 225: I'm worried about this path, seems like an opportunity to blow away a file we shouldn't.  Maybe we can make a temp file and use that?

nan-yu: `TEST_GITHUB_APP_PRIVATE_KEY_FILE` points to a temp file when it is either unset or blank. Another option is to always use the temp file regardless of whether the variable is set, i.e., remove the if condition in line 222.

```
if [[ -v TEST_GITHUB_APP_PRIVATE_KEY && -n "${TEST_GITHUB_APP_PRIVATE_KEY}" ]]; then
  TEST_GITHUB_APP_PRIVATE_KEY_FILE="${DIR}/${LOCAL_GITHUB_APP_PRIVATE_KEY_FILE}"
  echo "${TEST_GITHUB_APP_PRIVATE_KEY}" > "${TEST_GITHUB_APP_PRIVATE_KEY_FILE}"
fi
```

thockin: I'm making a few edits as I look at this, will post them here.  May not finish today.

thockin: OK, I dug into the testing in depth and set up my own app.  With the below diff I am happy with this PR, but I have not yet looked at making CI work (with GH secrets)

```diff
diff --git a/test_e2e.sh b/test_e2e.sh
index 9b2023f..142d7c1 100755
--- a/test_e2e.sh
+++ b/test_e2e.sh
@@ -41,6 +41,11 @@ function fail() {
     return 42
 }
 
+function skip() {
+    echo "SKIP" >&3
+    return 43
+}
+
 function pass() {
     echo "PASS"
 }
@@ -204,37 +209,43 @@ function final_cleanup() {
 # Set the trap to call the final_cleanup function on exit.
 trap final_cleanup EXIT
 
-skip_github_app_test="${SKIP_GITHUB_APP_TEST:-false}"
+skip_github_app_test="${SKIP_GITHUB_APP_TEST:-true}"
 required_env_vars=()
 LOCAL_GITHUB_APP_PRIVATE_KEY_FILE="github_app_private_key.pem"
-GITHUB_APP_PRIVATE_KEY_MOUNT=""
+GITHUB_APP_PRIVATE_KEY_MOUNT=()
 if [[ "${skip_github_app_test}" != "true" ]]; then
   required_env_vars=(
     "TEST_GITHUB_APP_AUTH_TEST_REPO"
     "TEST_GITHUB_APP_APPLICATION_ID"
     "TEST_GITHUB_APP_INSTALLATION_ID"
     "TEST_GITHUB_APP_CLIENT_ID"
-    "TEST_GITHUB_APP_PRIVATE_KEY_FILE"
   )
 
-  # TEST_GITHUB_APP_PRIVATE_KEY, if set, overrides TEST_GITHUB_APP_PRIVATE_KEY_FILE
-  if [[ -v TEST_GITHUB_APP_PRIVATE_KEY && -n "${TEST_GITHUB_APP_PRIVATE_KEY}" ]]; then
-    if [[ ! -v TEST_GITHUB_APP_PRIVATE_KEY_FILE || -z "${TEST_GITHUB_APP_PRIVATE_KEY_FILE}" ]]; then
-      TEST_GITHUB_APP_PRIVATE_KEY_FILE="${DIR}/${LOCAL_GITHUB_APP_PRIVATE_KEY_FILE}"
-    fi
-    echo "${TEST_GITHUB_APP_PRIVATE_KEY}" > "${TEST_GITHUB_APP_PRIVATE_KEY_FILE}"
+  if [[ -n "${TEST_GITHUB_APP_PRIVATE_KEY_FILE:-}" && -n "${TEST_GITHUB_APP_PRIVATE_KEY:-}" ]]; then
+      echo "ERROR: Both TEST_GITHUB_APP_PRIVATE_KEY_FILE and TEST_GITHUB_APP_PRIVATE_KEY were specified."
+      exit 1
+  fi
+  if [[ -n "${TEST_GITHUB_APP_PRIVATE_KEY_FILE:-}" ]]; then
+    cp "${TEST_GITHUB_APP_PRIVATE_KEY_FILE}" "${DIR}/${LOCAL_GITHUB_APP_PRIVATE_KEY_FILE}"
+  elif [[ -n "${TEST_GITHUB_APP_PRIVATE_KEY:-}" ]]; then
+    echo "${TEST_GITHUB_APP_PRIVATE_KEY}" > "${DIR}/${LOCAL_GITHUB_APP_PRIVATE_KEY_FILE}"
+  else
+    echo "ERROR: Neither TEST_GITHUB_APP_PRIVATE_KEY_FILE nor TEST_GITHUB_APP_PRIVATE_KEY was specified."
+    echo "       Either provide a value or skip this test (SKIP_GITHUB_APP_TEST=true)."
+    exit 1
   fi
 
   # Validate all required environment variables for the github-app-auth tests are provided.
   for var in "${required_env_vars[@]}"; do
     if [[ ! -v "${var}" ]]; then
-      echo "Error: Required environment variable '${var}' is not set or empty. Either provide a value or skip the GitHub App test by setting SKIP_GITHUB_APP_TEST to 'true'."
+      echo "ERROR: Required environment variable '${var}' is not set."
+      echo "       Either provide a value or skip this test (SKIP_GITHUB_APP_TEST=true)."
       exit 1
     fi
   done
 
   # Mount the GitHub App private key file to the git-sync container
-  GITHUB_APP_PRIVATE_KEY_MOUNT=(-v "${TEST_GITHUB_APP_PRIVATE_KEY_FILE}":"/${LOCAL_GITHUB_APP_PRIVATE_KEY_FILE}":ro)
+  GITHUB_APP_PRIVATE_KEY_MOUNT=(-v "${DIR}/${LOCAL_GITHUB_APP_PRIVATE_KEY_FILE}":"/${LOCAL_GITHUB_APP_PRIVATE_KEY_FILE}":ro)
 fi
 
 # WORK is temp space and in reset for each testcase.
@@ -2232,7 +2243,7 @@ function e2e::auth_askpass_url_slow_start() {
 ##############################################
 function e2e::auth_github_app_application_id() {
     if [[ "${skip_github_app_test}" == "true" ]]; then
-        return
+        skip
     fi
     GIT_SYNC \
         --one-time \
@@ -2247,7 +2258,7 @@ function e2e::auth_github_app_application_id() {
 
 function e2e::auth_github_app_client_id() {
     if [[ "${skip_github_app_test}" == "true" ]]; then
-        return
+        skip
     fi
     GIT_SYNC \
         --one-time \
@@ -3658,6 +3669,8 @@ for t; do
         run_test RUN_RET "${TEST_FN}" >"${LOG}.${RUN}" 2>&1
         if [[ "$RUN_RET" == 0 ]]; then
             pass
+        elif [[ "$RUN_RET" == 43 ]]; then
+            true # do nothing
         else
             TEST_RET=1
             if [[ "$RUN_RET" != 42 ]]; then
```

thockin:
Since this is a kubernetes repo, I'll have to see about making the kubernetes org own the app

HorusTheSonOfOsiris:
> Since this is a kubernetes repo, I'll have to see about making the kubernetes org own the app

Any plans on releasing this soon? This could help our team a lot.

thockin:
I'm going to merge this and followup.  I Still don't have CI for GH apps up, though.

PR #882, opened 2024-06-08
Support -? as an alias for --help

thockin:
Also print errors before and after usage, like pflag.


thockin:
This is leading to better env var handling

PR #883, opened 2024-06-08
Clean up flag-related messages

thockin:
@nan-yu - this would change the exact strings that are logged to `--error-file` - we don't really have a spec or tests for the contents of that file.  Do you think this change is OK?  If not, we should spec why not.


nan-yu (approved):
/lgtm

The message in the error log file is purely informational and not meant to be interpreted as exact output.

Issue #891, opened 2024-06-13
TODO: shellcheck test_e2e.sh and test_git.sh


thockin:
Its kind of a mess

rul:
I'd like to give this a try. :slightly_smiling_face: 

/assign

thockin:
Retitled to reference test_git.sh :)

thockin:
test_e2e.sh is done - now for test_git.sh ?

rul:
> test_e2e.sh is done - now for test_git.sh ?

Yup! Will put up a PR soon.

PR #892, opened 2024-06-19
test_e2e.sh: quote variables to prevent globbing and word splitting

rul:
Fixes SC2086. Part of #891. 

```
git-sync#fix-e2e-sc2086> shellcheck -i SC2086 test_e2e.sh  && echo $?
0
```
As explained in 390f63cc6ed458955d0d66567ee5faa97cdf6106, some strings don't require to be fully quoted, but doing
so anyway to keep code consistency.


thockin (commented):
Thanks!

/lgtm
/approve

PR #894, opened 2024-06-20
consistently use bash shebang for shell scripts

sdowell:
This updates the build scripts to use a bash shebang for consistency with other scripts as well as better portability. If using a non-alpine BUILDIMAGE (e.g. debian) the build scripts are not valid sh syntax.

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
 # Copyright 2016 The Kubernetes Authors.
 #
--- a/build/test.sh
+++ b/build/test.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
 # Copyright 2016 The Kubernetes Authors.
 #


nan-yu (approved):
/lgtm

thockin:
As #895 shows, "bash" is not available in the BUILD_IMAGE (golang:1.22-alpine).  We should make the scripts be stricter `sh`

sdowell:
> As #895 shows, "bash" is not available in the BUILD_IMAGE (golang:1.22-alpine). We should make the scripts be stricter `sh`

Could we switch the default BUILD_IMAGE to be the debian-based `golang:1.22`? That image ships with bash and would be a more consistent OS with `BASEIMAGE`.

I'm good with also making the script more strict/portable, but I don't see a benefit in using the alpine image in this instance.

sdowell:
Are there not presubmits that verify the build?

thockin:
There should be.  I'm not sure why this would have passed

PR #898, opened 2024-06-21
consistently use bash shebang for shell scripts

sdowell:
This updates the build scripts to use a bash shebang for consistency with other scripts as well as better portability. If using a non-alpine BUILDIMAGE (e.g. debian) the build scripts are not valid sh syntax.

Also updates BUILDIMAGE to use the debian based golang image for consistency with BASEIMAGE. The debian based golang image also comes with bash installed, whereas the alpine base image does not.


thockin:
I went the other way and made them more sh compatible.

sdowell:
> I went the other way and made them more sh compatible.

Can we switch the BUILDIMAGE to debian for better consistency? Why use alpine?

PR #900, opened 2024-06-21
use debian build image and bash build scripts

sdowell:
This makes several changes to the build environment in the interest of consistency:
- Changes the build image to use a debian based image for consistency with the BASEIMAGE
- Changes the build scripts to use bash for consistency with other scripts in the repo


thockin:
Discussed offline.

/lgtm
/approve

PR #901, opened 2024-06-22
test_e2e.sh: Use index to expand FUNCNAME

rul:
Fixes SC2128. Part of #891.


thockin:
Thanks!

/lgtm
/approve

thockin:
You could make make of these same changes to test_git.sh, too :)

rul:
> You could make make of these same changes to test_git.sh, too :)

Sounds good. Will do!

PR #902, opened 2024-06-23
test_e2e.sh: shellcheck entire file

rul:
This contribution shellchecks the whole `test_e2e.sh` file. Check individual commits for more details. I've ran `make test` and verified that all tests still pass. Part of #891.

### DIFF
--- a/test_e2e.sh
+++ b/test_e2e.sh
@@ -50,12 +50,12 @@ function wait_for_file_exists() {
     local file=$1
     local ticks=$(($2*10)) # 100ms per tick
 
-    while (( $ticks > 0 )); do
+    while (( ticks > 0 )); do
         if [[ -f "$file" ]]; then
             break
         fi
         sleep 0.1
-        ticks=$(($ticks-1))
+        ticks=$((ticks-1))
     done
 }
 
--- a/test_e2e.sh
+++ b/test_e2e.sh
@@ -18,6 +18,7 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+# shellcheck disable=SC2120
 function caller() {
   local stack_skip=${1:-0}
   stack_skip=$((stack_skip + 1))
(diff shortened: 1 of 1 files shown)


test_e2e.sh
```diff
@@ -1658,7 +1658,7 @@ function e2e::sync_depth_across_updates() {
     assert_metric_eq "${METRIC_GOOD_SYNC_COUNT}" 1
     assert_metric_eq "${METRIC_FETCH_COUNT}" 1
     depth=$(git -C "$ROOT/link" rev-list HEAD | wc -l)
-    if [[ $expected_depth != $depth ]]; then
+    if [[ $expected_depth != "$depth" ]]; then
```
thockin, line 1661: We should probably quote both sides

thockin:
just a couple picks

thockin:
Another pass might be to use `local` for all variables defined inside functions and to convert them all to common name style (lowercase for locals)

test_e2e.sh
```diff
@@ -36,7 +36,7 @@ function caller() {
 }
 
 function fail() {
-    echo "FAIL: line $(caller):" "$@" >&3
+    echo "FAIL: line $(caller "$@"):" "$@" >&3
```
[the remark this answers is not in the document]
rul: I think you're right. Fixed!

test_e2e.sh
rul: `shellcheck` was probably not detecting this because `expected_depth` is getting set only on two occasions to the static value 1:

```
rg expected_depth=
test_e2e.sh
1644:    expected_depth="1"
2709:    expected_depth="1"
```

Anyways, I agree with quoting it to keep the code consistent. Fixed!

rul:
> Another pass might be to use `local` for all variables defined inside functions and to convert them all to common name style (lowercase for locals)

Makes sense. I think it's a bit out of scope for #801, so I've created #903.

PR #910, opened 2024-07-16
test: harden e2e image build logic

sdowell:
This introduces several changes to make the e2e test image logic more robust:

- Rename IMAGE to GIT_SYNC_E2E_IMAGE to minimize accidental misuse
- Ensure that the image built by make container is the same as the image used by docker run
- Do not build the container if a custom image is specified

As an example, with the prior logic the following command always fails:
IMAGE="" make test
...
Unable to find image 'gcr.io/k8s-staging-git-sync/git-sync:latest' locally

Because the docker run command pulls a different image from the one built by make container.


thockin (commented):
Thanks!

/lgtm
/approve

PR #933, opened 2025-02-05
fix: Correct typo in code comment

lkb85:
Minor typo fixes in comments.

Changed files:
- vendor/github.com/prometheus/client_golang/prometheus/wrap.go
- vendor/github.com/prometheus/common/model/labels.go
- vendor/github.com/spf13/pflag/count.go

This change does not include functional changes.


thockin:
Thanks, but we don't edit vendor'ed code - if you would like to fix those, they need to be fixed in the relevant upstream repositories.

Issue #944, opened 2025-04-29
Git Sync v4.4.0 sync error with one-time and HEAD


AlanZhang2002:
Hello! 
```
{"logger":"","ts":"2025-04-28 23:19:03.199264","caller":{"file":"main.go","line":626},"level":0,"msg":"starting up","version":"v4.4.0","pid":14,"uid":65533,"gid":65533,"home":"/tmp","flags":[]}                  │
│ {"logger":"","ts":"2025-04-28 23:19:03.199411","caller":{"file":"main.go","line":663},"level":0,"msg":"normalized root path","root":"/git/src/","result":"/git/src"}                                               │
│ {"logger":"","ts":"2025-04-28 23:19:03.200899","caller":{"file":"main.go","line":731},"level":0,"msg":"git version","version":"git version 2.39.5"}                                                                │
│ {"logger":"","ts":"2025-04-28 23:19:06.128665","caller":{"file":"main.go","line":1775},"level":0,"msg":"update required","ref":"HEAD","local":"f6c7eeaf26e32d86140c7e5d18e8f49b111e3f65","remote":"f6c7eeaf26e32d8 │
│ 6140c7e5d18e8f49b111e3f65","syncCount":0}                                                                                                                                                                          │
│ {"logger":"","ts":"2025-04-28 23:19:06.141481","caller":{"file":"main.go","line":924},"msg":"too many failures, aborting","error":"Run(git reset --soft f6c7eeaf26e32d86140c7e5d18e8f49b111e3f65): exit status 128 │
│ : { stdout: \"\", stderr: \"fatal: ambiguous argument 'f6c7eeaf26e32d86140c7e5d18e8f49b111e3f65': both revision and filename\\nUse '--' to separate paths from revisions, like this:\\n'git <command> [<revision>. │
│ ..] -- [<file>...]'\" }","failCount":1}
```
On v4.4.0, I've set up git-sync to sync on HEAD with `GITSYNC_ONE_TIME` set to true within a K8 Job, but am getting this error related to git reset. I'm not sure if it is intended for it to try to sync when it is on the same commit hash, or if there is something else going on. There's no file in the repo with the same name as the commit hash, HEAD, or the HEAD branch name. Should the reset command be updated to fix, as suggested by the error message? Setting it to sync on a different commit has no issue, but syncing on HEAD, the commit hash itself, or the HEAD branch name all fails with the same error.

PR #946, opened 2025-04-29
added separator after ref to remove git reset --soft ambiguity

AlanZhang2002:
Fixes https://github.com/kubernetes/git-sync/issues/944

### DIFF
--- a/main.go
+++ b/main.go
@@ -1780,7 +1780,7 @@ func (git *repoSync) SyncRepo(ctx context.Context, refreshCreds func(context.Con
 		// Reset the repo (note: not the worktree - that happens later) to the new
 		// ref.  This makes subsequent fetches much less expensive.  It uses --soft
 		// so no files are checked out.
-		if _, _, err := git.Run(ctx, git.root, "reset", "--soft", remoteHash); err != nil {
+		if _, _, err := git.Run(ctx, git.root, "reset", "--soft", remoteHash, "--"); err != nil {
 			return false, "", err
 		}
 
--- a/test_e2e.sh
+++ b/test_e2e.sh
@@ -1492,6 +1492,40 @@ function e2e::sync_crash_no_worktree_cleanup_retry() {
 }
 
 ##############################################
+# Test syncing if a file named for the SHA exists
+##############################################
+function e2e::sync_sha_shafile_exists() {
+    echo "${FUNCNAME[0]} 1" > "$REPO/file"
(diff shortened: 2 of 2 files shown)


thockin:
Thanks!

This SEEMS like an appropriate change, but can you help me with a repro case in the e2e (./test_e2e.sh)

Clearly it doesn't fail the current e2e test, so we need to encode whatever scenario you hit into a test case.

AlanZhang2002:
Sure, I'll try to reproduce in testing, though I'm honestly not completely sure how I hit it. 
Here's a little snippet of part of the setup (besides auth):
```
        - name: *******
          image: registry.k8s.io/git-sync/git-sync:v4.4.0
          imagePullPolicy: Always
          env:
            - name: GITSYNC_REPO
              value: {{ .Values.kubernetes.git.repo }}
            - name: GITSYNC_REF
              value: {{ .Values.kubernetes.git.ref }} # for this case, just HEAD
            - name: GITSYNC_ONE_TIME
              value: "true"
            - name: GITSYNC_ROOT
              value: "/git/src/"
            - name: GITSYNC_LINK
              value: "repo"
```
One thing to note is that when this happens, I have an existing k8 git-sync deployment that doesn't run with --one-time, and when I run this version with the --one-time flag as a k8 job on the same repo and ref that it is already synced on, the error happens. Running in an environment where there is only the --one-time version doing the updating does not make this error happen, when syncing on a repo that was already synced. Not sure exactly why, or how to put this in a test case.

thockin:
Are they sharing a volume on disk?

thockin:
Oh.  That is not at all something I test for, and I am not too surprised it breaks.  The default settings are pretty aggressive about pruning inactive data (as determined by the sync being performed).

Can you explain more about why you have it set up this way?

AlanZhang2002:
It was just temporarily how it was set up while I was testing, not a setup we actually stick with. I'm sure no actual git-sync setup would have something like this also, but it's probably a good change to add `--` into the command anyways.

thockin:
Yeah, I think the `--` is likely correct, but I want to have a test.

AlanZhang2002:
I think I'm unable to test locally for some reason, my docker setup is a bit scuffed atm probably. I added the test to the pr though, I'll see if I can get it running locally.

thockin:
OK.  I ran it locally and passes.  Putting it in the PR means we don't get half of it.

/lgtm
/approve

PR #952, opened 2025-06-16
fix: gracefully handle nil err in Logger

sdowell:
This handles the case where a null error is passed into the Logger without causing a null pointer dereference / segfault. There are currently a couple places that invoke this function with a nil error and were resulting in a segfault.


sdowell:
> This seems reasonable - can you show how this might be hit? It has never come up before .

At the time of writing there are two places that explicitly call this with `nil`:
- https://github.com/kubernetes/git-sync/blob/v4.4.1/main.go#L1334
- https://github.com/kubernetes/git-sync/blob/v4.4.1/main.go#L1352

thockin:
Hmm, I guess we don't have good tests for these.

PR #954, opened 2025-06-25
add --hooks-async and --hooks-before-symlink flags

### LINKED ISSUE #937: How to Prevent .env and upload from Being Deleted During git-sync (Similar to .gitignore)
The issue is that after git-sync successfully completes, all untracked files are deleted, including the external file links and .env file connections. However, recreating the links after syncing feels a bit odd, as the file connections are lost during the sync process.

How can this be solved? Should I recreate the symbolic links immediately after the sync finishes?

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: web-deployment
spec:
  replicas: 1
  revisionHistoryLimit: 3
  selector:
    matchLabels:
      app: example
  template:
    metadata:
      labels:
        app: example
    spec:
      containers:
        - name: example
          image: docker.redgistry.kr/example:latest
          imagePullPolicy: Always
          volumeMounts:
          - name: web-source
            mountPath: /var/www/html
          - name: storage
            mountPath: /var/www/html/upload
          - name: htaccess
            mountPath: /var/www/html/.htaccess
            subPath: .htaccess
          - name: config
            mountPath: /var/www/html/.env
            subPath: .env
          command: ["/bin/sh", "-c"]
          args:
            - |
              chown -R root:www-data /var/www/html/upload && \
              chmod -R 775 /var/www/html/upload && \
              ln -s /var/www/html/.env /var/www/html/web/.env && \
              ln -s /var/www/html/.htaccess /var/www/html/web/.htaccess && \
              ln -s /var/www/html/upload /var/www/html/web/upload && \
              apache2-foreground
          ports:
            - containerPort: 80
          env:
          - name: TZ
            value: Asia/Seoul
        - name: git-sync
          image: registry.k8s.io/git-sync/git-sync:v4.0.0
          securityContext:
            runAsUser: 1000
            runAsGroup: 1000
          args:
            - "--repo=https://$(GIT_USERNAME):$(GIT_TOKEN)@github.com/redgistry-kr/example.git"
            - "--branch=$(BRANCH)"
            - "--root=/git"
            - "--dest=web"
            - "--one-time=false"
            - "--period=60s"
            - "--wait=60"
          env:
            - name: BRANCH
              value: branch
            - name: GIT_USERNAME
              valueFrom:
                secretKeyRef:
                  name: git-credentials
                  key: GIT_USERNAME
            - name: GIT_TOKEN
              valueFrom:
                secretKeyRef:
                  name: git-credentials
                  key: GIT_TOKEN
          volumeMounts:
            - name: web-source
              mountPath: /git
      imagePullSecrets:
        - name: redgistry-registry-secret
      volumes:
        - name: storage
          persistentVolumeClaim:
            claimName: web-pvc
        - name: htaccess
          configMap:
            name: htaccess
        - name: config
          configMap:
            name: web-env
        - name: web-source
          emptyDir: {}

### PR DESCRIPTION
**Pull request by tylenwells**
https://github.com/kubernetes/git-sync/issues/937

**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
Adds flags for --hooks-async and --hooks-before-symlink
**Which issue(s) this PR fixes**:

Fixes #937

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
- adds --hooks-async: Whether to run the exechook and webhook commands in an async or sync way. 
- adds --hooks-before-symlink: When to run the --exechook-command.  If true, will run the exechook or webhook commands before the symlink is updated. 
            ExecHook is asynchronous, so setting this will
            not block the sync process. If you need the hook to run to completion
            before the symlink is updated, you should use this in conjunction with --hooks-async=false.
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs
```

### DIFF
--- a/_test_tools/exechook_command_with_sleep.sh
+++ b/_test_tools/exechook_command_with_sleep.sh
@@ -19,4 +19,6 @@
 
 sleep 3
 cat file > exechook
+cat exechook
+if [[ "$(pwd)" != "$(pwd -P)" ]]; then echo "true" > delaycheck; fi
 echo "ENVKEY=$ENVKEY" > exechook-env
--- a/main.go
+++ b/main.go
@@ -240,6 +240,9 @@ func main() {
 	flHooksAsync := pflag.Bool("hooks-async",
 		envBool(true, "GITSYNC_HOOKS_ASYNC", "GIT_SYNC_HOOKS_ASYNC"),
 		"run hooks asynchronously")
+	flHooksBeforeSymlink := pflag.Bool("hooks-before-symlink",
+		envBool(false, "GITSYNC_HOOKS_BEFORE_SYMLINK", "GIT_SYNC_HOOKS_BEFORE_SYMLINK"),
+		"run hooks before creating the symlink (defaults to false)")
 
 	flUsername := pflag.String("username",
 		envString("", "GITSYNC_USERNAME", "GIT_SYNC_USERNAME"),
@@ -879,6 +882,25 @@ func main() {
 		go exechookRunner.Run(context.Background())
 	}
 
+	runHooks := func(hash string) error {
+		var err error
+		if exechookRunner != nil {
+			log.V(3).Info("sending exechook")
+			err = exechookRunner.Send(hash)
+			if err != nil {
+				return err
+			}
+		}
+		if webhookRunner != nil {
+			log.V(3).Info("sending webhook")
+			err = webhookRunner.Send(hash)
+		}
+		if err != nil {
+			return err
+		}
+		return nil
+	}
+
 	// Setup signal notify channel
 	sigChan := make(chan os.Signal, 1)
 	if syncSig != 0 {
@@ -919,11 +941,12 @@ func main() {
 
 	failCount := 0
 	syncCount := uint64(0)
+
 	for {
 		start := time.Now()
 		ctx, cancel := context.WithTimeout(context.Background(), *flSyncTimeout)
 
-		if changed, hash, err := git.SyncRepo(ctx, refreshCreds); err != nil {
+		if changed, hash, err := git.SyncRepo(ctx, refreshCreds, runHooks, *flHooksBeforeSymlink); err != nil {
 			failCount++
 			updateSyncMetrics(metricKeyError, start)
 			if *flMaxFailures >= 0 && failCount >= *flMaxFailures {
@@ -944,11 +967,10 @@ func main() {
 						log.V(3).Info("touched touch-file", "path", absTouchFile)
 					}
 				}
-				if webhookRunner != nil {
-					webhookRunner.Send(hash)
-				}
-				if exechookRunner != nil {
-					exechookRunner.Send(hash)
+				// if --hooks-before-symlink is set, these will have already been sent and completed.
+				// otherwise, we send them now.
+				if !*flHooksBeforeSymlink {
+					runHooks(hash)
 				}
 				updateSyncMetrics(metricKeySuccess, start)
 			} else {
@@ -968,14 +990,17 @@ func main() {
 				// Assumes that if hook channels are not nil, they will have at
 				// least one value before getting closed
 				exitCode := 0 // is 0 if all hooks succeed, else is 1
-				if exechookRunner != nil {
-					if err := exechookRunner.WaitForCompletion(); err != nil {
-						exitCode = 1
+				// This will not be needed if async == false, because the Send func for the hookRunners will wait
+				if *flHooksAsync {
+					if exechookRunner != nil {
+						if err := exechookRunner.WaitForCompletion(); err != nil {
+							exitCode = 1
+						}
 					}
-				}
-				if webhookRunner != nil {
-					if err := webhookRunner.WaitForCompletion(); err != nil {
-						exitCode = 1
+					if webhookRunner != nil {
+						if err := webhookRunner.WaitForCompletion(); err != nil {
+							exitCode = 1
+						}
 					}
 				}
 				log.DeleteErrorFile()
@@ -1723,7 +1748,7 @@ func (git *repoSync) currentWorktree() (worktree, error) {
 // SyncRepo syncs the repository to the desired ref, publishes it via the link,
 // and tries to clean up any detritus.  This function returns whether the
 // current hash has changed and what the new hash is.
-func (git *repoSync) SyncRepo(ctx context.Context, refreshCreds func(context.Context) error) (bool, string, error) {
+func (git *repoSync) SyncRepo(ctx context.Context, refreshCreds func(context.Context) error, runHooks func(hash string) error, flHooksBeforeSymlink bool) (bool, string, error) {
 	git.log.V(3).Info("syncing", "repo", redactURL(git.repo))
 
 	if err := refreshCreds(ctx); err != nil {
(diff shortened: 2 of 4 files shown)


tylenwells:
adding a hold while I continue working on this. 

Still need to figure out testing + resolving build errors.

thockin:
Thanks!  I am traveling the rest of this week, but will take a look when I can.

@sdowell can also offer feedback

main.go
```diff
@@ -1802,6 +1819,31 @@ func (git *repoSync) SyncRepo(ctx context.Context, refreshCreds func(context.Con
 			return false, "", err
 		}
 
+		// If --delay-symlink is true, we don't update the symlink or finish the post-sync tasks until hooks complete.
+		if *flDelaySymlink {
+			git.log.V(0).Info("delaying symlink update", "ref", git.ref, "remote", remoteHash, "syncCount", git.syncCount)
+
+			if webhookRunner != nil {
+				webhookRunner.Send(remoteHash)
```
thockin, line 1827: First, I think it would be clearer to frame this as "--exechook-wait" and equivalent for webhook.

Then. How about factoring the send and wait logic to a function, so the same logic is called in both sync and async cases?

tylenwells: I think splitting this as --exechook-wait and --webhook-wait would be fine. 


If I'm understanding correctly then, you're suggesting:

- separate flags for delaying symlink publishing for each of the hooks
- a single function for triggering a hook runner as well as the WaitForCompletion? (one for each of the hooks) 


Then if we do it like that, we would end up with


--exechook-when
--exechook-wait
--webhook-when
--webhook-wait

That sounds fine to me. 



I did just push another commit though that I'd like feedback on. I was trying to reference webhookRunner, exechookRunner, the ExechookWhen flag, and the DelaySymlink flag in the SyncRepo function but it was failing as they were out of scope. 

I did some reading and found that I could make it work by passing those values through with the context, which is what I did in [29eacef](https://github.com/kubernetes/git-sync/pull/954/commits/29eacefe718d25a65a58e1b6bf87b8acf625987c). 

Most of my days are spent staring at YAML vs actually coding, so feel free to let me know if I'm running off the grid here. I'm taking this as a learning experience as I go.

tylenwells: I've been looking at this more over the evening and trying to get a better understanding of how the async vs sync runners work. 

Looks like I'll need to do some further thinking about how to go about measuring completion on async runners.

main.go
```diff
@@ -912,9 +930,17 @@ func main() {
...
+		"webhook":              webhookRunner,
+		"exechook":             exechookRunner,
+		"flHooksAsync":         *flHooksAsync,
+		"flHooksBeforeSymlink": *flHooksBeforeSymlink,
+	}}
+
 	for {
 		start := time.Now()
-		ctx, cancel := context.WithTimeout(context.Background(), *flSyncTimeout)
+		ctx, cancel := context.WithTimeout(context.WithValue(context.Background(), "hookvalues", v), *flSyncTimeout)
```
thockin, line 943: Rather than squirreling it into ctx, how about crafting a func(hash) which you pass alongside the credentials func.  That way, all the hooks logic is defined here.

It might change the order of things like hooks vs. touch file, but I think that is OK.  May need to include the setRepoReady.  I am on mobile so this SEEMS like a good idea but I can't go try it to be  sure :)

pkg/hook/hook.go
```diff
@@ -110,6 +112,10 @@ type HookRunner struct {
 	// Used to send a status result when running in one-time mode.
 	// Should be initialised to a buffered channel of size 1.
 	oneTimeResult chan bool
+	// Used for non-async hook execution status result, to allow
+	// WaitForCompletion to be called even if not using oneTime.
+	// Should be initialised to a buffered channel of size 1.
+	nonAsyncResult chan bool
```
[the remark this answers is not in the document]
tylenwells: Yeah, this makes sense and should be fine. I'll make the change.

pkg/hook/hook.go
```diff
@@ -124,6 +130,17 @@ func (r *HookRunner) Send(hash string) {
 	r.data.send(hash)
 }
 
+// SendAndWait ends sends hash to hookdata and waits for the hook to complete.
+func (r *HookRunner) SendAndWait(hash string) error {
```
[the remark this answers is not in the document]
tylenwells: I agree with this, I will consolidate.

_test_tools/exechook_command_with_sleep.sh
```diff
@@ -19,4 +19,7 @@
 
 sleep 3
 cat file > exechook
+cat file
```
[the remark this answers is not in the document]
tylenwells: Yep, ty. Will clean it up.

test_e2e.sh
```diff
@@ -2509,6 +2509,32 @@ function e2e::exechook_success_hooks_non_async() {
...
 }
 
+##############################################
+# Test exechook-success with --hooks-async=false and --hooks-before-symlink
+##############################################
+function e2e::exechook_success_hooks_before_symlink_non_async() {
+    cat /dev/null > "$RUNLOG"
+
+    GIT_SYNC \
+        --hooks-async=false \
```
thockin, line 2519: args-under-test go at the end of the list

test_e2e.sh
```diff
@@ -2509,6 +2509,32 @@ function e2e::exechook_success_hooks_non_async() {
...
+        --exechook-backoff=1s \
+        &
+    sleep 5 # give it time to run
+    assert_link_exists "$ROOT/link"
+    assert_file_exists "$ROOT/link/file"
+    assert_file_exists "$ROOT/link/exechook"
+    assert_file_eq "$ROOT/link/file" "${FUNCNAME[0]}"
+    assert_file_eq "$ROOT/link/exechook" "${FUNCNAME[0]}"
+    assert_file_eq "$ROOT/link/exechook-env" "$EXECHOOK_ENVKEY=$EXECHOOK_ENVVAL"
+    assert_file_absent "$ROOT/link/delaycheck"
```
thockin, line 2535: I am unclear on what this is actually proving?  Can you explain?

main.go
```diff
@@ -2574,6 +2574,13 @@ OPTIONS
     -?, -h, --help
             Print help text and exit.
 
+	--hooks-async, $GITSYNC_HOOKS_ASYNC
+			Whether to run the --exechook-command asynchronously.
```
thockin, line 2578: exec and web

main.go
```diff
@@ -2574,6 +2574,13 @@ OPTIONS
...
+
+	--hooks-before-symlink, $GITSYNC_HOOKS_BEFORE_SYMLINK
+			Whether to run the --exechook-command before updating the symlink. Use in combination with --hooks-async set
```
thockin, line 2581: exec and web

thockin: linewrap to 80 please

main.go
```diff
@@ -2574,6 +2574,13 @@ OPTIONS
     -?, -h, --help
             Print help text and exit.
 
+	--hooks-async, $GITSYNC_HOOKS_ASYNC
```
thockin, line 2577: also need to update the manual in README.md

main.go
```diff
@@ -961,14 +990,17 @@ func main() {
 				// Assumes that if hook channels are not nil, they will have at
 				// least one value before getting closed
 				exitCode := 0 // is 0 if all hooks succeed, else is 1
-				if exechookRunner != nil {
-					if err := exechookRunner.WaitForCompletion(); err != nil {
-						exitCode = 1
+				// This will not be needed if async == false, because the Send func for the hookRunners will wait
+				if *flHooksAsync {
+					if exechookRunner != nil {
+						if err := exechookRunner.WaitForCompletion(); err != nil {
```
thockin, line 996: Looking at it more, this is a bit tricky - you might get a result on the channel that indicates that the hook failed, but it will be retried (async).  For one-time that's fine, but is it OK in non-one-time mode?  If you ask for synchronous hooks, and the hook fails -- do we think that's "good enough" or do we wait for a success?  What if success never comes?

thockin:
Is this still something you intend to pursue?

Issue #970, opened 2026-01-16
Feature Request: availibity of tar and zip in the docker image to use with git-archive


NBardelot:
It seems that a reasonable usage of ` --exechook-command` / `$GITSYNC_EXECHOOK_COMMAND` is to run a `git-archive` command on the local repository after it has been synced from the original source (for example, it's a very convenient way to build a backup).

See the `git-archive` manpage: the `--format=` of such an archive is usually some very classical `tar`, or `zip`, or `tgz`.

Thus, it would seem helpful that the `git-sync` container image provides the `tar`, `gzip` and or `zip` (ideally all of them). Especially since it is hard to add them afterwards. For example, when building a custom image based on `git-sync` the environnement is already minimal, with no package management tool ready to use, and very few dynamic libraries.

Another option would be to continue having a basic `git-sync` image, but offer a complementary image with busybox tools on top of it, more tailored to the needs of those who run exec hooks and need a little bit of tooling.

thockin:
The more tools we include in the image, the more opportunities there are for bugs and the more I have to rebuild just for CVEs.  By default it runs as non-root, so it's not ACTUALLY a big deal, but the way scanners work today ignores that, and people's dashboards light up anyway.

I don't think I understand the use-case -- why do you need a backup of a git repo that you, just moments ago, synced from an upstream?

NBardelot:
@thockin  thanks for the quick answer. Here are two _git centric_ real world use-cases where we'd want `git-archive` running as a hook: 

1. Running `git-archive` as a post-sync hook in order to keep an archive of the _flattened_ content a complex git projet with multiple submodules. Such an archive is easier to store as a short-term backup that does not rely on the central git repository.
2. Serving the same archive from a shared ephemeral RAMFS volume using `git-daemon` as a local cache, decoupling the consumers from connecting to the central repository, while not changing the consumers behavior from an architecture standpoint (we don't want to switch to object storage, with consumers needing to switch API/SDK right now).

As a user of `git-sync`, I feel it's made to help _git centric_ designs in Kubernetes, and `git-archive` without `tar` or `tgz` means having to rebuild a container image from `git-sync` and adding `tar` as a 100% static binary. Which means building from sources with `cc`, something like this:

```
ARG CACHE_REGISTRY
ARG GITSYNC_IMAGE_VERSION="latest"

# Use Alpine to build a fully statically linked tar binary,
# because we can't rely on the libs in the git-sync image
[...]

PR #971, opened 2026-01-23
Improve local builds with another buildx builder

NBardelot:
Introduces to new parameters in the Makefile:
- BUILDX_BUILDER_NAME, allowing to select the builder
- BUILDX_BUILDER_SKIP_CREATION, to skip the builder creation when the builder already exists and is not managed by the Makefile

NOTE : I propose this in parallel of the feature request [#970](https://github.com/kubernetes/git-sync/issues/970) as I had to make some adaptations locally to be able to build the project, and I think they could be used by others.


thockin (commented):
Thanks!

/lgtm
/approve

PR #972, opened 2026-01-26
Improve running end-to-end tests locally

NBardelot:
* Allow users to run end-to-end tests with a fully-qualified and/or private `alpine` image
* Allow users to run end-to-end tests behind a proxy
* Introduces a new `NO_PROXY` parameter in the `Makefile` that is propagated alongside `HTTP_PROXY` and `HTTPS_PROXY`
* Introduces a new `LIST_OF_E2E_TESTS` parameter in the `Makefile` that drives the names of the tests passed to `./test_e2e.sh` to run a selection of tests, running all tests by default
* Allow users to override the git-sync repository URL used by end-to-end tests, by exporting `GIT_SYNC_REPOSITORY_URL` in their environment, in order to use a mirror if need be
* Sort `.gitgnore` for convenience for future diffs
* Add `.idea/` to `.gitignore`

NOTE : I propose this in parallel of the feature request [#970](https://github.com/kubernetes/git-sync/issues/970) as I had to make some adaptations locally to be able to run the end-to-end tests and I think they could be used by others. It completes the [PR #971](https://github.com/kubernetes/git-sync/pull/971) that has already been merged.

/kind feature

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,10 +34,14 @@ DBG ?=
 # These are passed to docker when building and testing.
 HTTP_PROXY ?=
 HTTPS_PROXY ?=
+NO_PROXY ?=
 
 # Allow some buildx adaptation for local builds
 BUILDX_BUILDER_NAME := git-sync
 BUILDX_BUILDER_SKIP_CREATION ?=
+
+# Allow alpine to be pulled from a private registry when building the end-to-end tests images
+ALPINE_REGISTRY_PREFIX ?=
 
 ###
 ### These variables should not need tweaking.
@@ -134,6 +138,7 @@ $(OUTBIN): .go/$(OUTBIN).stamp
 	    -v $$(pwd)/.go/cache:/.cache                           \
 	    --env HTTP_PROXY=$(HTTP_PROXY)                         \
 	    --env HTTPS_PROXY=$(HTTPS_PROXY)                       \
+	    --env NO_PROXY=$(NO_PROXY)                             \
 	    $(BUILD_IMAGE)                                         \
 	    /bin/sh -c "                                           \
 	        ARCH=$(ARCH)                                       \
@@ -190,6 +195,7 @@ container: .container-$(DOTFILE_IMAGE) container-name
 	    --platform "$(OS)/$(ARCH)"                               \
 	    --build-arg HTTP_PROXY=$(HTTP_PROXY)                     \
 	    --build-arg HTTPS_PROXY=$(HTTPS_PROXY)                   \
+	    --build-arg NO_PROXY=$(NO_PROXY)                         \
 	    -t $(IMAGE):$(OS_ARCH_TAG)                               \
 	    -f .dockerfile-$(OS)_$(ARCH)                             \
 	    .
@@ -248,17 +254,27 @@ test: $(BUILD_DIRS)
 	    -v $$(pwd)/.go/cache:/.cache                           \
 	    --env HTTP_PROXY=$(HTTP_PROXY)                         \
 	    --env HTTPS_PROXY=$(HTTPS_PROXY)                       \
+	    --env NO_PROXY=$(NO_PROXY)                             \
 	    $(BUILD_IMAGE)                                         \
 	    /bin/sh -c "                                           \
 	        ./build/test.sh ./...                              \
 	    "
+	@if [ -n "$(HTTP_PROXY)" ]; then \
+		export HTTP_PROXY="$(HTTP_PROXY)"; \
+	fi
+	@if [ -n "$(HTTPS_PROXY)" ]; then \
+		export HTTPS_PROXY="$(HTTPS_PROXY)"; \
+	fi
+	@if [ -n "$(NO_PROXY)" ]; then \
+		export NO_PROXY="$(NO_PROXY)"; \
+	fi
 	VERBOSE=1 ./test_e2e.sh
 
 TEST_TOOLS := $(shell find _test_tools/* -type d -printf "%f ")
 test-tools: $(foreach tool, $(TEST_TOOLS), .container-test_tool.$(tool))
 
 .container-test_tool.%: _test_tools/% _test_tools/%/*
-	docker build -t $(REGISTRY)/test/$$(basename $<) $<
+	docker build --build-arg ALPINE_REGISTRY_PREFIX="$(ALPINE_REGISTRY_PREFIX)" -t $(REGISTRY)/test/$$(basename $<) $<
 	docker images -q $(REGISTRY)/test/$$(basename $<) > $@
 
 # Help set up multi-arch build tools.  This assumes you have the tools
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,9 @@ BUILDX_BUILDER_SKIP_CREATION ?=
 # Allow alpine to be pulled from a private registry when building the end-to-end tests images
 ALPINE_REGISTRY_PREFIX ?=
 
+# By default all end-to-end tests are executed, but this allows for a manual selection
+LIST_OF_E2E_TESTS ?=
+
 ###
 ### These variables should not need tweaking.
 ###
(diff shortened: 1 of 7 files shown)


thockin (commented):
I appreciate the PR, but I don't think we want this.  The argument to test_e2e.sh is a pattern, not just a list, and besides that, you can just run the script.  It's easier than passing a named argument to make.

NBardelot:
> You can just run the script. It's easier than passing a named argument to make.

Cf. my comment in the `test_e2e.sh script`, the script itself makes a call to `make` without any possibility to override the makefile's extra parameters (for _buildx_, or to use fully-qualified images). Thus, you cannot run `test_e2e.sh` alone in the same conditions you build the rest of the project.

Running a subset of tests directly from `make` is the easiest workaround, with no refactoring needed. In an ideal world though, the `make` calls in `test_e2e.sh` would probably better be refactored out of the script itself, and in the makefile as a dependency goal.

By the way, you might already know of [`bats-core`](https://bats-core.readthedocs.io/en/stable/) as a shell test framework, but if you don't it could be of interest to you, given it offloads a lot of the mechanisms you have to maintain in the `test_e2e.sh`. It could also help split the tests in smaller test suits, easier to manage for a newcomer like me :)

test_e2e.sh
```diff
@@ -3670,6 +3676,9 @@ if [[ "$#" == 0 ]]; then
 fi
 
 # Build it
+# NOTE: If you want to run the end-to-end tests locally and you need specific Makefile arguments
```
thockin, line 3681: why not pass the important args?

thockin:
Somehow I only saw part of this PR when I looked at it before.

I learned of bats well after I did all the work, so didn't bother to convert it.

Makefile
```diff
@@ -248,17 +254,27 @@ test: $(BUILD_DIRS)
 	    -v $$(pwd)/.go/cache:/.cache                           \
 	    --env HTTP_PROXY=$(HTTP_PROXY)                         \
 	    --env HTTPS_PROXY=$(HTTPS_PROXY)                       \
+	    --env NO_PROXY=$(NO_PROXY)                             \
 	    $(BUILD_IMAGE)                                         \
 	    /bin/sh -c "                                           \
 	        ./build/test.sh ./...                              \
 	    "
+	@if [ -n "$(HTTP_PROXY)" ]; then \
```
thockin, line 270: Does this do anything?  Each of these blocks runs in its own shell, so `export` has no meaning:

```
thockin@thockin-glaptop4:/tmp/m$ cat Makefile 
foo:
	@export FOO=bar
	@echo $$FOO

bar:
	@export FOO=bar; \
	echo $$FOO
thockin@thockin-glaptop4:/tmp/m$ make foo

thockin@thockin-glaptop4:/tmp/m$ make bar
bar
```

test_e2e.sh
```diff
@@ -3451,7 +3451,7 @@ function e2e::touch_file_abs_path() {
 function e2e::github_https() {
     GIT_SYNC \
         --one-time \
-        --repo="https://github.com/kubernetes/git-sync" \
+        --repo="${GIT_SYNC_REPOSITORY_URL:-https://github.com/kubernetes/git-sync}" \
```
thockin: Can I get something like:

```
local default_repo="https://github.com/kubernetes/git-sync"
local repo="${REMOTE_TEST_REPO_URL:-$default_repo}"
```

Makefile
```diff
@@ -43,6 +43,9 @@ BUILDX_BUILDER_SKIP_CREATION ?=
 # Allow alpine to be pulled from a private registry when building the end-to-end tests images
 ALPINE_REGISTRY_PREFIX ?=
 
+# By default all end-to-end tests are executed, but this allows for a manual selection
```
thockin: clarify that the values here are regexes against the testcase names

Makefile
```diff
@@ -268,7 +271,7 @@ test: $(BUILD_DIRS)
 	@if [ -n "$(NO_PROXY)" ]; then \
 		export NO_PROXY="$(NO_PROXY)"; \
 	fi
-	VERBOSE=1 ./test_e2e.sh
+	VERBOSE=1 ./test_e2e.sh $(LIST_OF_E2E_TESTS)
```
thockin, line 279: does make pass "" here if the list is empty?  I think not, but I am not sure

test_e2e.sh
NBardelot: At first glance I was surpised that the script was running its prerequisites by calling the Makefile. And I ended up calling the given goals ahead of time myself, and finding it simpler. The alternative would be to map 1..1 every Makefile arg as an env variable in the script, which is possible but not very practical.

Makefile
NBardelot: My bad. Let me add a commit to fix this.

I think someone who uses a proxy will have all three variables already configured in a dotenv file or profile, expecting the NO_PROXY standard variable to be passed alongside the two others. It looks strange when this one is not. So the goal here was to be consistent (and not having to debug why something will be routed through the proxy when the user's config contains an exclusion, for example: the private/local images registries, or the github project's local mirror).

Makefile
NBardelot: Without quotes no argument is passed.

Example:

`printargs.sh`
```bash
#!/bin/bash
for a in "$@"; do
  echo "Arg: >$a<"
done
```

```bash
> ./printargs.sh foo $not_declared bar $not_declared
Arg: >foo<
Arg: >bar<
```

vs.

```bash
> ./printargs.sh foo "$not_declared" bar "$not_declared"
Arg: >foo<
Arg: ><
Arg: >bar<
Arg: ><
```

Makefile
```diff
@@ -237,6 +249,8 @@ release:
 version:
 	echo $(VERSION)
 
+# Run the 'test' goal in a single shell
+test: .SHELLFLAGS = -c
```
thockin, line 253: The default is already `-c`

Makefile
thockin: I'm fine with setting the variables for subprocesses.  I don't understand what this specific change is doing.

If I apply this delta:

```diff
diff --git Makefile Makefile
index 4b1a9c0..3348685 100644
--- Makefile
+++ Makefile
@@ -278,6 +278,9 @@ test: $(BUILD_DIRS)
 	fi
 	VERBOSE=1 ./test_e2e.sh $(LIST_OF_E2E_TESTS)
 
+foo:
+	echo $$HTTP_PROXY $$HTTPS_PROXY $$NO_PROXY
+
 TEST_TOOLS := $(shell find _test_tools/* -type d -printf "%f ")
 test-tools: $(foreach tool, $(TEST_TOOLS), .container-test_tool.$(tool))
 
```

Here's what I see:

```
$ HTTP_PROXY=123 HTTPS_PROXY=456 NO_PROXY=789 make foo
123 456 789
$ make foo HTTP_PROXY=123 HTTPS_PROXY=456 NO_PROXY=789
123 456 789
$ sh -c "export HTTP_PROXY=123; export HTTPS_PROXY=456; export NO_PROXY=789; make foo"
123 456 789
```

In all 3 cases the variable are already passed down.  Moreover, if I copy this pattern you have:

```diff
diff --git Makefile Makefile
index 4b1a9c0..1212d2f 100644
--- Makefile
+++ Makefile
@@ -278,6 +278,18 @@ test: $(BUILD_DIRS)
 	fi
 	VERBOSE=1 ./test_e2e.sh $(LIST_OF_E2E_TESTS)
 
+foo:
+	@if [ -n "$(HTTP_PROXY)" ]; then \
+		export HTTP_PROXY="111 $(HTTP_PROXY)"; \
+	fi
+	@if [ -n "$(HTTPS_PROXY)" ]; then \
+		export HTTPS_PROXY="222 $(HTTPS_PROXY)"; \
+	fi
+	@if [ -n "$(NO_PROXY)" ]; then \
+		export NO_PROXY="333 $(NO_PROXY)"; \
+	fi
+	echo $$HTTP_PROXY $$HTTPS_PROXY $$NO_PROXY
+
 TEST_TOOLS := $(shell find _test_tools/* -type d -printf "%f ")
 test-tools: $(foreach tool, $(TEST_TOOLS), .container-test_tool.$(tool))
 
```

I get:

```
$ HTTP_PROXY=123 HTTPS_PROXY=456 NO_PROXY=789 make foo
123 456 789
$ HTTP_PROXY=123 HTTPS_PROXY=456 NO_PROXY=789 make foo
123 456 789
$ sh -c "export HTTP_PROXY=123; export HTTPS_PROXY=456; export NO_PROXY=789; make foo"
123 456 789
```

The change you made to this rule doesn't DO anything.  What problem spurred you to make this specific piece of the change?

test_e2e.sh
thockin: How many would we actually need to pass?  On one hand this doesn't affect me personally so I could just not care.  OTOH it feels like leaving a trap for future contributors...

PR #976, opened 2026-01-30
Automatically re-read password-file during each sync cycle

### LINKED ISSUE #975: Feature request: add --password-file-reload flag to pick up refreshed tokens
- We have been using the `--password-file` flag to read a token generated by vault injector. This has been working great
- Going forward, we'll switch to Vault Dynamic Engine which mints short lived tokens and are automatically rotated
- Current behavior of `--password-file` is to read the token once, store in-memory and keep reusing for all subsequent calls. This will no longer work with dynamic engine because the token would have changed
- Requesting the addition of `--password-file-reload` boolean flag which when set will always read the token from file system during each sync cycle. Default value will be false so the change will be backward compatibility

I am already working on the change. Will open a PR soon. Thanks.

vaibhavdesai137:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it:**
Adds flags for --hooks-async and --hooks-before-symlink

**Which issue(s) this PR fixes:**
Fixes https://github.com/kubernetes/git-sync/issues/975

Special notes for your reviewer:

Does this PR introduce a user-facing change?:
- No, the initial approach to create a new flag is gone based on review comments
- The PR now changes the default behavior of `--askpass-url` to read password from file on every sync

/kind feature

### DIFF
--- a/main.go
+++ b/main.go
@@ -245,6 +245,9 @@ func main() {
 	flPasswordFile := pflag.String("password-file",
 		envString("", "GITSYNC_PASSWORD_FILE", "GIT_SYNC_PASSWORD_FILE"),
 		"the file from which the password or personal access token for git auth will be sourced")
+	flPasswordFileReload := pflag.Bool("password-file-reload",
+		envBool(false, "GITSYNC_PASSWORD_FILE_RELOAD"),
+		"reload the password from --password-file on each sync cycle (useful when password is rotated by an external system like Vault Dynamic Engine)")
 	flCredentials := pflagCredentialSlice("credential", envString("", "GITSYNC_CREDENTIAL"), "one or more credentials (see --man for details) available for authentication")
 
 	flSSHKeyFiles := pflag.StringArray("ssh-key-file",
@@ -883,7 +886,20 @@ func main() {
 	refreshCreds := func(ctx context.Context) error {
 		// These should all be mutually-exclusive configs.
 		for _, cred := range *flCredentials {
-			if err := git.StoreCredentials(ctx, cred.URL, cred.Username, cred.Password); err != nil {
+			password := cred.Password
+
+			// If reload flag is set and this credential has a password file,
+			// re-read it from disk to pick up token rotation (e.g. from Vault)
+			if *flPasswordFileReload && cred.PasswordFile != "" {
+				passwordFileBytes, err := os.ReadFile(cred.PasswordFile)
+				if err != nil {
+					return fmt.Errorf("can't reload password file %q: %w", cred.PasswordFile, err)
+				}
+				password = string(passwordFileBytes)
+				git.log.V(3).Info("reloaded password from file", "file", cred.PasswordFile)
+			}
+
+			if err := git.StoreCredentials(ctx, cred.URL, cred.Username, password); err != nil {
 				return err
 			}
 		}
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -245,6 +245,9 @@ func main() {
 	flPasswordFile := pflag.String("password-file",
 		envString("", "GITSYNC_PASSWORD_FILE", "GIT_SYNC_PASSWORD_FILE"),
 		"the file from which the password or personal access token for git auth will be sourced")
+	flPasswordFileReload := pflag.Bool("password-file-reload",
+		envBool(false, "GITSYNC_PASSWORD_FILE_RELOAD"),
+		"reload the password from --password-file on each sync cycle (useful when password is rotated by an external system like Vault Dynamic Engine)")
```
thockin: let's not name specific products :)


```suggestion
		"reload the password from --password-file on each sync cycle, useful when password is rotated by an external system")
```

main.go
```diff
@@ -883,7 +886,20 @@ func main() {
 	refreshCreds := func(ctx context.Context) error {
 		// These should all be mutually-exclusive configs.
 		for _, cred := range *flCredentials {
-			if err := git.StoreCredentials(ctx, cred.URL, cred.Username, cred.Password); err != nil {
+			password := cred.Password
```
thockin, line 873: Is there a reason NOT to just re-read the file on every loop?

IOW, why not get rid of the "Finish populating credentials" block above (L752) and reload it here, if cred.PasswordFile is set?

If you set the period to milliseconds it might hurt (but this is the wrong tool for that anyway).  On the order of seconds or longer it seems irrelevant and net simpler.

thockin (commented):
What I need to make this happy is an e2e test (test_e2e.sh).  Use a file in $WORK which we can change (write tmp file, rename).

The test could do something like:

* Run once with the correct password, verify the result.
* Change the password file contents to the wrong password
* Commit a change to the repo
* Verify the local sync did not change
* Change the password file back
* Verify it synced to the new change

Look at `function e2e::auth_http_password_file()` and `function e2e::auth_askpass_url_sometimes_wrong()` as examples?

main.go
vaibhavdesai137: > Is there a reason NOT to just re-read the file on every loop?

Funny you mention this. In my local fork, I ended up re-reading from file everytime. Then I thought maybe others would find the feature helpful too so I decide to contribute back. Then I thought I don't want to upset idealists by forcing a disk read on every sync. And to keep changes backward compatible, I introduced a new flag.

If this were my repo, I certainly would have changed the default behavior :) 

Let me update the PR. Will add tests too. 

Thanks for the quick review

vaibhavdesai137: Updated the PR to reload on each sync. Added e2e test as well.

main.go
```diff
@@ -2584,7 +2584,10 @@ OPTIONS
     --password-file <string>, $GITSYNC_PASSWORD_FILE
             The file from which the password or personal access token (see
             github docs) to use for git authentication (see --username) will be
-            read.  See also $GITSYNC_PASSWORD.
+            read.  The file is re-read before each sync attempt, allowing
+            git-sync to pick up token rotations automatically (e.g. when using
+            dynamic credentials from Vault's dynamic GitHub secrets engine).
```
thockin:
```suggestion
            dynamic credentials from an external secrets system).
```

thockin:
Only 1 small nit, and I need you to sign the foundation's CLA

main.go
vaibhavdesai137: Oh shoot. I thought I had removed references to Vault. My bad.

vaibhavdesai137:
> Only 1 small nit, and I need you to sign the foundation's CLA

I started the signing process but it then asked for my home address so I left it

vaibhavdesai137:
> Only 1 small nit, and I need you to sign the foundation's CLA

I have signed the CLA

thockin:
Thanks!

/lgtm
/approve

thockin:
Add `--max-failures=100` to the commandline in this test case and it should pass tests

vaibhavdesai137:
> Add `--max-failures=100` to the commandline in this test case and it should pass tests

Thank you 🙏 . I have been struggling to get the build and test working on my laptop. All sorts of weird bash and docker issues. Directly made your proposed change via GitHub web UI 🙃 🤞

thockin:
/approve
/lgtm

assuming tests pass, it will merge

thockin:
Re-titled for posterity

vaibhavdesai137:
> Re-titled for posterity

Thank you. It bothered me too but I didn't have the permission to update the title. Btw, we are now using the updated version and the reload works like a charm. Thanks again for your help.

thockin:
Thanks for the PR!  We already have someone asking for what I think is exactly this, and (hopefully) they get it just by updating.

PR #980, opened 2026-04-04
add --init-period and --init-max-failures flags

### LINKED ISSUE #959: Feature Request: Different period setting for the initial sync
Is it possible to specify a different `period` option for the initial sync?

What I'd like to be able to do is have git-sync retry every 500ms until it manages to complete the initial sync and thereafter update every 60s.

Our use case is an envoy sidecar that takes a second or so to start up and provide internet access, unfortunatly git-sync is too fast and the initial sync fails; we then have to wait a minute before it retries.

Cheers

knQzx:
## summary
- adds `--init-period` flag - used for retries until the first successful sync, then switches to normal `--period`
- adds `--init-max-failures` flag - aborts after N consecutive failures during the initial sync phase, then --max-failures takes over
- adds a SYNC PHASES section to --man and README explaining init vs steady-state behavior
- both flags are optional and backward compatible - behavior is unchanged if not set

## test plan
- unit tests for `chooseWaitTime()` and `isInitFailuresExceeded()` helper functions
- e2e tests for both flags
- verified all existing tests still pass with `go test ./...`
- build tested with `go build ./...`

fixes #959


main.go
```diff
@@ -2589,6 +2636,18 @@ OPTIONS
             dynamic credentials from an external secrets system).
             See also $GITSYNC_PASSWORD.
 
+    --init-period <duration>, $GITSYNC_INIT_PERIOD
```
thockin: please keep this section alphabetical -- these should go earlier

thockin: Also, please add to README

thockin (commented):
Thanks. The code looks good, but I'd really like e2e tests to cover these (test_e2e.sh)

knQzx:
thanks for the feedback, fixed - moved flags to alphabetical order, added readme docs, and added e2e tests for both init-period and init-timeout

thockin (commented):
Thanks!

/lgtm
/approve

main.go
```diff
@@ -912,20 +924,35 @@ func main() {
...
 	failCount := 0
 	syncCount := uint64(0)
+	initialSyncDone := false
+	var initStart time.Time
+	if *flInitPeriod != 0 || *flInitTimeout != 0 {
+		initStart = time.Now()
+	}
 	for {
 		start := time.Now()
 		ctx, cancel := context.WithTimeout(context.Background(), *flSyncTimeout)
```
thockin, line 934: Your e2e doesn't pass.

Don't you need to pass the initTimeout here until you are past init stage?  I am seeing this with a very large number (120s) which defeats your e2e test (which waits 2s).

Why is it better to frame this as init-timeout rather than init-max-failures ?

We should maybe add a description of the initialization vs. regular run to the `--man` string.

knQzx:
pushed a fix: the init-timeout test was using a valid repo with a slow fetch wrapper, so git-sync would succeed after 2s and run forever instead of timing out. swapped to a non-existent repo path so fetches actually fail and init-timeout fires as expected.

main.go
knQzx: thanks for testing this locally.

1. e2e fix is in the latest commit. old test used a slow-fetch wrapper over a valid repo so git-sync succeeded and kept running forever. now uses an unreachable repo URL so fetches actually fail and init-timeout triggers.

2. you're right about line 934. --sync-timeout (default 120s) dominates, so --init-timeout doesn't really bound user-visible time like users would expect. can cap the SyncRepo context at min(syncTimeout, remainingInitTimeout) while initialSyncDone is false. want me to go that route?

3. on init-timeout vs init-max-failures: time-based fits k8s pod-start deadlines where wall-clock matters. count-based is simpler and mirrors --max-failures. happy either way. if you'd prefer --init-max-failures I can swap, or add both.

4. will add an init-phase section to the man text regardless.

direction you want: init-timeout with context fix, or pivot to init-max-failures?

thockin: I think init-max-failures is easier to reason about and is more consistent with the "main" loop.  I assume the real goal here is that the init phase retries more aggressively and "gives up" after a while (in case of some connectivity issue or something) but once it works, you enter steady-state polling.

Regarding e2e: I think it was correct to use SLOW_GIT_FETCH on a real repo as long as the feature is "init timeout".  If we switch to init-max-failures then testing a bad server is the right approach

Thanks

knQzx:
reworked per the review:
- renamed --init-timeout to --init-max-failures (int, counts failures)
- e2e test now targets a bad server path as you suggested
- added a SYNC PHASES section in --man and README describing init vs steady-state
- also skipped --max-failures during init when --init-max-failures is set - otherwise main-max could override init tolerance, caught this while reviewing

lmk if the new shape works

main.go
```diff
@@ -183,6 +183,12 @@ func main() {
 	flErrorFile := pflag.String("error-file",
 		envString("", "GITSYNC_ERROR_FILE", "GIT_SYNC_ERROR_FILE"),
 		"the path (absolute or relative to --root) to an optional file into which errors will be written (defaults to disabled)")
+	flInitMaxFailures := pflag.Int("init-max-failures",
```
thockin: super nit: can this be next to the regular max-failures definition?

main.go
```diff
@@ -469,6 +475,12 @@ func main() {
 	if *flPeriod < 10*time.Millisecond {
 		fatalConfigErrorf(log, true, "invalid flag: --period must be at least 10ms")
 	}
+	if *flInitPeriod != 0 && *flInitPeriod < 10*time.Millisecond {
```
thockin: Would it make later code simpler to do something like:

```
if *flInitPeriod == 0 {
  *flInitPeriod = *flPeriod
}
```

?

main.go
```diff
@@ -912,20 +924,35 @@ func main() {
...
+	initialSyncDone := false
 	for {
 		start := time.Now()
 		ctx, cancel := context.WithTimeout(context.Background(), *flSyncTimeout)
 
 		if changed, hash, err := git.SyncRepo(ctx, refreshCreds); err != nil {
 			failCount++
 			updateSyncMetrics(metricKeyError, start)
-			if *flMaxFailures >= 0 && failCount >= *flMaxFailures {
+			if isInitFailuresExceeded(*flInitMaxFailures, initialSyncDone, failCount) {
```
thockin: I find this hard to reason about.  Could it be reduced to something like:

```
getMaxFailures := func() int {
    if initPhase && *flInitMaxFailures > 0 {
        return *flInitMaxFailures
    }
    if *flMaxFailures > 0 {
        return *flMaxFailures
    }
    return -1
}
if maxFails := getMaxFailures(); maxFails >= 0 && failCount >= maxFails {
```

main.go
```diff
@@ -989,12 +1016,14 @@ func main() {
 			log.DeleteErrorFile()
 		}
 
-		log.V(3).Info("next sync", "waitTime", flPeriod.String(), "syncCount", syncCount)
+		// Use init-period for retries before the first successful sync.
+		waitTime := chooseWaitTime(*flPeriod, *flInitPeriod, initialSyncDone)
```
thockin: Rather thsan calling out to a func, why not just track a variable `waitTime` outside the loop?  Initialize it simply and flip from InitPeriod to Period when we leave init phase.

main.go
```diff
@@ -2556,6 +2623,19 @@ OPTIONS
             Enable the pprof debug endpoints on git-sync's HTTP endpoint at
             /debug/pprof.  Requires --http-bind to be specified.
 
+    --init-max-failures <int>, $GITSYNC_INIT_MAX_FAILURES
+            The number of consecutive failures allowed before aborting during
+            the initial sync phase (before the first successful sync).  Once
+            the initial sync succeeds, --max-failures applies instead.  Set
+            to 0 (the default) to disable this separate limit and fall
```
thockin: If we set it to 0, that means there's no way to express "do not retry".

Doesn't pflag have a way to tell if the flag was specified, so we could leave this with -1 meaning unlimited, 0 meaning 0, and unspecified meaning to fall back to regular max-fail?

knQzx:
applied all five. small deviation in getMaxFailures: kept `return *flMaxFailures` instead of the literal `> 0` gate, otherwise `--max-failures=0` would flip from "abort on first failure" (current docs) to "never abort". outer guard is unchanged. lmk if you'd rather have the literal shape and we touch the main docs too

thockin:
Thanks!

/lgtm
/approve

PR #982, opened 2026-04-06
Add "--filter" flag to enable git partial clone

### LINKED ISSUE #981: Add --filter flag for partial clone support
## Summary

I'd like to propose adding `--filter` support (partial clone) to git-sync. This was previously explored in #701, but was closed without implementation after a benchmark suggested that `--filter=blob:none` combined with `--depth` offered no benefit over `--depth=1` alone.

However, that benchmark checked out the full working tree, which negated the benefit of `--filter`. When combined with **sparse-checkout** (which git-sync already supports via `--sparse-checkout-file`), the picture changes significantly.

## Analysis

| Phase | `--depth` only | `--filter` + `--depth` + sparse-checkout |
|---|---|---|
| Fetch: commits/trees | 1 commit | 1 commit |
| Fetch: blobs | **All blobs** | **None** |
| Checkout: blobs | All files | **Only files matching sparse patterns** |

With all three combined, data transfer is minimized at every stage:

1. `--depth` limits commit and tree objects
2. `--filter=blob:none` eliminates blob transfer at fetch time
3. `sparse-checkout` ensures only the needed blobs are fetched on demand at checkout time

This is particularly valuable for large monorepos where users only need a subset of files — a common use case for git-sync as a sidecar.

## Proposed design

- Add a new `--filter` flag (e.g., `--filter=blob:none`)
- It should work alongside `--depth` and `--sparse-checkout-file`
- The implementation should be straightforward since git-sync calls the `git` CLI directly — it's essentially adding `--filter` to the `fetch` command

kane8n:
## Summary
Closes #981

- Add `--filter` flag (`$GITSYNC_FILTER`) to support git partial clone
- Passes `--filter` to `git fetch`, enabling filters like `blob:none` or `tree:0`
- Most effective when combined with `--depth` and `--sparse-checkout-file` to minimize data transfer at every stage (commits, trees, and blobs)

## Details
For large monorepos where users only need a subset of files, combining `--filter=blob:none` + `--depth` + `--sparse-checkout-file` ensures:

1. `--depth` limits commit and tree objects
2. `--filter=blob:none` eliminates blob transfer at fetch time
3. `sparse-checkout` fetches only needed blobs on demand at checkout

### DIFF
--- a/main.go
+++ b/main.go
@@ -120,6 +120,7 @@ type repoSync struct {
 	repo           string         // remote repo to sync
 	ref            string         // the ref to sync
 	depth          int            // for shallow sync
+	filter         string         // for partial clone
 	submodules     submodulesMode // how to handle submodules
 	gc             gcMode         // garbage collection
 	link           absPath        // absolute path to the symlink to publish
@@ -167,6 +168,9 @@ func main() {
 	flDepth := pflag.Int("depth",
 		envInt(1, "GITSYNC_DEPTH", "GIT_SYNC_DEPTH"),
 		"create a shallow clone with history truncated to the specified number of commits")
+	flFilter := pflag.String("filter",
+		envString("", "GITSYNC_FILTER"),
+		"use partial clone with the specified filter (e.g. 'blob:none', 'tree:0')")
 	flSubmodules := pflag.String("submodules",
 		envString("recursive", "GITSYNC_SUBMODULES", "GIT_SYNC_SUBMODULES"),
 		"git submodule behavior: one of 'recursive', 'shallow', or 'off'")
@@ -709,6 +713,7 @@ func main() {
 		repo:         *flRepo,
 		ref:          *flRef,
 		depth:        *flDepth,
+		filter:       *flFilter,
 		submodules:   submodulesMode(*flSubmodules),
 		gc:           gcMode(*flGitGC),
 		link:         absLink,
@@ -1860,6 +1865,9 @@ func (git *repoSync) fetch(ctx context.Context, ref string) error {
 			args = append(args, "--unshallow")
 		}
 	}
+	if git.filter != "" {
+		args = append(args, "--filter", git.filter)
+	}
 	if _, _, err := git.Run(ctx, git.root, args...); err != nil {
 		return err
 	}
@@ -2444,6 +2452,13 @@ OPTIONS
             number of commits.  If not specified, this defaults to syncing a
             single commit.  Setting this to 0 will sync the full history of the
             repo.
+
+    --filter <string>, $GITSYNC_FILTER
+            Use partial clone with the specified filter.  This can reduce
+            the amount of data transferred when cloning large repositories.
+            Common values are 'blob:none' (omit all blobs, fetch on demand)
+            and 'tree:0' (omit all trees and blobs).  This is most effective
+            when combined with --depth and --sparse-checkout-file.
 
     --error-file <string>, $GITSYNC_ERROR_FILE
             The path to an optional file into which errors will be written.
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -2445,6 +2453,13 @@ OPTIONS
             single commit.  Setting this to 0 will sync the full history of the
             repo.
 
+    --filter <string>, $GITSYNC_FILTER
```
thockin: Please keep these flags alphabetically sorted

main.go
```diff
@@ -2445,6 +2453,13 @@ OPTIONS
...
+            Use partial clone with the specified filter.  This can reduce
+            the amount of data transferred when cloning large repositories.
+            Common values are 'blob:none' (omit all blobs, fetch on demand)
+            and 'tree:0' (omit all trees and blobs).  This is most effective
```
thockin: Can you reference `git fetch` docs so users can learn about the syntax?  E.g. `See docs for "git fetch --filter" for more information`.

main.go
thockin: You also need to update README.md which has these docs embedded (you can delete the last section then just run `git-sync --man >> README.md`)

kane8n:
@thockin 
Thank you for the review! I've addressed all the feedback:

- Sorted `--filter` flag alphabetically in the help text
- Added a reference to the partial clone documentation (https://git-scm.com/docs/partial-clone)
- Regenerated README.md from `--man` output

PTAL when you get a chance.

thockin:
Thanks!

/lgtm
/approve

thockin:
e2e fails for `filter_partial_clone_blob_none`

test_e2e.sh
```diff
@@ -3339,6 +3339,99 @@ function e2e::sparse_checkout() {
     assert_file_eq "$ROOT/link/file2" "${FUNCNAME[0]}"
 }
 
+##############################################
+# Test filter (partial clone) with blob:none
+##############################################
+function e2e::filter_partial_clone_blob_none() {
+    echo "${FUNCNAME[0]}" > "$REPO/file"
```
thockin: Tests begin with the funcname already in the file, which makes this redundant

test_e2e.sh
```diff
@@ -3339,6 +3339,99 @@ function e2e::sparse_checkout() {
...
+        --root="$ROOT" \
+        --link="link" \
+        --filter="blob:none" \
+        --depth=0
+    assert_link_exists "$ROOT/link"
+    assert_file_exists "$ROOT/link/file"
+    assert_file_eq "$ROOT/link/file" "${FUNCNAME[0]}"
+
+    # Verify the repo is a partial clone
+    if ! git -C "$ROOT/link" config --get remote.origin.promisor >/dev/null 2>&1; then
```
thockin: Can you explain in comments what this block is doing?  It doesn't seem to work - `remote.origin.partialclonefilter` fails (which you are hiding with `|| true`)

test_e2e.sh
```diff
@@ -3339,6 +3339,99 @@ function e2e::sparse_checkout() {
...
+}
+
+##############################################
+# Test filter with sparse-checkout
+##############################################
+function e2e::filter_with_sparse_checkout() {
+    echo "!/*" > "$WORK/sparseconfig"
+    echo "!/*/" >> "$WORK/sparseconfig"
+    echo "file2" >> "$WORK/sparseconfig"
+    echo "${FUNCNAME[0]}" > "$REPO/file"
```
thockin: same as above - this is a noop

kane8n:
@thockin 
Thanks for the review. I've addressed the points you raised.
As for the e2e tests, I'm not able to run them in my local environment, so it would be helpful if you could verify them via CI.

thockin:
Thanks!

/lgtm
/approve

PR #990, opened 2026-06-11
[Feat] Pre exechook

### LINKED ISSUE #937: How to Prevent .env and upload from Being Deleted During git-sync (Similar to .gitignore)

Mohamed-elg:
/kind feature

What this PR does:
Fixes https://github.com/kubernetes/git-sync/issues/937

adds flags for `--pre-publish-exechook-command`, `--pre-publish-exechook-timeout` and `--pre-publish-exechook-backoff` to run a command similarly to the exechook but before publishing the new symlink

Follows https://github.com/kubernetes/git-sync/issues/937#issuecomment-2845597988

### DIFF
--- a/main.go
+++ b/main.go
@@ -237,15 +237,15 @@ func main() {
 		envDuration(3*time.Second, "GITSYNC_EXECHOOK_BACKOFF", "GIT_SYNC_EXECHOOK_BACKOFF"),
 		"the time to wait before retrying a failed exechook")
 
-	flPreExechookCommand := pflag.String("pre-exechook-command",
-		envString("", "GITSYNC_PRE_EXECHOOK_COMMAND", "GIT_SYNC_PRE_EXECHOOK_COMMAND"),
+	flPrePubExechookCommand := pflag.String("pre-publish-exechook-command",
+		envString("", "GITSYNC_PRE_PUBLISH_EXECHOOK_COMMAND"),
 		"an optional command to be run before syncs complete (must be idempotent)")
-	flPreExechookTimeout := pflag.Duration("pre-exechook-timeout",
-		envDuration(30*time.Second, "GITSYNC_PRE_EXECHOOK_TIMEOUT", "GIT_SYNC_PRE_EXECHOOK_TIMEOUT"),
-		"the timeout for the pre-exechook")
-	flPreExechookBackoff := pflag.Duration("pre-exechook-backoff",
-		envDuration(3*time.Second, "GITSYNC_PRE_EXECHOOK_BACKOFF", "GIT_SYNC_PRE_EXECHOOK_BACKOFF"),
-		"the time to wait before retrying a failed pre-exechook")
+	flPrePubExechookTimeout := pflag.Duration("pre-publish-exechook-timeout",
+		envDuration(30*time.Second, "GITSYNC_PRE_PUBLISH_EXECHOOK_TIMEOUT"),
+		"the timeout for the pre-publish-exechook")
+	flPrePubExechookBackoff := pflag.Duration("pre-publish-exechook-backoff",
+		envDuration(3*time.Second, "GITSYNC_PRE_PUBLISH_EXECHOOK_BACKOFF"),
+		"the time to wait before retrying a failed pre-publish-exechook")
 
 	flWebhookURL := pflag.String("webhook-url",
 		envString("", "GITSYNC_WEBHOOK_URL", "GIT_SYNC_WEBHOOK_URL"),
@@ -558,12 +558,12 @@ func main() {
 		}
 	}
 
-	if *flPreExechookCommand != "" {
-		if *flPreExechookTimeout < time.Second {
-			fatalConfigErrorf(log, true, "invalid flag: --pre-exechook-timeout must be at least 1s")
+	if *flPrePubExechookCommand != "" {
+		if *flPrePubExechookTimeout < time.Second {
+			fatalConfigErrorf(log, true, "invalid flag: --pre-publish-exechook-timeout must be at least 1s")
 		}
-		if *flPreExechookBackoff < time.Second {
-			fatalConfigErrorf(log, true, "invalid flag: --pre-exechook-backoff must be at least 1s")
+		if *flPrePubExechookBackoff < time.Second {
+			fatalConfigErrorf(log, true, "invalid flag: --pre-publish-exechook-backoff must be at least 1s")
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -2528,6 +2583,21 @@ OPTIONS
             The timeout for the --exechook-command.  If not specifid, this
             defaults to 30 seconds ("30s").
 
+    --pre-exechook-backoff <duration>, $GITSYNC_PRE_EXECHOOK_BACKOFF
```
thockin: Flags should be in alphabetic-sorted order, sorry

main.go
```diff
@@ -2528,6 +2583,21 @@ OPTIONS
...
+            The time to wait before retrying a failed --pre-exechook-command.  If
+            not specified, this defaults to 3 seconds ("3s").
+
+    --pre-exechook-command <string>, $GITSYNC_PRE_EXECHOOK_COMMAND
+            An optional command to be executed before syncing a new hash of the
```
thockin: This isn't quite right.

I think we should update the docs on `--exechook-command` to say something like:

"""
An optional command to be executed after syncing a new hash of the remote repository and publishing the symlink (see --link).
"""

Then this can be:

"""
An optional command to be executed after syncing a new hash of the remote repository but before publishing the symlink (see --link).
"""

The rest should the same.  What do you think?

main.go
```diff
@@ -948,6 +996,10 @@ func main() {
 		start := time.Now()
 		ctx, cancel := context.WithTimeout(context.Background(), *flSyncTimeout)
 
+		if preExechookRunner != nil {
```
thockin: This doesn't seem right?  This will send the old hash on every loop, then sync a new hash.  Unless I totally misunderstood the request, you want to try to sync, and ONLY IF there is a new hash, call the hook before publishing the symlink.

Did I misunderstand?

Unfortunately, the sync and the publish are inside `git.SyncRepo`.  You can either decompose that (which would need to share a bunch of state) or you could pass function pointers into it like we do for `refreshCreds` - pass a struct like:

```
type syncHooks struct {
    refreshCreds func(context.Context) error
    beforePublish func(context.Context, changed bool, hash string) error
    afterPublish func(context.Context, changed bool, hash string) error
}
```

thockin (commented):
Thanks for the PR!  Notes below.  Also, this will need e2e tests

main.go
Mohamed-elg: you are right, i changed the logic to have the hooks around the symlink rather than the sync event itself

Mohamed-elg:
> Thanks for the PR! Notes below. Also, this will need e2e tests

will add the e2e tests once the changes are validated

main.go
Mohamed-elg: correct, fixed it by moving the logic inside `git.SyncRepo` as suggested

pkg/hook/exechook.go
```diff
@@ -68,10 +71,10 @@ func (h *Exechook) Do(ctx context.Context, hash string) error {
 	env := os.Environ()
 	env = append(env, envKV("GITSYNC_HASH", hash))
 
-	h.log.V(0).Info("running exechook", "hash", hash, "command", h.command, "timeout", h.timeout)
+	h.log.V(0).Info("running hook", "name", h.name, "hash", hash, "command", h.command, "timeout", h.timeout)
```
thockin: super nitpicky, but I would leave the key as "exechook".  If this is the only thing I flag, I will overlook it :)

main.go
```diff
@@ -230,6 +237,16 @@ func main() {
 		envDuration(3*time.Second, "GITSYNC_EXECHOOK_BACKOFF", "GIT_SYNC_EXECHOOK_BACKOFF"),
 		"the time to wait before retrying a failed exechook")
 
+	flPreExechookCommand := pflag.String("pre-exechook-command",
```
thockin: I am struggling with the name -- "pre-exechook" makes it sound like it runs before the exechook.

How about "prepub-exechook-*" ?  Any better ideas?

main.go
```diff
@@ -230,6 +237,16 @@ func main() {
...
+		envString("", "GITSYNC_PRE_EXECHOOK_COMMAND", "GIT_SYNC_PRE_EXECHOOK_COMMAND"),
```
thockin: We don't need the old-style names "GIT_SYNC_..." - those are just for backwards compat in older flags.

main.go
```diff
@@ -2515,7 +2591,8 @@ OPTIONS
 
     --exechook-command <string>, $GITSYNC_EXECHOOK_COMMAND
             An optional command to be executed after syncing a new hash of the
-            remote repository.  This command does not take any arguments and
+            remote repository and publishing the symlink (see --link).
+            This command does not take any arguments and
```
thockin: Please reflow this to 80 columns (it looks like "executes" can fold back up a line) but make sure to use spaces and not tabs.  Sorry, this is annoying, but I like the extra detail

main.go
```diff
@@ -2676,6 +2753,22 @@ OPTIONS
...
+    --pre-exechook-backoff <duration>, $GITSYNC_PRE_EXECHOOK_BACKOFF
+            The time to wait before retrying a failed --pre-exechook-command.  If
+            not specified, this defaults to 3 seconds ("3s").
+
+    --pre-exechook-command <string>, $GITSYNC_PRE_EXECHOOK_COMMAND
+            An optional command to be executed after syncing a new hash of the 
+            remote repository but before publishing the symlink (see --link).
+            This command does not take any arguments and
+            executes with the synced repo as its working directory. The
+            $GITSYNC_HASH environment variable will be set to the previous git hash that
```
thockin: please reflow this whole block to 80 columns

main.go
```diff
@@ -1842,7 +1908,12 @@ func (git *repoSync) SyncRepo(ctx context.Context, refreshCreds func(context.Con
 
 		// If we have a new hash, update the symlink to point to the new worktree.
 		if changed {
```
thockin, line 1915: This is super subtle, and we should leave a comment - I had to walk through it to understand the subtlety.

Something like:

"""
If the previous run crashed before publishing the link, then we must call the pre-publish hook, and since changed is true, we will.  If the previous run crashed after publishing the link, then we do not need to call the pre-publish hook, and since changed is false, we won't.  The post-publish hooks are called in both cases.

thockin (commented):
Thanks!  My comments are pretty superficial.  I think the logic looks good.  e2e tests will prove it.

main.go
Mohamed-elg: ideally it should look like `--pre-publish-symlink-exechook-command` but went for the shorter `--pre-publish-exechook-command`

Mohamed-elg:
> Thanks! My comments are pretty superficial. I think the logic looks good. e2e tests will prove it.

Thanks for the suggestions, added the e2e tests

Issue #955, opened 2025-07-02
Feature Request: Support Azure DevOps Git Repos via Managed Identity (WIF) Authentication


javierherrer:
Hi team 👋

We’d like to request support for cloning Azure DevOps (ADO) Git repositories using Managed Identity via Workload Identity Federation (WIF) in AKS, similar to the existing support for GitHub App authentication.

🧩 Context
In enterprise environments—especially those running on Azure Kubernetes Service (AKS)—it's increasingly common to use OIDC-based workload identity to authenticate workloads securely without secrets. While GitSync currently supports GitHub App-based authentication, there is no equivalent support for Azure DevOps using a managed identity.

🚧 Problem
Due to strict security policies, Personal Access Tokens (PATs) are not allowed in our environment. This makes it impossible to use GitSync with ADO repos today, even though the managed identity already has access to the repo via Entra ID.

✅ Proposal
We’d love to see GitSync support a new authentication mode that:

Uses a federated token from the pod’s projected identity (via Azure Workload Identity).
Exchanges that token for an ADO access token via Microsoft Entra ID.
Uses the resulting bearer token to authenticate Git operations (e.g., git clone) against ADO.
This would mirror the GitHub App support already in place and align with Microsoft’s push toward secretless infrastructure.

🔐 Benefits
Enables secure, secretless GitOps workflows with ADO.
Aligns with enterprise security and compliance standards.
Reduces operational overhead of managing PATs or SSH keys.

🙏 Ask
Would the maintainers be open to a PR or collaboration to explore this feature? We’re happy to help test or contribute if needed.

Thanks for your great work on GitSync!

javierherrer:
🧪 Usage example
```bash
gitsync \
    --repo=https://dev.azure.com/<org>/<project>/<repo> \
    --root=tmp/git/root \
    --ado-managed-identity=<mi-client-id>
```

📖 References
1. [RefreshGitHubAppToken gitsync function](https://github.com/kubernetes/git-sync/blob/c86639029c3d9dcdfdf65d4ce2813f350f4695f2/main.go#L2028)
2. [Authentication with Azure Repos | Microsoft Entra OAuth tokens](https://learn.microsoft.com/en-us/azure/devops/repos/git/auth-overview?view=azure-devops&tabs=Windows#microsoft-entra-oauth-tokens)
[...]

PR #996, opened 2026-07-14
Add Azure Workload Identity Federation auth for Azure DevOps

namnt2307:
## What
Adds a secretless authentication mode for Azure DevOps Services repos using Azure Workload Identity Federation. When `--azure-workload-identity` is set, git-sync exchanges the pod's projected ServiceAccount token with Microsoft Entra ID for a short-lived access token and uses it as the git credential — no PAT, SSH key, or long-lived secret is stored. Mirrors the existing GitHub App auth mode.

## Why
Implements the request in #955: environments whose policies forbid PATs currently can't use git-sync with Azure DevOps even when the workload identity already has repo access.

## Config
New flags (each with `GITSYNC_*` and `AZURE_*` env fallbacks):
- `--azure-workload-identity` — opt in
- `--azure-client-id` / `--azure-tenant-id` / `--azure-federated-token-file` / `--azure-authority-host` — default to the `AZURE_*` vars injected by the azure-workload-identity webhook; overridable as flags
- `--azure-scope` — defaults to the Azure DevOps resource ID

Docs in `docs/azure-workload-identity.md`; flags documented in `README.md`.

## Testing
- Unit tests exercise the Entra token exchange and endpoint construction against an `httptest` fake — no real Azure required, per the maintainer note in #955.
- Manually verified end-to-end against real Entra ID + Azure DevOps: a pod's federated token successfully clones a private ADO repo.

Ref #955

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,6 +242,39 @@ OPTIONS
             (200) and produce a series of key=value lines, including
             "username=<value>" and "password=<value>".
 
+    --azure-workload-identity, $GITSYNC_AZURE_WORKLOAD_IDENTITY
+            Use Azure Workload Identity Federation to authenticate to Azure
+            DevOps.  The pod's projected ServiceAccount token is exchanged for
+            a Microsoft Entra ID access token which is used as the git
+            credential, so no PAT or secret is stored.  May not be combined
+            with --username/--password, --askpass-url, --credential, or the
+            --github-app-* flags.
+
+    --azure-client-id <string>, $GITSYNC_AZURE_CLIENT_ID
+            The Microsoft Entra ID client ID to use for Azure workload identity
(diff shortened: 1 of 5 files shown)


thockin (commented):
I'm not really enthusiastic about adding cloud-provider-specific logic to this tool.

If you can find a way to frame it that works on multiple providers, I might be open to it, but I'm not going to embed a bunch of Azure-specific logic which I can't maintain or test.

We have `--askpass-url` already - can you bend that to do what you want?  E.g. run another sidecar (sidecar-sidecar) which just has a localhost-only webserver which does what you describe here and turns it into what git expects?  Then you set something like:

`git-sync --askpass-url=localhost:8086 ...`

And that's all you need.

PR #1002, opened 2026-09-17
fix: skip nonexistent paths from minimal base images

tiffanny29631:
When building with minimal or slim base images, non-essential files (such as /usr/share/common-licenses) may be absent while still listed in dpkg manifests.
In stage_one_file(), missing paths fail the directory check and fall through to `cp -lpP`, causing a fatal "No such file or directory" error. Check that the path exists before attempting to stage it.


stage_binaries.sh
```diff
@@ -193,6 +193,9 @@ function stage_one_file() {
     for elem in "${elems[@]}"; do
         path="${path}/${elem}"
         if [[ ! -e "${staging}${path}" ]]; then
+            if [[ ! -e "${path}" && ! -L "${path}" ]]; then
```
sdowell: How do we know it's a non-essential file?

sdowell: In the case where a file is explicitly specified to be copied (using `-f` option), this will fail silently. It should probably continue to return an error for the `-f` case

stage_binaries.sh
```diff
@@ -193,6 +193,9 @@ function stage_one_file() {
...
+                continue
```
sdowell: If a prefix is missing, every path under that will also be missing. This can return instead of continue

sdowell: Suggest adding a debug trace here so that it is clear from the logs that a path/file was skipped

stage_binaries.sh
sdowell, line 196: Discussion: `stage_one_package()` already filters the manifest:
```bash
dpkg -L "${pkg}" | grep_allow_nomatch -vE '(/\.|/usr/share/(man|doc|.*-completion))'
```
Could the filter be updated to include the unnecessary files?

---
Otherwise:

1. stage_one_file is also the path for user-requested -f files, so this makes -f /typo/path silently succeed instead of failing the build. Can we keep -f strict in stage_files()?
1. Nits: return 0 rather than continue (deeper paths can't exist if the prefix doesn't exist), and a debug logging "skipping nonexistent path ..." so this isn't invisible.

PR #1003, opened 2026-09-18
fix: align toolchain with root go.mod and verify licenses in CI

tiffanny29631:
/kind bug
/kind cleanup

This PR fixes a build failure during make container caused by a Go toolchain / GOROOT mismatch in go-licenses, and adds a lightweight CI check to ensure license verification is tested on every PR.


.github/workflows/main.yml
```diff
@@ -29,7 +29,10 @@ jobs:
         working-directory: git-sync
         run: |
           make all-build
-
+      - name: Check licenses
+        working-directory: git-sync
+        run: |
+          make licenses
```
sdowell, line 35: This doesn't verify licenses, this generates licenses

What is the intent of the change?

thockin:
> fixes a build failure during make container

Can you explain more?  It doesn't fail for me.

